PR #1, opened 2016-05-31
Porting over Ephemeralis's medical rework from Yogstation.

Super3222:


X-TheDark:
Do we also spruce up the code as we port it? For example, this PR could use alot of `src.var -> var` type of sprucing.

oisin100:
Ach, Ill get travis working.

Bawhoppen:
Before you do anything I still say you guys should just run pure tg code for a little bit, to get the full experience

Super3222:
that was a bad commit message. what I meant was src.var -> var

X-TheDark:
@Bawhoppen Regardless of whether or not we keep certain /tg/ features, we still need to port our own into the new code.

Also, relevant (basically says that we are actually rebasing, no-joke/serious mode) : https://forums.yogstation.net/index.php?threads/yogstation-split-2016.10630/

Bawhoppen:
yes I know, but you should run pure code for a little bit to try the full experience is what I'm saying

PR #8, opened 2016-06-01
Ports TehFlaminTaco's Throwable Spells

Super3222:


X-TheDark:
Did you have the time to test this? There was a problem with how this interacts with crit on our codebase. Does this still happen on /tg/ code?

Reason I'm asking is because I had to change how the spell datum works in https://github.com/yogstation13/yogstation/pull/1227 (basically needed to add the user explicit passing (aka usermob/user that's added everywhere).

Super3222:
@X-TheDark,
Yes, I looked at your bug fix on our main repo and tried to port them as well.

But  /TG/ had already done the same thing.

PR #26, opened 2016-06-02
Ports Oisin100's mindslave implant and Unmovable's zombie implant.

Super3222:
In the future, I may be reworking Unmovable's zombie implant. It's easy to tell that someone just didn't randomly die because it will only display that they are dead.

I'm planning on making it either seem like you suffocated to death or you have committed suicide.

// EDIT:
Zombie implant has been reworked with Adam El-Tablawy's idea.
Instead of using zombie powder and antitoxin, it will secrete a chemical known as "capilletum".
Instead of one use, you'll get two uses with Capilletum. It will also only work if you are lying on the ground.
Capilletum basically makes small damage look like bigger damage. So say, you punch yourself, activate this implant, and than lie down on the ground. 
It'll appear like you are dead & have a large amount of brute damage.

A competent MD/CMO would be able to tell you aren't /really/ dead by using a medhud.


X-TheDark:
Suffocation sounds like a good way of doing it. Unfortunately, you'd need to mess around with "fake" Brute/Burn damage if you go that route, as Suiciding gives you tons of brute/burn damage, if I remember correctly.

Cruix:
Suicide only gives oxyloss, unless you are doing it with an item that has a special suicide_act (i.e. the banhammer).

FluffySurvivor:
NONONO THE ZOMBIE IMPLANT WAS 7 NOT 12 TC

oisin100:
Fluffy is right you know.

code/game/objects/items/weapons/implants/implant_mindslave.dm
```diff
@@ -0,0 +1,90 @@
...
+<b>Important Notes:</b> <font color='red'>Illegal</font><BR>
+<HR>
+<b>Implant Details:</b> <BR>
+<b>Function:</b> Makes the injected a slave to the owner of the implant.<HR>"}
+	return dat
+
+/obj/item/weapon/implant/mindslave/implant(mob/source, var/mob/user)
+
+	if(!source.mind)
+		user.mind << "<span class='notice'>[source] doesent posses the mental capabilities to be a slave.</span>"
```
ShadowDeath6: does not or doesn't with apostrophe pls, I'm triggered

code/game/objects/items/weapons/implants/implant_mindslave.dm
```diff
@@ -0,0 +1,90 @@
...
+		return 0
+
+	var/mob/living/carbon/human/target = source.mind.current
+	var/mob/living/carbon/human/holder = user.mind.current
+
+	if(target == holder)
+		holder << "<span class='notice'>You cant implant yourself!</span>"
```
ShadowDeath6: Same here

code/modules/reagents/chemistry/reagents/toxin_reagents.dm
```diff
@@ -673,6 +673,14 @@
 	..()
 
 
+/datum/reagent/toxin/capilletum
+	name = "Capilletum"
+	id = "capilletum"
+	description = "A powerful toxin that exemplifies the patterns of punctured skin, matching their pigments and and shapes and than expanding them across the body. Unlike other toxins, it does not have any negative effects."
```
ShadowDeath6: then instead of than

oisin100:
Not a bad idea this Capilletum.

PR #52, opened 2016-06-05
Admin tickets

X-TheDark:
I ported way too many things to get this to work...
- Admin Tickets - obviously;
  ported_1

Some stuff that I really, really shouldn't have touched but did on the way to porting Admin Tickets (mostly because I touched the admin verbs - which contained alot of differences that snowballed into porting alot of misc stuff over)
- Our server-join disclaimer thing (buttons properly updated and working);
- Donor slots;
- The Ready button thing;
- Credit system and verbs (Not included in the .dme because it caused the merge conflict);
- Donor verbs (not the whole system);
- The rule html file (but the button on the top right isn't updated to this);
- Some role toggle defines (but not our role selection system - it still uses bitflags for now);
- Admin freeze stuff;
- Antag weight verbs (but not the full system);
- High-risk item destruction notification (but not z-level change, I think);
- Antag tokens (again, they may or may not work as they were ported as part, not the main focus);

I really, really need to stop porting everything "on the way" to whatever I'm porting.


code/controllers/subsystem/ticker.dm
```diff
@@ -105,15 +109,34 @@ var/datum/subsystem/ticker/ticker
...
 			check_maprotate()
 
+			if(world.time > next_alert_time && next_check_admin)
+				next_alert_time = world.time+1800 /* 6000 */
+
+				var/admins_online = total_admins_active()
+				var/unresolved_tickets = total_unresolved_tickets()
+
+				if(!admins_online)
+					next_check_admin = 0
```
X-TheDark: This is ported from the Discord bot thing, but I removed the line that actually sent the message

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -70,7 +163,8 @@
 
 /client/verb/adminhelp(msg as text)
```
oisin100: Mayaswell just comment this out or remove it entirely. 
Same goes for all the other unused verbs and procs.

code/modules/admin/admin_verbs.dm
```diff
@@ -332,10 +355,97 @@ var/list/admin_verbs_hideable = list(
 
 	src << "<span class='interface'>All of your adminverbs are now visible.</span>"
 	feedback_add_details("admin_verb","TAVVS") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
+/*
```
oisin100: Whats the point in commenting it out? Just keep it... Maybe we'll use it eventually... Commenting it out will just cause a headache when someone decides to use the credit system later on.

X-TheDark: Take note of the commits that I've made - including credits.dm was causing a merge conflict.
Had to uninclude credits.dm, which means I had to comment out all these procs (they are defined in credits.dm), I commented it out because to add it, I will simply include credits.dm in another PR (after this merged and I update my local repo, etc) and uncomment this.

Issue #64, opened 2016-06-08
Portin' : Yog's Changelings


X-TheDark:
I'll take a look at those.

X-TheDark:
Ready to be merged.

PR #69, opened 2016-06-09
Ports over Yogstation's changeling stuff + Hallucination String buff

X-TheDark:
**MERGE READY**

Issue #64 

Ports over Yogstation's changeling stuff:
- Allowing you to DNA Sting/Absorb people whose DNA you already have are "refreshed" to their latest clothes/etc. Still costs the same, doesn't increase your absorbed count again;
- Costs/Effects tweaks - removes that which we didn't have in our codebase;

**TL;DR : Hallucinogenic Pathogen is a serious issue now and you are going to suffer if you don't heal it and it actually works now, because hallucination is increased at least to the point it starts hallucinating**

**Buffs the crap out of Hallucinogenic Pathogen** (Yogstation's version of Hallucination sting - a disease that spreads to crew) - the ability just plain sucked because it never got your hallucination counter up to 20 (where it actually starts hallucinating) and was cured WAY too fast on it's own. Plus, since it's a disease, you got resistance to it when you got cured, so...to remedy this:
- The increase in hallucination is controlled via **increment** that gets set in the code. Starts at 5 and then increases. For all intents and purposes - if you get a "hallucination message", it will increase the counter. The exception is Stage 3 where the increment can just increase randomly (since that's the heavyweight stage).
- Increment ramps up and controls by how much to increase your hallucination, but it will always increase it to at least 30/40/60 for Stages 1/2/3.
- The chances for reaching next stage have been adjusted: 1->2 = 3% chance; 2->3 = 2% chance; 3->4 = 1% chance.
- If your hallucination is > 500, I spare you from having it increased even more -> move to Stage 4 (the self-curing stage thing) and set it to 400. This is actually pretty damn easy to reach now.
- At Stage 4, it has 1% chance to return to Stage 3, but it does reset increment to 20. Otherwise, cure it automatically if your hallucination reaches 20.


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -229,12 +229,11 @@
 					var/end_T_descriptor = "<font color='#6b4400'>tile at [end_T.x], [end_T.y], [end_T.z] in area [get_area(end_T)]</font>"
 					add_logs(src, throwable_mob, "thrown", addition="from [start_T_descriptor] with the target [end_T_descriptor]")
 
-	if (I.prethrow_at(target))
-		return;
-
-	if(!(I && istype(I))) return //Grab processing has a chance of returning null
-
+	if(I && I.prethrow_at(target))
```
X-TheDark, line 235: This prevents runtimes when we try to throw a person at a wall due to the throwable fireball thing we have.

code/game/gamemodes/changeling/evolution_menu.dm
```diff
@@ -369,6 +369,13 @@ var/list/sting_paths
 			if(!mind.changeling.has_sting(S))
 				mind.changeling.purchasedpowers+=S
 			S.on_purchase(src)
+	//move any revive abilities to the bottom of the list, like people are used to.
+	var/revive_abilities = list()//there is only one regenerate abiliy at the moment, but there may be more in the future. Hence the list. Hopefully if more are added, the coders extend the current revive ability or add to this proc.
+	for(var/obj/effect/proc_holder/changeling/p in mind.changeling.purchasedpowers)
+		if(istype(p, /obj/effect/proc_holder/changeling/revive))
+			revive_abilities += p
+			mind.changeling.purchasedpowers -= p
+	mind.changeling.purchasedpowers += revive_abilities
```
Cruix: This is going to be handled in a better way when Oisin pulls some /tg/ fixes, you can remove it here.

X-TheDark: Ok.

Cruix:
You saw my comment too fast before I deleted it. That change isn't actually an issue, I thought it was part of a different change from the same commit that would have conflicted with the tg fix, but you only added the part where it moves the ability to the bottom of the list, so it's fine.

X-TheDark:
Conflicts resolved. This is ready to be merged.

PR #70, opened 2016-06-10
Ports YogPredators (Kn0ss0s) + additions

Super3222:
Additions for fun following this:
- If you try to pick up a predator weapon w/ being a predator it'll deal damage to you (and forcibly make you drop them).
- Predators get their own variant of combat gloves which allow them to toggle wristblades (that have a force of 20).
- Everytime a predator talks, it will play a clicking noise. (Thank you KMC2000)
- Predator Helmets will give the Predator a verb (in a tab called "Predator") which will toggle their helmets vision functions. It's a mix of minor night vision and thermals.
- Predators will also get an ability to "remove skull" in their Predaor Tab.
  This ability requires an aggressive grab (much like Changelings absorb ability) which will begin with the Predator plunging their dagger-like finger nails into the targets eyesockets. This'll weaken them for a very long time and deal forcibly blind them. Than, the Predator will take some more time to wrap their fingers around the inside of their targets brain (which'll weaken them for even longer), before taking more time to pull out their skull carrying their brain and spinal cord. In the end, spilling all of the targets organs across the floor and dealing 200 brute damage to the target.

More bugfixes and revisions will need to be made for Predators of course... these are definately not permanent revision.

But yeah, this is a pretty safe port. I had to sprite the limbs for the Predators because of the body part rework.


code/modules/mob/living/carbon/predator/predator_powers.dm
```diff
@@ -0,0 +1,68 @@
...
+///////////////////////
+
+/datum/action/innate/predator_remove_skull
+	name = "Remove Skull"
+	button_icon_state = "godspeak"
+	check_flags = AB_CHECK_CONSCIOUS
+	var/busy
+
+/datum/action/innate/predator_remove_skull/Activate()
+	if(!ispredator(usr)) qdel(src)
```
X-TheDark: Would you mind trying out this instead?

```
if(!ispredator(usr))
    Remove(usr)
```

This uses the functions defined for removing action as per https://github.com/yogstation13/-tg-station-yog-rebase/blob/master/code/datums/action.dm#L43

Not an issue in the slightest, but I'd like to have our code try to be consistent.

Super3222: I see, heh. I'm still trying to learn more with action buttons.Thanks

code/__DEFINES/is_helpers.dm
```diff
@@ -78,7 +78,7 @@
 
 #define ismovableatom(A) (istype(A, /atom/movable))
 
-#define isobj(A) istype(A, /obj) //override the byond proc because it returns true on children of /atom/movable that aren't objs
```
oisin100, line 81: You should leave this here.

code/modules/mob/living/carbon/predator/predator_powers.dm
```diff
@@ -0,0 +1,69 @@
...
+	if(!src.pulling || !iscarbon(src.pulling))
+		src << "<span class='warning'>You must hold the target in a head-lock to remove their skull!</span>"
+		return
+	if(src.grab_state <= GRAB_NECK)
+		src << "<span class='warning'>You have to carry a tighter grip on the target to remove their skull!</span>"
+		return
+
+	var/mob/living/carbon/ptarget = src.pulling
+
+	src.visible_message("<span class='danger'>[src] has plunged their sharp, dagger-like fingers into [ptarget]'s eyes!</span>", \
```
oisin100: Should the second argument not be something like `<span class='danger'>You plunge your sharp, dagger-like fingers into [ptarget]'s eyes!</span>`

oisin100:
Just checking, How did you port the icon files? Using copy and paste or using the inbuilt DreamMaker export/import functions?

Super3222:
Time-Green himself already added the icon files. I only added a seperate .dmi file containing all of the predator icons within it, just in case he missed some.

I also had to do the limbs by myself.

PR #73, opened 2016-06-10
Hopefully makes Travis stop freaking out

Cruix:
Fixes plant analyzer paths on unused maps.

No I did not type in the branch name thinking it was the commit description, why would you even think that?


Cruix:
Success.

oisin100:
I ignored the _maps directory(Except for the tgstation map) and forgot to update the paths in these maps because of it.
Thanks!

Issue #71, opened 2016-06-10
I Want To Be The Port : Shadowlings


X-TheDark:
I'll take a look at shadowlings, if noone minds.

FluffySurvivor:
Remove slings pls

X-TheDark:
Is there anything wrong with them, @FluffySurvivor ? If so, I'd like to hear it, as I don't play alot these days (and have, in fact, to this day, never played a primary roundstart antag ever).

timkoster1:
Ignore fluffy, shadowlings r awesome

X-TheDark:
Ready to be merged.

PR #75, opened 2016-06-12
Ports Yogstation's shadowlings

X-TheDark:
**MERGE READY**

Issue : #71

**Buffs**
- Movement in dark is basically incorporeal movement where you move at sanic speeds. The difference between move_delay of 0 and 1 is incredibly noticeable.

**Changes (or things kept from /tg/)**
- Shadowlings can Empower thralls with the revive thing (aka "The ability noone uses");
- Shadowlings can delay shuttle by 15 minutes ONCE;

**The great ability shuffle (aka /tg/'s changes to when you unlock what)**
- Screech acquired at 3 thralls (instead of Yog's 7);
- Blind Smoke acquired at 5 (instead of Yog's 3);
- Drain Life @ 7 (instead of Yog's 5);

As a note, our shadowlings have permanent shadow walk in between empty spaces that are dark.


X-TheDark:
In this commit, shadowling speed in light decreased to our shadowling levels (speedmod = 4), but in the dark they basically gain incorporeal movement without any delay whatsoever - think incorporeal movement (a buff).

X-TheDark:
In this commit:
- Ports Thrall Organ + dethralling as a organ manipulation (instead of it's own surgery);
- Removes Enthrall from Shadowling starting spells. Hatch to get it. (No point in having it if you can't use it - you can't do it even on Yogs);
- Ports a crapton of Icons from /tg/, as we seem to be having some issues with things like security (red) uniforms as well as Shadowling item stuff;

**This is now ready to be merged**

code/modules/client/client_procs.dm
```diff
@@ -178,12 +178,14 @@ var/next_external_rsc = 0
 	directory[ckey] = src
 
 	//Admin Authorisation
+	
 	var/localhost_addresses = list("127.0.0.1", "::1")
 	if(address && (address in localhost_addresses))
-		var/datum/admin_rank/localhost_rank = new("!localhost!", 65535)
+		var/datum/admin_rank/localhost_rank = new("!localhost!", R_MAXPERMISSION - 1 - R_NOJOIN)
```
X-TheDark, line 184: This is needed because we have the R_NOJOIN "admin right" that prevents you from joining.

code/game/gamemodes/shadowling/shadowling_abilities.dm
```diff
@@ -768,14 +941,15 @@
...
 
-/obj/effect/proc_holder/spell/targeted/annihilate/cast(list/targets,mob/living/simple_animal/ascendant_shadowling/user = usr)
+/obj/effect/proc_holder/spell/targeted/shadow/annihilate/cast(list/targets, mob/living/simple_animal/ascendant_shadowling/user = usr)
 	if(user.incorporeal_move)
 		user << "<span class='warning'>You are not in the same plane of existence. Unphase first.</span>"
 		revert_cast()
 		return
 	for(var/mob/living/boom in targets)
-		if(is_shadow(boom)) //Used to not work on thralls. Now it does so you can PUNISH THEM LIKE THE WRATHFUL GOD YOU ARE.
+		if(is_shadow_or_thrall(boom))
```
oisin100, line 952: Why :(

X-TheDark, line 952: This actually doesn't even matter, because @MacHac's targetting changes in https://github.com/yogstation13/yogstation/pull/684 don't even allow this in the first place. This check is nothing more than sanity in case the targetting is ever changed to allow allied units (which it currently doesn't).

Plus, again, this would be an administrative issue, not a coder one. Do we allow "griefing" of thralls after shadowlings have ascended? If the admins would say this is ok, I can change this to allow thrall targetting.

ShadowDeath6:
Screech is massively powerful against a good counter to early shadowlings(borgs). Against its unlock being decreased so much

X-TheDark:
Yeah, more feedback on the ability shuffle would be appreciated.

PR #84, opened 2016-06-14
Borer Port (the real one)

MacHac:
Sorry this took so long to get fixed, I'm a huge slacker.

This ports borers as they currently exist on the server, including recent changes by ShadowDeath6.


code/modules/mob/living/simple_animal/borer/borer.dm
```diff
@@ -0,0 +1,278 @@
...
+	borer_chems += /datum/borer_chem/mannitol
+	borer_chems += /datum/borer_chem/bicaridine
+	borer_chems += /datum/borer_chem/kelotane
+	borer_chems += /datum/borer_chem/charcoal
+	borer_chems += /datum/borer_chem/ephedrine
+	borer_chems += /datum/borer_chem/leporazine
+	borer_chems += /datum/borer_chem/perfluorodecalin
+	borer_chems += /datum/borer_chem/spacedrugs
+	borer_chems += /datum/borer_chem/mutadone
+//	borer_chems += /datum/borer_chem/creagent
```
oisin100: Actually, Could you re-add this please?

code/modules/mob/living/simple_animal/borer/borer_chemicals.dm
```diff
@@ -0,0 +1,66 @@
...
+	//needed_influence = 50
+	//influence_change = -5
+	chemuse = 75
+
+/datum/borer_chem/mutadone
+	chemname = "mutadone"
+	chem_desc = "Heals your genetic mutations."
+	chemuse = 100
+
+/*/datum/borer_chem/creagent
```
oisin100: And uncomment this.

code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -0,0 +1,415 @@
...
+
+	if(CanInfect(H))
+		src << "<span class='warning'>You slither up [H] and begin probing at their ear canal...</span>"
+		src.layer = MOB_LAYER
+		//H << "<span class='userdanger'>You feel something trying to enter your ear...</span>"
+		if(!do_after(src, 30, 5, 0, H))
+			src << "<span class='warning'>As [H] moves away, you are dislodged and fall to the ground.</span>"
+			return
+
+		if(!H || !src) return
```
oisin100: Just thinking about it, We could really do with another check to make sure the host doesn't have a borer in them. Just change this to `if(!H || !src || H.borer) return`

code/modules/mob/living/simple_animal/borer/borer.dm
```diff
@@ -0,0 +1,278 @@
...
+	if(controlling)
+		detatch()
+
+	src.loc = get_turf(victim)
+
+	victim.borer = null
+	victim = null
+//	influence = 0
+
+/mob/living/simple_animal/borer/proc/transfer_personality(var/client/candidate)
```
oisin100, line 206: While your at it could you set the borer ckey to the new host too?

code/modules/mob/living/simple_animal/borer/borer.dm
ShadowDeath6: Why do you want colorful reagent

oisin100: Because its fun? And non-lethal.

ShadowDeath6: 1. It isn't realistic for borers to have and should only be gotten through chemistry memes. 
2. It's annoying as fuck to fix if you don't want it and the borer does, since it can spam it
3. It's an obvious tell that a person has a borer, which will get it removed as soon as somebody notices they're randomly changing colors

Maybe if it was adminbus meme only, but it's not realistic for borers to have, and the state they're in, RP wise, is stable enough without them spamming creagent into a host who doesn't want it or having someone who begs the borer to give them the rainbow.

oisin100: > It isn't realistic for borers to have and should only be gotten through chemistry memes.

This is SS13, Nothing is realistic.

> It's annoying as fuck to fix if you don't want it and the borer does, since it can spam it

So? They are borers, Mindslugs that dig into your head, They are meant to be annoying. I agree the price should be increased however, 50-75

> It's an obvious tell that a person has a borer, which will get it removed as soon as somebody notices they're randomly changing colors

Since we dont have influence anymore it can be used anytime, it used to be a endgame thing that was broken.

ShadowDeath6: still going to be against it because memes in a serious feature aren't welcome in my book

Issue #18, opened 2016-06-01
Port: Minebot improvements


Cruix:

ToGWtF:
https://github.com/yogstation13/yogstation/pull/1088
This was a direct port from /tg/ already, except for the "Inactive mode".

https://github.com/yogstation13/yogstation/pull/1240/files
These are still probably worth porting

PR #87, opened 2016-06-15
Ports Yog's minebot improvements.

Cruix:
Basically just idle mode and emag mode. A sentient minebot cannot be emagged, and you cannot make an emagged minebot sentient.


Cruix:
(FIxes #18)

PR #96, opened 2016-06-18
Fixes Travis

X-TheDark:
It now compiles the renamed yogstation.dme


oisin100:
Hehe, Forgot all about that. Thanks.

PR #104, opened 2016-06-19
Predator Updates

Super3222:
In this revision:
- Deletes old comments
- Makes Predators spells not lazy by rewriting them to follow 'procholder/spell' paths
- Rewrites the gloves code so they can be removed now instead of being NODROP and so you don't need to use awkward methods to retract the wristblades
- Fixes a bug with the combatstick
- Reworks the Predators biomask, so toggling it (via action button) will activate thermals + nightvision instead of using a verb.
- Nerfs Predators so they need to breath and have blood. However they'll be kept virus immune.

### DIFF
--- a/code/modules/mob/living/carbon/predator/predator.dm
+++ b/code/modules/mob/living/carbon/predator/predator.dm
@@ -9,7 +9,6 @@
 	languages_understood = HUMAN | PREDATOR
 	gender = NEUTER
 	ventcrawler = 0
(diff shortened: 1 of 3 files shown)


ParagonPrime:
> -Nerfs Predators so they need to breath and have blood.

[:^)](https://youtu.be/eNr0WXQ3Ho4?t=6s)

Super3222:
I forgot to hit the space bar there.

X-TheDark:
Code looks good.

PR #107, opened 2016-06-21
Ports bomb collars

Super3222:
@Oisin100 you'll need to put the bomb collar locker back on the map (armory)


X-TheDark:
Code looks good.

I'll add them to the armory after this is merged.

PR #114, opened 2016-06-25
Uprights the bar stools

X-TheDark:
Bar stools were put in as roundstart items (can't sit on them, would have to pick them up then click on them in-hand), rather than structures, this has been fixed.

Was:
stools_1

Became:
stools_2


GrayRachnid:
mybad

PR #116, opened 2016-06-25
Ports: Flying over tables in zero gravity with a jetpack

Super3222:
I decided to allow people to pass over tables when they have a jetpack ON, on their BACK, and a gravity check fails.

https://github.com/yogstation13/yogstation-old/pull/266
https://github.com/tgstation/tgstation/pull/15866

:cl: Super3222
tweak: You can now fly over tables with a jetpack when gravity is down!
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/tanks/jetpack.dm
+++ b/code/game/objects/items/weapons/tanks/jetpack.dm
@@ -35,22 +35,22 @@
 		return
 
 	if(!on)
-		turn_on(user)
+		turn_on()
 		user << "<span class='notice'>You turn the jetpack on.</span>"
 	else
-		turn_off(user)
+		turn_off()
 		user << "<span class='notice'>You turn the jetpack off.</span>"
 	for(var/X in actions)
 		var/datum/action/A = X
 		A.UpdateButtonIcon()
 
 
-/obj/item/weapon/tank/jetpack/proc/turn_on(mob/user)
+/obj/item/weapon/tank/jetpack/proc/turn_on()
 	on = TRUE
 	icon_state = "[initial(icon_state)]-on"
 	ion_trail.start()
 
-/obj/item/weapon/tank/jetpack/proc/turn_off(mob/user)
+/obj/item/weapon/tank/jetpack/proc/turn_off()
 	on = FALSE
 	stabilizers = FALSE
 	icon_state = initial(icon_state)
@@ -60,12 +60,12 @@
 	if(!on)
 		return
 	if((num < 0.005 || air_contents.total_moles() < num))
-		turn_off(user)
+		turn_off()
 		return
 
 	var/datum/gas_mixture/removed = air_contents.remove(num)
 	if(removed.total_moles() < 0.005)
-		turn_off(user)
+		turn_off()
 		return
 
 	var/turf/T = get_turf(user)
--- a/code/game/objects/items/weapons/tanks/jetpack.dm
+++ b/code/game/objects/items/weapons/tanks/jetpack.dm
@@ -35,23 +35,33 @@
 		return
 
 	if(!on)
-		turn_on()
+		turn_on(user)
 		user << "<span class='notice'>You turn the jetpack on.</span>"
 	else
-		turn_off()
+		turn_off(user)
 		user << "<span class='notice'>You turn the jetpack off.</span>"
 	for(var/X in actions)
 		var/datum/action/A = X
 		A.UpdateButtonIcon()
 
 
-/obj/item/weapon/tank/jetpack/proc/turn_on()
+/obj/item/weapon/tank/jetpack/proc/turn_on(mob/user)
 	on = TRUE
+	var/mob/living/C = user
+
+	if(C.pass_flags & JETPACKTABLE) // a check to make sure we don't start with it again.
+		C.pass_flags &= ~JETPACKTABLE
+
+	if (!has_gravity(user))
+		C.pass_flags |= JETPACKTABLE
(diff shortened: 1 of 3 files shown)


ParagonPrime:
I'd be partial to having zero G just flat out negate tables, but maybe that's too far.

code/game/objects/items/weapons/tanks/jetpack.dm
```diff
@@ -35,23 +35,33 @@
...
-/obj/item/weapon/tank/jetpack/proc/turn_on()
+/obj/item/weapon/tank/jetpack/proc/turn_on(mob/user)
 	on = TRUE
+	var/mob/living/C = user
+
+	if(C.pass_flags & JETPACKTABLE) // a check to make sure we don't start with it again.
+		C.pass_flags &= ~JETPACKTABLE
+
+	if (!has_gravity(user))
+		C.pass_flags |= JETPACKTABLE
```
X-TheDark: Won't this just check for gravity ONCE, when the jetpack is turned on, meaning that even if gravity comes on later, you still retain the tablepassing. Would you mind testing this with gravity being turned on AFTER you turn on the jetpack?

After you report the results of your testing (if it turns out true), I'll give you some things I think you need to move around/change.

Super3222: I see what you're getting at...
I gave it a test, and it seems like you're right. So after turning gravity off and then back on again (leaving my jetpack toggled on)... It doesn't check for gravity, thus it still the player still carries `JETPACKTABLE` flag which allows you to still fly over tables.

X-TheDark: To fix this, move the `!has_gravity(user)` check to somewhere after line 136 in tables_racks.dm (the new if you made). Add it as another && like `if(jetpacktable && jetpacktable.on && !has_gravity(C))`. Leave this as just `C.pass_flags |= JETPACKTABLE` .

Also, pretty sure the if in lines 52-53 is unnecessary. Why unset the flag if we are just going to set it anyway using bitflag operations?

X-TheDark:
Also, what about dropping the active jetpack? Is there a check/something to turn the jetpack off/remove the bitflag when `dropped()` is called?

GrayRachnid:
Mining cyborgs also have jetpacks

Super3222:
> Also, what about dropping the active jetpack? Is there a check/something to turn the jetpack off/remove the bitflag when dropped() is called?

No, not currently. I made a check in tables_racks.dm right after the check for JETPACKTABLEwhich makes sure you have a jetpack on and it's toggled on.

ParagonPrime:
Set the mineborg jetpack to be a special type that doesn't have this flag.

Super3222:
> Also, what about dropping the active jetpack? Is there a check/something to turn the jetpack off/remove the bitflag when dropped() is called?

No, not currently, but I'll add a check eventually.
The reason for that is, in tables_racks.dm, it'll only allow you over the table if you specifically have a jetpack which is already turned on.

The check on 52-53 is so that someone doesn't turn on their jetpack, take it off, put on another jetpack, turn that one on, and then break themselves (where they'll be able to walk through airlocks and a lot of other broken shit).

X-TheDark:
Wait, how does that issue even occur? I thought bitflag operation `|=` in BYOND just set the flag (if it is unset) or not do anything if the flag is already set. It's not the same as `+=`, which is the one that does what you described.

Super3222:
Ah, right! I was referring to `+=`. My fault. It shouldn't happen

Super3222:
@GrayRachnid @ParagonPrime 

All borgs have jetpacks. They just need an ion thruster module from robotics.
However, they won't be able to jump tables.

code/game/objects/items/weapons/tanks/jetpack.dm
```diff
@@ -35,23 +35,26 @@
...
 
-/obj/item/weapon/tank/jetpack/proc/turn_on()
+/obj/item/weapon/tank/jetpack/proc/turn_on(mob/user)
 	on = TRUE
+	user.pass_flags |= JETPACKTABLE
 	icon_state = "[initial(icon_state)]-on"
 	ion_trail.start()
 
-/obj/item/weapon/tank/jetpack/proc/turn_off()
+/obj/item/weapon/tank/jetpack/proc/turn_off(mob/user)
```
oisin100: Just wondering, Is this called when a player takes the jetpack off? If its not then bad things are going to happen.

Super3222: I'm going to rewrite this, and not use bitflags. I've thought up an easier method.

oisin100: Looked back at the code, I realized you had a check to see if they had a jetpack. My bad.

code/game/objects/items/weapons/tanks/jetpack.dm
```diff
@@ -35,22 +35,22 @@
...
-		turn_off()
+		turn_off(user)
 		user << "<span class='notice'>You turn the jetpack off.</span>"
 	for(var/X in actions)
 		var/datum/action/A = X
 		A.UpdateButtonIcon()
 
 
-/obj/item/weapon/tank/jetpack/proc/turn_on()
+/obj/item/weapon/tank/jetpack/proc/turn_on(mob/user)
```
oisin100: Not much point in the mob/user var anymore

Super3222: Right

PR #119, opened 2016-06-26
Makes humans and lizards roundstart species

Cruix:


GrayRachnid:
lizards are also missing I believe.

PR #121, opened 2016-06-26
Deletes the honkbots off the map

Super3222:
Fixes botany's vent.
Deletes honkbots off the map, apparently they are broken.


X-TheDark:
FUCK, RIGHT.

/tg/ repathed their bots into simple_mobs...I didn't do that with the honkbot.

X-TheDark:
Reason for this is that honkbots are probably not repathed to simple_mob/ type (`/mob/living/simple_animal/bot/`)

Super3222:
Wait a minute, what the hell? Botany fixes aren't in this.

PR #125, opened 2016-06-26
Fixes admin chat not being yellow and admin log being yellow

X-TheDark:


X-TheDark:
Ok, found the issues with admin logs not being the correct color

PR #131, opened 2016-06-27
Fixes survival capsules

X-TheDark:
Fixes issue : https://github.com/yogstation13/yogstation/issues/1297

We didn't have the template file. Just needs to be added to the server mapfiles, don't think you even need to compile anything.


MacHac:
:+1: :+1: :+1: mining lavaland is a huge pain in the ass without this
speedmerge

Issue #80, opened 2016-06-13
Remove the mini-hypospray.


GrayRachnid:
It is completely unbalanced and should not be in the rebase.

MacHac:
:+1:
With patches being just all-around better, and lavaland not needing hardsuits, it's really not neccesary.

PR #147, opened 2016-06-27
Officially removes mini hypospray

SKR6DarkVoid:
Officially removes the mini hypospray from our code, as it is redundant and useless, since our miners got some new bad ass suits.

_I have backup code if we ever need to re-port it._


SKR6DarkVoid:
This is referring to issue #80 We no longer need this, I do have backup code if we ever choose to re-port it.

Issue #160, opened 2016-06-28
Security transcievers are missing


EMT321:
[Note]: #160 We no longer have security transcievers, making it that more difficult for security to communicate in the event of a blackout on a secure channel.

Reproduction: Log in.

Solution: Re-add security transcievers.

PR #171, opened 2016-06-28
Adds security transceivers (non-lazy edition)

X-TheDark:
Issue : #160

Adds security transceivers to lockers and the map. In addition, Warden/HoS keep the bounced station radios as well too! (Previously, they were replaced with security radios, I just added them, instead of replacing)


oisin100:
Dank.

EMT321:
Thank you very much for doing this!

PR #173, opened 2016-06-28
(Library) Book Printing and Bookbag Tweaks

Cabledragon:


oisin100:
I imagine the 6 second delay between printing books is to prevent people spamming the database with print queries. 3 Seconds should be fine. _Famous last words_

PR #182, opened 2016-06-28
Library Printing Queue

Cabledragon:
When you print a book from the external archive, it is now added to a queue. This allows multiple books to be added at once so the librarian can do other things while the list of books prints!

### DIFF
--- a/code/modules/library/lib_machines.dm
+++ b/code/modules/library/lib_machines.dm
@@ -178,6 +178,8 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	var/page = 1	//current page of the external archives
 	var/print_busy = 0 // LOL NO SPAM (1 minute delay) -- Doohl
 	var/list/print_queue = list()
+	var/max_print_queue_len = 50;
+	var/clearprintqueue = 0;
 
 /obj/machinery/computer/libraryconsole/bookmanagement/proc/build_library_menu()
 	if(libcomp_menu)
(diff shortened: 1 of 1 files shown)


code/modules/library/lib_machines.dm
```diff
@@ -299,6 +300,34 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	popup.set_title_image(user.browse_rsc_icon(src.icon, src.icon_state))
 	popup.open()
 
+/obj/machinery/computer/libraryconsole/bookmanagement/proc/print_book(book_id)
+	if(!book_id)
+		return
+	establish_db_connection()
+	if(!dbcon.IsConnected())
+		alert("Connection to Archive has been severed. Aborting.")
```
Cabledragon: need to add a return after this line.

code/modules/library/lib_machines.dm
```diff
@@ -414,30 +442,10 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 				href_list["targetid"] = orderid
 	if(href_list["targetid"])
 		var/sqlid = sanitizeSQL(href_list["targetid"])
```
oisin100: We should probably sanitize this in the actual proc now, instead of here.

code/modules/library/lib_machines.dm
```diff
@@ -299,6 +300,35 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	popup.set_title_image(user.browse_rsc_icon(src.icon, src.icon_state))
 	popup.open()
 
+/obj/machinery/computer/libraryconsole/bookmanagement/proc/print_book(book_id)
+	if(!book_id)
+		return
+	establish_db_connection()
```
oisin100, line 318: We can sanitize it with `book_id = sanitizeSQL(book_id)`

Cabledragon, line 318: Uhh.. so remove the other sanitize, pass (href_list["targetid"]) to the proc and sanitize book_id 
after establish_db_connection?

oisin100, line 318: Yea, Well line 445 would become `print_queue += href_list["targetid"]`

PR #184, opened 2016-06-28
Fixes the infinite exoskeleton bug

Super3222:
Ignore the branch name, I was "in a rush".

Bug: Pull on an exoskeleton, click on it using a MMI (with an acceptable brain), borg is created but the exoskeleton doesn't disappear.

The bug happened because the skeletons loc was stuck because the user was still pulling.
I fixed it by ending anyone's pull when the MMI would transfer into the borg.

> ez fix
> merge token pls


oisin100:
> fixles

PR #188, opened 2016-06-29
OOC tags including Donor and Admins/Coders

Super3222:
Non broken version.

### DIFF
--- a/code/game/verbs/ooc.dm
+++ b/code/game/verbs/ooc.dm
@@ -69,7 +69,6 @@
 					if(check_rights_for(src, R_ADMIN))
 						if(src.holder.rank.name == ("AdminObserver"||"CoderInitiate"))
 							C << "<span class='adminobserverooc'><span class='prefix'>OOC:</span> <EM>[keyname][holder.fakekey ? "/([holder.fakekey])" : ""]:</EM> <span class='message'>[msg]</span></span>"
-							return
 						else
 							var/client_rank = find_admin_rank(src)
 							if(!client_rank)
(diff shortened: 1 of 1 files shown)


code/game/verbs/ooc.dm
```diff
@@ -53,16 +53,28 @@
...
 
 	for(var/client/C in clients)
 		if(C.prefs.chat_toggles & CHAT_OOC)
 			if(holder)
 				if(!holder.fakekey || C.holder)
 					if(check_rights_for(src, R_ADMIN))
-						C << "<span class='adminooc'>[config.allow_admin_ooccolor && prefs.ooccolor ? "<font color=[prefs.ooccolor]>" :"" ]<span class='prefix'>OOC:</span> <EM>[keyname][holder.fakekey ? "/([holder.fakekey])" : ""]:</EM> <span class='message'>[msg]</span></span></font>"
+						if(src.holder.rank.name == ("AdminObserver"||"CoderInitiate"))
+							C << "<span class='adminobserverooc'><span class='prefix'>OOC:</span> <EM>[keyname][holder.fakekey ? "/([holder.fakekey])" : ""]:</EM> <span class='message'>[msg]</span></span>"
+							return
```
X-TheDark: This won't work for this reason:
Basically, the entire `for(client/C in clients)` thing is the code that sends our OOC message to every client, if you put a `return` here, if the rank is AdminObserver or CoderInitiate, it will only send it to ONE client (because it `return`'s from the proc after sending it ONCE). This return needs to be removed.

oisin100: In future, If you want to use `return;` inside a loop without exiting the loop use `continue`

Super3222: I see.

oisin100:
This is VERY Hacky, But i suppose theres not much you can do about that.

Issue #149, opened 2016-06-27
Antag status is being given out regardless of preference.


GrayRachnid:
People are getting antag roles with their preference of it being turned off.

timkoster1:
DB issue

PR #190, opened 2016-06-29
Probably fixed antag prefs not working.

Cruix:
This is difficult to test on a local server, but I am fairly certain this fixes an obscure order-of-operations bug that was causing everyone's prefs to appear to be off for everything, causing the game to select from everyone since it did not have enough candidates.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -379,7 +379,7 @@
 
 	for(var/mob/new_player/player in players)
 		if(player.client && player.ready)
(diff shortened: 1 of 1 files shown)


X-TheDark:
That's such an amazingly obscure thing. My bad, I'm the one who ported the `!(player.mind.quiet_round)` part, I completely forgot that you can use operators such as `list1|list2` and such on the same line as `in` (with that operator being processed first).

Cruix:
(bit late but fixes #149)

PR #195, opened 2016-06-29
Fixes swarmer event not spawning properly

X-TheDark:
**No, this is a legit and proper bugfix, I don't care what you say.**

/tg/ have the gateway where we have the Clerk shop. That's what's used to spawn in swarmer shells. 

Now it's replaced with gravity generator, because, hey, why the hell not, right, what's the worst that could happen? Only works if the gravity generator is currently on (aka, if station has gravity).

Ghosts get notified the same way as if golem shells/drones being made.

### DIFF
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer_event.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer_event.dm
@@ -13,16 +13,18 @@
 /datum/round_event/spawn_swarmer/start()
 	if(find_swarmer())
 		return 0
-	if(!the_gateway)
+	if(!gravity_generators.len)
 		return 0
-	new /obj/item/device/unactivated_swarmer(get_turf(the_gateway))
+	var/turf/spawn_loc = get_turf(pick(gravity_generators["[ZLEVEL_STATION]"]))
+	if(!spawn_loc)
+		return 0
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
should have made it spawn in the clerk's backroom, smhtbhfam

MacHac:
> not removing clerk and replacing it with gateway room

pls

Cruix:
"We have shipped an unusual order directly to your gift shop's back room via bluespace parcel service. When our distribution center recieved it, the 'shipment sent' date was around 20,000 years ago, so please check the contents to see if they have expired."

X-TheDark:
Clerk's backroom doesn't have any vents for swarmers to ventcrawl away, though.

oisin100:
So everyone is O.K with the gravity generator being the first thing the swarmers eat?

X-TheDark:
Swarmers can't eat the Gravity generator, it shows the blue animation thing, but the generator never breaks down and since they can't eat APCs or SMES'es, they can't disable gravity, in addition, they can't break most of the walls in there, since that would be classified as hull breach (and it literally doesn't let them do that), so the only real benefit are the Vent placement and the fact that it's out of the way.

ShadowDeath6:
just let swarmers slip under doors like drones and we won't need to worry about vents :^)

X-TheDark:
Oh god, no, not happening.

PR #198, opened 2016-06-29
blob stuff 

ShadowDeath6:
ports all the changes we had to blob chemicals
also makes flammable goo ignite on any burn damage
also makes electromagnetic web make pulses on attacking instead of using emp_act
also, blob splitting is back

keep in mind that I couldn't feasibly check the blob splitting since my computer has a really hard time putting two keys on a local server

@X-TheDark since you dealt with blob splitting and the like

:cl: ShadowDeath6
rscadd: Re-added blob splitting
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/powers.dm
+++ b/code/game/gamemodes/blob/powers.dm
@@ -137,7 +137,7 @@
 
 /mob/camera/blob/verb/create_blobbernaut()
 	set category = "Blob"
-	set name = "Create Blobbernaut (40)"
+	set name = "Create Blobbernaut (20)"
 	set desc = "Create a powerful blobbernaut which will attack enemies."
 	var/turf/T = get_turf(src)
 	var/obj/effect/blob/factory/B = locate(/obj/effect/blob/factory) in T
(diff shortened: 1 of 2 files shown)


FluffySurvivor:
sentient blobbernauts is good, keep it

and nerf the new blobchem pls

GrayRachnid:
Yeah, don't remove sentient blobbernauts

ShadowDeath6:
new blobchem is actually weaker than the other ones
also sentient blobbernauts are stupidly powerful and you can spam them with no drawback

GrayRachnid:
There is drawback, you can only have 1 blobbernaught per factory, and you can only have one factory per node(Or within 7 tiles of another factory). 
They're also not that powerful, their health stinks and if they leave the blob area they lose health a lot faster when taking damage.

ShadowDeath6:
Dunno, sentient blobbernauts AND more overminds might be a tad powerful, I'll wait for more feedback before changing anything.
cue requires feedback tag xddd

Time-Green:
You can log in on guest if you need multiple keys.

ShadowDeath6:
> you can log in on guest if you need multiple keys
> 
> > > "keep in mind that I couldn't feasibly check the blob splitting since my computer has a really hard time putting two keys on a local server"

it's more spec issues, my computer shits out on me

PR #199, opened 2016-06-29
Fixes chameleon jumpsuit adjustability settings

X-TheDark:
Prevents you from being able to adjust unadjustable jumpsuits chameleon is disguised as, causing icon issues. Also blacklists some useless/broken types.

**Before any coders even ask:** No, as strange as it may seem, `istype(picked_item, /obj/item/clothing/under)` (or even `/obj/item/clothing`, for that matter) _doesn't work_ (or `istype(I...)` or `istype(PCL...)`). As in, it returns 0, even if the picked_item IS (and can be typecast into) the appropriate type. I have no idea why it is the way it is, but it is the way it is. `picked_item` will have to be assumed to be the same type as the object calling the button (which is very risky, I know, but I have no idea how to fix this in any other way).

### DIFF
--- a/code/modules/clothing/chameleon.dm
+++ b/code/modules/clothing/chameleon.dm
@@ -75,7 +75,7 @@
 	if(button)
 		button.name = "Change [chameleon_name] Appearance"
 	chameleon_blacklist += target.type
-	var/list/temp_list = typesof(chameleon_type)
+	var/list/temp_list = subtypesof(chameleon_type)
 	for(var/V in temp_list - (chameleon_blacklist))
 		chameleon_list += V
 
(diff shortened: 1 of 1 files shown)


Cruix:
picked_item is not actually an instance of type obj/item, it is a typepath var. Some coders inexplicably type typepath vars as the most general type they could be, while in reality they should be defined as plain vars. Using ispath() instead of istype() should work for what you want.

The fact picked_item is a type rather than an instance is why every point where you are pulling a var out of it uses intial(), which uses forbidden voodoo magic to pull var values out of types.

<I>Arguably</I> this use of initial() gives you a reason to give types to typepath vars, but I personally do not like this practice. Just cast the typepath whenever you need to use your voodoo.

X-TheDark:
Ah, that makes sense, thanks for the answer, I see. In your opinion, should I include `ispath()` checks or is the assumption that the `picked_item` will always have the same type as `target` fair?

Cruix:
Assuming is probably fine.

Issue #196, opened 2016-06-29
You can't build communication consoles(The one used to call the shuttle)


GrayRachnid:
Says the frame doesn't fit with the board or some shit.

Super3222:
On it

PR #201, opened 2016-06-29
Repathes '/obj/item/weapon/circuitboard/cooldown_holder' 

Super3222:
Repathes '/obj/item/weapon/circuitboard/cooldown_holder' to '/obj/item/weapon/circuitboard/computer/cooldown_holder'.

I realize that it could have been much easier to create a check for '/obj/item/weapon/circuitboard/cooldown_holder', but I decided to just repath it to pass some time.
And then again to be clear, computers won't always be the only things that run on a cooldown. I'll probably just add a check in the future and reverse this.
#196


_maps/map_files/TgStation/tgstation.2.1.3.dmm
```diff
@@ -46002,7 +46002,7 @@
 	pixel_x = 2;
 	pixel_y = -2
 	},
-/obj/item/weapon/circuitboard/cooldown_holder/computer/communications{
+/obj/item/weapon/circuitboard/computer/cooldown_holder{
```
X-TheDark, line 46005: /obj/item/weapon/circuitboard/computer/cooldown_holder/communications

_maps/map_files/TgStation/tgstation.2.1.3.dmm
```diff
@@ -46002,10 +46003,6 @@
 	pixel_x = 2;
 	pixel_y = -2
 	},
-/obj/item/weapon/circuitboard/cooldown_holder/computer/communications{
-	pixel_x = 5;
```
oisin100, line 46006: Assuming these are the comms  consoles in the tech storage, dont you want the pixel_x/y to be 5 and -5 so they dont overlap with the other boards?

PR #202, opened 2016-06-29
Made water and space lube tiles more visible.

### LINKED ISSUE #174: Water tiles are no longer as visible as they have been.
The tiles are set to TG's visibility, about 3x less than what we had.

Solution: Re-add our own overlay.

Cruix:
Fixes #174


X-TheDark:
Looks good, just changes the alpha of `static_wet` and `wet` icon_states.

Groudonmaster:
pls no not again ;___;

PR #205, opened 2016-06-30
Adds emergency flares to standard survival boxes

Super3222:
Emergency flares are weaker then normal flares.
Security get heavy-duty flares which have more fuel than the normal flare.

Remember that this isn't anything like an intentional hit on shadowlings. The only thing I'm doing here is increasing the amount of equipment in a survival box. It already has equipment for emergencies, but still it lacks some (via a quick light source. I also realize your PDA has a light).
#### Changelog

:cl: Super3222
rscadd: Nanotrasen has funded the production of emergencyflares for all of their employees. 
rscadd: Security has been distributed heavy-duty flares which carry more fuel than the standard addition.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -196,10 +196,12 @@ obj/item/device/flashlight/lamp/bananalamp
 	var/fuel = 0
 	var/on_damage = 7
 	var/produce_heat = 1500
+	var/frng_min = 800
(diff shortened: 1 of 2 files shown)


oisin100:
How long does an emergency flare last? 400-800 ticks? Seems like a bit too much, 400 ticks is around 600 seconds.

Super3222:
Hmm, I'll definitely reduce it then. It's meant to be a short-lived flare.

Issue #236, opened 2016-07-01
Smilescreen mediborg doesn't have a transform animation


ShadowDeath6:
title, this was fixed on old code but wasn't ported over probably

Super3222:
Port it, @ShadowDeath6

ShadowDeath6:
too busy doing more important things :^)

PR #251, opened 2016-07-02
Smile Screen Medborgs now have a robot transformation.

Super3222:
#236

Smilescreen medborgs now have a robot transformation.
They also lose their :^ ) when dead.


GrayRachnid:
> They also lose their :^ ) when dead.

Madman

PR #252, opened 2016-07-02
Making miners' lives easier and encouraging them to do their job

ShadowDeath6:
Makes the miner hardsuit have a use, it's now ashproof and fireproof. Price has been doubled to 4000 points to reflect this. 
HOPEFULLY this actually makes miners want to mine ore for science instead of exploring for the entire shift, and 4000 points means miners need to bring in a good amount of ore if they want to be ashproof. 
Reminder that this does not protect against lava in any way, but it protects against ash and plasma fire. 

Also, makes the cursed heart drop from necropolis chests actually worth getting, as it's now the wizard version which heals you every pump and is a bit more lenient on the death part. One, this makes an item found from killing a difficult tendril worth using (unless it's a legion tendril I guess), while still keeping with the challenge of lavaland since you need to multitask pump for your heals. There are more powerful items than it, it certainly wouldn't be unbalanced.

:cl: ShadowDeath6
rscadd: NanoTrasen has realized that their hardsuits could be put to better use in the newfound Lavaland, and they have adapted them with fireproof plating to weather the ash storms there. Note that NanoTrasen has not spared the expense to make them lavaproof. Its price has been adjusted accordingly.
tweak: The wizard variant of the cursed heart, that heals the user, now has a chance to spawn in necropolis chests along with the original that does not posses healing traits.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/hardsuit.dm
+++ b/code/modules/clothing/spacesuits/hardsuit.dm
@@ -167,23 +167,25 @@
 	//Mining hardsuit
 /obj/item/clothing/head/helmet/space/hardsuit/mining
 	name = "mining hardsuit helmet"
-	desc = "A special helmet designed for work in a hazardous, low pressure environment. Has reinforced plating for wildlife encounters and dual floodlights."
+	desc = "A special helmet designed for work in a hazardous, low pressure environment. Has reinforced plating for wildlife encounters and ash storms. Has dual floodlights on the helmet."
 	icon_state = "hardsuit0-mining"
 	item_state = "mining_helm"
 	item_color = "mining"
 	armor = list(melee = 30, bullet = 5, laser = 10, energy = 5, bomb = 50, bio = 100, rad = 50)
 	brightness_on = 7
 	allowed = list(/obj/item/device/flashlight,/obj/item/weapon/tank/internals, /obj/item/weapon/resonator, /obj/item/device/mining_scanner, /obj/item/device/t_scanner/adv_mining_scanner, /obj/item/weapon/gun/energy/kinetic_accelerator)
+	max_heat_protection_temperature = FIRE_IMMUNITY_SUIT_MAX_TEMP_PROTECT
 
 
 /obj/item/clothing/suit/space/hardsuit/mining
 	icon_state = "hardsuit-mining"
 	name = "mining hardsuit"
-	desc = "A special suit that protects against hazardous, low pressure environments. Has reinforced plating for wildlife encounters."
+	desc = "A special suit that protects against hazardous, low pressure environments. Has reinforced plating for wildlife encounters and ash storms."
 	item_state = "mining_hardsuit"
 	armor = list(melee = 30, bullet = 5, laser = 10, energy = 5, bomb = 50, bio = 100, rad = 50)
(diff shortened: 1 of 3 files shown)


ShadowDeath6:
note to self make commits better

X-TheDark:
I absolutely disagree on the wizard cursed heart being there. It heals 25 brute/burn/oxydamage (total of 75 healing) per pump (that's every 5-6 seconds). It's way too good in keeping someone in an armored suit alive.

Stealthkibbler:
4000 points seems like way to much, 3000 would be more reasonable because the suit also slows you down making it harder to combat a lot of the fauna, even stimpacks barely speed you up when wearing it.

ShadowDeath6:
@X-TheDark You have to remember that you have to correctly time the pumps while fighting monsters and in really bad cases ash storms where you need to be focusing on everything else and you might forget a pump which will make you lose however much blood. Reminder that this is in a necropolis chest, which also contains such gems as a nightvision + xray vision pair of glasses and a holoparasite generator, which if it's set to healing it's infinitely better than a cursed heart.

@Stealthkibbler it's full space and fire immunity, and it can be upgraded. We're trying to make miners actually get ores for science before fucking off.

Super3222:
Healing from every type of damage (brute, burn, toxin, and oxy) in 5 seconds flat is BAD. It's more OP than the cabbages current healing prowess. I'd prefer the current one we have, with no good effects, and better off researched in RnD for level 6 biotech.

ShadowDeath6:
You have to remember that it doesn't automatically heal, you need to pump the heart. If you're stuck in a goliath tentacle, you're going to lose a lot of blood, and you need to keep a close eye on timing your beats if you want to stay alive. It fits with the "benefit, but still challenging" theme of lavaland and making an item that fucks you over in every conceivable way is not good. You can get automatic healing in the form of legion souls implanted in you, and you can still die, it's not like you're immortal.

Stealthkibbler:
@ShadowDeath6 You should talk to Mac about the 6k point ELITE hardsuit he's porting over and probably just have it so this suit upgrades into that, otherwise spending 4k points on this and then trying to upgrade to the ELITE one is a pain in the ass since you gotta start from 0k again.

Super3222:
@Stealthkibbler It's not a port from anywhere. It's his own idea.

GrayRachnid:
@Stealthkibbler The point is you can either get this at 4000 points or you can save another 2000 for more benefits.

Stealthkibbler:
@Grayrachnid That's very counter productive as a miner though, it'd be better if this upgraded into the Elite so that you're investing into something that will help you even more in the long run.

ShadowDeath6:
Could be an upgrade purchasable for 2k points for the hardsuit which upgrades its armor and makes it lavaproof, but we'll see

TheSadFox:
Make the chests have 50/50 chance of spawning either the shitty or wizard heart, and don't give miners any way to see which one is it. Its really powerfull and the need to pump it is hardly an issue, just move the action button next to your inventory or wherever you keep your mouse at most of the time.

> Difficult tendrils

These don't exist.

ShadowDeath6:
the wizard cursed hearts now have the same chance of spawning as the regular ones, good luck telling which is which xddd

TheSadFox:
Theres no fun without risk.

ParagonPrime:
why is this STILL NOT MERGED

ShadowDeath6:
balance memes r ded

PR #255, opened 2016-07-02
Remove charter.

george99g:
Too much potential for abuse, especially when no admins are online.


Nirnael:
Make it only usable when there are admins on, don't remove it.

Super3222:
Changing the station's name is fun. Don't remove it, that'd be silly.

It's like changing a borgs name, but instead everyone sees it. If it violates the rules then they get punished for it, just like borg names or naming stuff in general.

EDIT:
After realizing how abusable this is, we should make it alert admins and allow them to decide if it's legit or not.

Issue #77, opened 2016-06-12
Port: Space dolphins


timkoster1:
HOW CAN WE FORGET THIS BEAUTY?!!

george99g:
Closing this since it was ported by #79

george99g:
Never mind.

timkoster1:
Gotta keep it open to prevent copies

PR #260, opened 2016-07-02
Port - Dolphins!

SKR6DarkVoid:
That's right, re-enabled, icons imported and includes set, hud
re-imported for tactical dolphins as well.

Edit _fixes time's accidental boop of forgetting to put them in the include, for the .dme._
#77


Time-Green:
Wot? It worked fine on muh local. May be because I never PR my DME.
Anyway, icons were fine. Wich icons did you port over that I did not?

Well, good job!

PR #265, opened 2016-07-03
Fixes random dark turfs

MrsTonedOne:
This was causing turfs changed at world init from a no lighting system turf to a turf that used the lighting system to have two lighting overlays, one randomly stuck at black depending on timing.


MacHac:
:+1: Thanks, MSO.

Although, to be honest, I've never experienced this bug myself.  But hey, what do I know?

MrsTonedOne:
Narnael says they see it on mining turfs near ruins

Nirnael:
Yeah it happens every round, particularly around generated lavaland ruins with lava around them.

PR #267, opened 2016-07-03
Coder OOC tags will now be properly presented

Super3222:
Gives coders their OOC tag.
#188

:cl: Super3222
bugfix: Coders now have proper OOC tags!
bugfix: Donators now have proper OOC tags!
/:cl:


Kmc2000:
What about admins.....

We don't have one anymore

Super3222:
You do.

PR #275, opened 2016-07-04
Malware

Cruix:
Added a software system, and changed the Honk virus, Mime virus, and Detomax to use it. In the future, they might be guarded against using antivirus software, but at the moment nothing has that unless an admin creates it. The Honk virus, Mime virus, and Detomax are functionally unchanged except that sending multiple honk viruses to one person will now stack.

Also added the Adware event, which has the same chance to occur as spontaneous appendicitis. It will infect three random PDAs with adware, which will cause all PDA button presses (it does not affect PDA uplinks) to have a chance of creating a popup-ad on the user's screen. The virus will disappear after a certain number of ads, and sending a PDA message to someone has a chance of infecting that persons PDA with a copy of the virus.

Also added the admin-spawn only stuxnet virus, which will spread through the station's powernet until if finds the singularity containment equipment, at which point it will max the PA and turn off the emitters. The only way to defend against it currently is to isolate the emmiters and PA from the general powernet, and deconstruct and reconstruct them all if they get infected. I do not recommend ever spawning it.

EMPs will delete all viruses on a machine unless that machine is EMP-proof, which all the singularity machines are. Active software can be viewed, created, and deleted with the new softwaremaster cartridge.

Here are the two image ads from KMC:
ad1

ad2

ad3

ad4

ad5

ad6

ad7


code/datums/software/malware.dm
```diff
@@ -0,0 +1,214 @@
...
+	flags = SOFTWARE_NOSPREAD|SOFTWARE_NOMUTATE|SOFTWARE_ANTIVIRUS_IMMUNE
+	can_infect = list(/obj/item/device/pda)
+
+/datum/software/malware/detomatix/infect()
+	if(!..())
+		return 0
+	if(host)
+		var/obj/item/device/pda/PDA = host
+		PDA.explode()
+		uninfect()
```
X-TheDark, line 92: I think the order needs to be reversed. PDA gets deleted when it explodes, so uninfect() will try to mess around (`get_software_list`/etc) with a qdel'ed (or `null`, depends on if it deletes it in time) object.

Cruix, line 92: explode() is not guaranteed to destroy the PDA, such as if this virus somehow got sent to a non-exploding PDA like the AI's. If the host is deleted before uninfect() is called, nothing bad will happen. The worst-case scenario is that qdel gets called on the virus twice.

ShadowDeath6:
not sure if the AI's internal PDA actually works like a PDA, but make sure the AI can't get malware since 
1. it's annoying as hell to have another popup as the role with a fuckton of popup lag
2. it's the most advanced piece of machinery on the station and it can't get detonated by detomatrix and (maybe) can't get infected by mime and clown viruses

code/datums/software/ads.dm
```diff
@@ -0,0 +1,241 @@
...
+	return ""
+
+/datum/popup_ad/ad1
+	size = "400x350"
+
+/datum/popup_ad/ad1/generate_content(var/topic_ref)
+	content = {"
+<head><title>NT Product Ads</title></head>
+<a href='byond://?src=\ref[topic_ref];ad_click=1'>
+<img src=ad1.png>
```
oisin100: So. Much. Hardcode. Could you not have a inherited proc generate the HTML and use an inputted image URL?

AsV9:
This is amazing.

Kmc2000:
M e r g e p l s
E
R
G
E

PR #279, opened 2016-07-04
Ports StonedMC

X-TheDark:
Rebases over StonedMC : original PR https://github.com/tgstation/tgstation/pull/17987 by @MrStonedOne 
Also ports the lightning fix : original PR https://github.com/tgstation/tgstation/pull/19060 by @FlavoredCactus

(Incorporates the cleanbot sound->New() thing since that was ported earlier)

Compiled and tested, seems to run and the game works and stuff.


MrsTonedOne:
> Zero idea how this got into here, probably during resolving conflicts

One of StonedMC's changes was to set the default fps to 20 (0.5 tick_lag) you can keep or leave it if you want.

X-TheDark:
I thought I saw that the fps change was done in `configuration.dm` file, rather than `world.dm`. And yes, we have like a dynamic fps management thing, so we don't really have any use for these vars anyway, but they are nice to have, nonetheless.

PR #280, opened 2016-07-04
Enables CTF after round end.

Nirnael:
Enables CTF after round end.

:cl: Nirnael
tweak: CTF can now be played after round ends.
/:cl:

### DIFF
--- a/code/modules/awaymissions/capture_the_flag.dm
+++ b/code/modules/awaymissions/capture_the_flag.dm
@@ -133,7 +133,7 @@
 /obj/machinery/capture_the_flag/attack_ghost(mob/user)
 	if(ctf_enabled == FALSE)
 		return
-	if(ticker.current_state >= GAME_STATE_PLAYING)
+	if(ticker.current_state < GAME_STATE_PLAYING)
 		return
 	if(user.ckey in team_members)
 		if(user.mind.current && user.mind.current.timeofdeath + respawn_cooldown > world.time)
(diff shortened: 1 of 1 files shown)


code/modules/awaymissions/capture_the_flag.dm
```diff
@@ -133,7 +133,7 @@
 /obj/machinery/capture_the_flag/attack_ghost(mob/user)
 	if(ctf_enabled == FALSE)
 		return
-	if(ticker.current_state != GAME_STATE_PLAYING)
+	if(ticker.current_state >= GAME_STATE_PLAYING)
 		return
```
X-TheDark, line 137:
``` c
#define GAME_STATE_PLAYING  3
#define GAME_STATE_FINISHED 4
```

Erm...not quite. Need `<=`

george99g:
I know this is a single character change, but...

``` c
//ticker.current_state values
#define GAME_STATE_STARTUP        0
#define GAME_STATE_PREGAME        1
#define GAME_STATE_SETTING_UP    2
#define GAME_STATE_PLAYING        3
#define GAME_STATE_FINISHED        4
```

Shouldn't this be `<=` instead, since `>=` will make it not work while the game is playing and after it already ends, due to the return.

Nirnael:
Yes sorry, my bad.

X-TheDark:
Wonder if CTF should enable automatically when the round end gets delayed (due to tickets).

Nirnael:
Hmm sounds like a good idea, i'll do it in a bit.

Nirnael:
Actually, gonna leave it at that for now, if anyone wants to do it, do it, if not I'll do it when I'm done with my stuff. I'm not coding anything longer than a few lines yet.

PR #282, opened 2016-07-04
Borgs now get their MMI deleted when they fall into chasm

### LINKED ISSUE #281: If a borg falls into a chasm, the MMI drops on the chasm tile and stays there.
dreamseeker_2016-07-04_16-48-04

X-TheDark:
fixes #281 

This is incredibly hacky, but I can't think of a better way, since Borgs eject their MMI on qdel().

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -626,6 +626,11 @@
 /turf/open/chasm/straight_down/lava_land_surface/drop(atom/movable/AM)
 	if(!AM.invisibility)
 		visible_message("[AM] falls into [src]!")
+		if(istype(AM, /mob/living/silicon/robot)) //MMI deletion
+			var/mob/living/silicon/robot/BORGO = AM
+			if(BORGO.mmi)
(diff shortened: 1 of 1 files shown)


X-TheDark:
I tried doing this in a less retarded way by adjusting `/robot/Destroy()` (and removing the snowflake from mining_turfs.dm, naturally) to something like

``` c
/mob/living/silicon/robot/Destroy()
    if(mmi && mind)//Safety for when a cyborg gets dust()ed. Or there is no MMI inside.
        var/turf/T = get_turf(loc)//To hopefully prevent run time errors.
        if(T)
            mmi.forceMove(T)
        if(!qdeleted(mmi)) //If we are NOT in the mythical nullspace...looking at you, chasms
            if(mmi.brainmob)
                    <snip>
```

This does work properly in moving your MMI and deleting it, but this fucks up your hotkeys, making it impossible to get the right click context menu until you relog.

X-TheDark:
Actually, this current (hacky) implementation also breaks your hotkeys. Until I (or anyone else) can figure this out, I'll close this...

Issue #272, opened 2016-07-04
Can we remove the ability to summon an infinite amount of ratvars?


GrayRachnid:
I mean, I understand the meme about having a spare gate somewhere else. However that is a bit OP.

And if we're keeping the meme of having spare gates, atleast make it so that there can only be 1 ratvar on the map, and once there's a ratvar you can't build any more gates.

No fun yeah yeah but the noise is unbearable.

timkoster1:
Feature

Super3222:
Not a feature. There is only one Rat'Var just like there is only one Nair'Se.

timkoster1:
But it's funny as fuck
;_;

PR #298, opened 2016-07-06
There is only one Rat'Var

Super3222:
Fixes https://github.com/yogstation13/yogstation/issues/272
Pr: When the gateway notices rat'var has already been summoned, it will shut it'self.

### DIFF
--- a/code/game/gamemodes/clock_cult/clock_ratvar.dm
+++ b/code/game/gamemodes/clock_cult/clock_ratvar.dm
@@ -98,6 +98,11 @@
 				sleep(131)
 				animate(src, transform = matrix() * 3, alpha = 0, time = 5)
 				sleep(5)
+				var/machine_god = /obj/structure/clockwork/massive/ratvar
+				var/obj/him = locate(machine_god, SSobj.processing)
+				if(him) // a check to make sure rat'var doesn't already exist
+					qdel(src)
+					return
 				new/obj/structure/clockwork/massive/ratvar(get_turf(src))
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/clock_cult/clock_ratvar.dm
```diff
@@ -98,6 +98,11 @@
 				sleep(131)
 				animate(src, transform = matrix() * 3, alpha = 0, time = 5)
 				sleep(5)
+				var/machine_god = /obj/structure/clockwork/massive/ratvar
+				var/obj/him = locate(machine_god, SSobj.processing)
```
X-TheDark: There's a perfectly good global `ratvar_awakens` variable that gets set to TRUE when ratvar/New() is invoked for you to use.

Super3222: Oh wow. I spent a while looking for definitions that tell if rat'var was summoned or not, and got nothing.

ChangelingRain:
You should just port all the clock cult changes from tg instead, this bug is fixed already.
https://github.com/tgstation/tgstation/pull/18502

There's like 100+ prs worth of clockcult fixes and you'll need basically all of them.

Stealthkibbler:
is our clock cult still outrageously outdated? We should probably just port the latest clock cult before actually making any changes to the mode.

Super3222:
Going to close this, because /tg/ has already fixed this. And we'll be doing one last update soon enough.

PR #317, opened 2016-07-08
Yogbot

oisin100:
Adds discord messaging to the following Elements:
- Asay
- OOC
- Tickets
- Round begin

Also adds a C++ DLL to send future POST requests to any website.

:cl:
rscadd: OOC And Asay are now sent to Discord.
rscadd: Admins are now alerted via discord when there are open tickets and no admins online.
rscadd: Round begin are now announced through discord.
/:cl:


code/__DEFINES/discord.dm
```diff
@@ -0,0 +1,9 @@
+#define DISCORD_PUBLIC             "134720091576205312"
+#define DISCORD_COUNCILPUBIC       "186490543398715392"
+#define DISCORD_GAMES              "162018753201045504"
+#define DISCORD_OOC                "161524149355806720"
+#define DISCORD_ASAY               "161564229852332032"
+#define DISCORD_CODER              "134722107190935552"
+#define DISCORD_ADMIN              "134722036353204224"
+#define DISCORD_IMPORTANTADMIN     "134722058688004096"
+#define DISCORD_COUNCIL            "134722015503319040"
```
george99g: Hardcode alert.

code/modules/client/client_procs.dm
```diff
@@ -178,14 +178,14 @@ var/next_external_rsc = 0
 	directory[ckey] = src
 
 	//Admin Authorisation
-	
+
```
oisin100, line 181: Inconsistent line endings ¯_(ツ)_/¯

code/__DEFINES/discord.dm
oisin100: Aye, Not much i can do about that unless i make another post request to retrieve the numbers, Which just seems like overkill.

george99g: Can't you just move this into a config file that gets loaded at runtime?

oisin100: I could yes, But then i wouldnt be able to use the swanky defines. It would have to be something like a global string index array, accessible via something like `discord['ASAY']`

george99g: How is that worse than a define?

skull132: Could define these as strings or integers, and then sift through them in proc/send_discord_message. Something like this, for example.

Define:

```
// Strings if you want to use liiists that are indexed with them.
// Otherwise, ints will probably be better.
#define DISCORD_ADMIN "discord_admin"
#define DISCORD_CODER "discord_coder"
```

some_file.dm:

```
some_proc()
    ..()
    send_message_discord(DISCORD_CODER, "Help, something awful just broke!")
```

code\modules\http\discord.dm:

```
/proc/send_discord_message(var/channel_str, var/message)
    if (config.discord_token == "nodiscord")
         return
    if (length(message) > 2000)
        return

    var/channel = get_discord_channel(channel_str)
    if (!channel)
        return

    var/json_message = "{ \"content\" : \"[message]\" }"

    call("ByondPOST.dll", "send_post_request")("https://discordapp.com/api/channels/[channel]/messages", json_message, "Authorization: [discord_token]", "Content-Type: application/json")

/proc/get_discord_channel(var/channel_str)
    switch (channel_str)
        if (DISCORD_ADMIN)
            return config.discord_ch_admin
        if (DISCORD_CODER)
            return config.discord_ch_coder
        else
            return null
```

You can also avoid the massive switch array if you store them in a list that is indexed with the define values:
`config.discord_channels = list(DISCORD_ADMIN = '21354', DISCORD_CODER = '45684')` and then fetch them with this: `return config.discord_channels[channel_str]`

skull132:
Also, any chance you could put a decompiled solution and project of ByondPOST.dll up somewhere? I'm curious to have a look at it, and it'd be useful for some shenanigans over at Aurora.

oisin100:
> Also, any chance you could put a decompiled solution and project of ByondPOST.dll up somewhere? I'm curious to have a look at it, and it'd be useful for some shenanigans over at Aurora.

Will do, Check back tommorow.

oisin100:
Oh dear...

PR #319, opened 2016-07-09
More fixes

### LINKED ISSUE #159: Tickets don't auto-update.
Talking about the ticket window, usually when an admin or the player makes a new message it auto-updates it and scrolls down. However I have to click "refresh" everytime.

### LINKED ISSUE #223: The search function on security consoles does not function.

Cruix:
:cl:
bugfix: The search function on the security record console now works.
/:cl:
Fixed a runtime when drones would click on a human with an empty hand.
Fixed security record console searching not working. Fixes #223
Fixed tickets not auto updating. Fixes #159 

Ticket and sec console searching was broken because the jquery library file was renamed. I have no idea why it still worked for Super3332.

### DIFF
--- a/code/game/machinery/computer/security.dm
+++ b/code/game/machinery/computer/security.dm
@@ -60,7 +60,7 @@
 					dat += {"
 
 		<head>
-			<script src="libraries.min.js"></script>
+			<script src="jquery.min.js"></script>
 			<script type='text/javascript'>
 
 				function updateSearch(){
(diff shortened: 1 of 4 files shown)


Super3222:
> Super3332

I believe it may have possibly been a cache problem. I investigated it with GrayRachnid, because they had recently cleared their cache and were having these problems (with tickets and the security console). I, as you already know, didn't have any problems at all, and this may have been happening because I had downloaded files (from the old code) still in my cache. It was until that I decided to clear my cache was when I was experiencing the same problems. But that's just my guess, I'm sure that there's a logical explanation for this that I'm not getting at.

Cruix:
If you did experience those problems after clearing your cache, then that was almost certainly it.

Issue #339, opened 2016-07-11
Inmunity to shadowlings


IgnisCorvus:
[Note]: # In a round of shadowling mode, two shadowling tried to convert me at the same time. When the first finished i was a thrall, but when the second finished moments later, i was deconverted. They tried again, but this time only one, the result: i could not be converted by shadowlings. I was not implanted and the lings spells could affect me. Just i couldnt be convert. We tried a couples of times more.
I wrote this here, but i will try to find the problem later. Honestly i am not sure about what can i do here.

FluffySurvivor:
LOL

IgnisCorvus:
what? i reporte in game two times and i didnt see here.

PR #367, opened 2016-07-15
Bug Fixes and Random Event Weight Drops

Super3222:
Fixes https://github.com/yogstation13/yogstation/issues/339.
- Swarmer event weight lowered to 3 (from 7).
- It can only spawn now 45 minutes into the round (from 30).
- Alien Infestation event weight lowered to 3 (from 5).
- Enthrall now has a check for whether the target is already a thrall or not.
- Enthrall works on loyalists again
- Abstract shoes (shoes like golems feet or shadowling chintin feet) will no longer "show up" when footprints are examined

:cl: Super3222
tweak: Swarmer weight chance is lowered from 5 to 2.
tweak: Alien Infestation event chance is lowered from 5 to 3.
tweak: Brand Intelligence event chance is lowered from 5 to 3.
tweak: Meteor Event chance is lowered from 5 to 3.
tweak: Now you're probably wondering what we're getting from all of these tweaks decreasing the chance of your favorite events. The catch is, that all of them now have a growth feature. So if the event doesn't run, for instance the Alien Event, it's weight will increase every hour the round goes on for. Are you up to the challenge?
tweak: Enthrall now works on loyalists again.
tweak: Golem and Shadowling feet no longer show up in footprints.
/:cl:


MacHac:
> lowering to 1

I appreciate that they weren't uncommon, but that's going to practically remove them from play.  At least leave it at 2-3.

GrayRachnid:
Agreed with Mac 2-3 sounds better. 1 is very VERY VERY uncommon.

oisin100:
Its not as bad as xeno, bump it up to 2.

ghost:
bump both to 2-3 

aliens are good and swarmers are good also

or instead, make the chance start low and increase scaling based on time passed so that towards the end of a round like 2 hours or shit bad shit starts to happen

Super3222:
the longer you stay on the station, the more of a chance chaos is bound to happen
will look into

Super3222:
In the last few commits:
- Brand Intelligence and Meteor weights have dropped.
- All of the events that have had weight drops now have 'growth' enabled to them, so in the instance that it's been an hour or two, the weight of these events will grow.

code/controllers/subsystem/events.dm
```diff
@@ -90,6 +90,12 @@ var/datum/subsystem/events/SSevent
 
 	for(var/datum/round_event_control/E in control)
 		if(!E.canSpawnEvent(players_amt, gamemode))
+			if (E.growth)
+				var/estimatedtime = PRESUMEDHOUR * E.growth
+				if(world.time > estimatedtime)
+					E.weight++
+					E.growth++
+					message_admins("Random event [E.name] could not spawn, so it's weight has grown to [E.weight].")
```
oisin100, line 96: Will this not output a spam of messages saying ""Random event [E.name] could not spawn, so it's weight has grown to [E.weight]."" For every event with growth?

Super3222, line 96: Yeah, I thought admins should know when rounds are close to becoming chaotic.
I should also probably also up some the event weights. 2 is pretty low.

code/game/objects/effects/decals/Cleanable/humans.dm
```diff
@@ -178,6 +178,8 @@
 		. += "You recognise the footprints as belonging to:\n"
 		for(var/shoe in shoe_types)
 			var/obj/item/clothing/shoes/S = shoe
+			if(S.flags & ABSTRACT)
+				break
```
Alek2ander: So examining a footprint made with abstract shoes will tell you "You recognise the footprints as belonging to:" without any follow-up lines
A better way would be to not add abstract shoes to the shoe list in the decal in the first place

Super3222: I don't know, I thought it'd be interesting if an abstract foot within one print would make it unreadable. Making the user question whether they should follow it or not.

Super3222:
Yup, I suspected something like this would happen. Give me a second.

PR #398, opened 2016-07-22
Yogbot

oisin100:
Adds discord messaging to the following Elements:
- Asay
- OOC
- Tickets
- Round begin

Also adds a C++ DLL to send future POST requests to any website.

:cl:
rscadd: OOC And Asay are now sent to Discord.
rscadd: Admins are now alerted via discord when there are open tickets and no admins online.
rscadd: Round begin are now announced through discord.
/:cl:


tools/ByondPostDLL/ByondDLL.cpp
```diff
@@ -0,0 +1,49 @@
+/*
+	Copyright (C) 2016 Oisin Carr
+
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
```
Iamgoofball, line 5: Nope.

This is AGPLv3 due to being in the codebase. Fix your license.

licenses/libcurl-license
```diff
@@ -0,0 +1,22 @@
+COPYRIGHT AND PERMISSION NOTICE
```
Iamgoofball, line 1: Nope, this is AGPLv3 due to the codebase.

RemieRichards, line 1: No, this isn't required to for the game to run (or shouldn't be) and as such can be added to the codebase under it's own licensing (so long as it's clearly stated, as it is)

tools/ByondPostDLL/ByondDLL.cpp
george99g, line 5: https://www.gnu.org/licenses/gpl-faq.en.html#GPLModuleLicense

Pretty sure the GPL is compatible with the AGPL.

Plus, I don't think you can actually even follow the AGPL with this.

PR #404, opened 2016-07-24
Removes "faggot" from brain damage lines

Cruix:


Iamgoofball:
not a fix, and I don't see why you're tagging it as such

Super3222:
True.

ParagonPrime:
snowfleks
snowfleks all

PR #415, opened 2016-07-26
Dismemberment fixes.

TheSadFox:
Steals https://github.com/tgstation/tgstation/pull/19261.

Adds IS_SHARP to some items missing them.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types.dm
+++ b/code/modules/mob/living/carbon/human/species_types.dm
@@ -122,10 +122,10 @@ datum/species/lizard/before_equip_job(datum/job/J, mob/living/carbon/human/H)
 /datum/species/lizard/spec_death(gibbed, mob/living/carbon/human/H)
 	if(H)
 		H.endTailWag()
-
 /*
  Lizard subspecies: ASHWALKERS
 */
+
 /datum/species/lizard/ashwalker
 	name = "Ash Walker"
 	id = "lizard"
--- a/code/game/objects/items/weapons/melee/energy.dm
+++ b/code/game/objects/items/weapons/melee/energy.dm
@@ -30,6 +30,7 @@
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	throw_speed = 3
 	throw_range = 5
+	sharpness = IS_SHARP
 	w_class = 3
 	w_class_on = 5
 	flags = CONDUCT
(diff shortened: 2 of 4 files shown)


code/modules/mob/living/carbon/human/species_types.dm
```diff
@@ -122,10 +122,10 @@ datum/species/lizard/before_equip_job(datum/job/J, mob/living/carbon/human/H)
 /datum/species/lizard/spec_death(gibbed, mob/living/carbon/human/H)
 	if(H)
 		H.endTailWag()
-
 /*
  Lizard subspecies: ASHWALKERS
 */
+
```
TheSadFox, line 128: I have no idea how'd that happen

Cruix, line 128: Literally unplayable.

Time-Green:
While your add it, make a kitchen and butcher knife sharp aswell.

Cruix:
You can sharpen knives and cleavers with a sharpening block in the chef vendor. I think it is to limit the ability of people who vault the table and steal knives as weapons, since there are fewer sharpening blocks than there are knives.

TheSadFox:
Both are `IS_SHARP_ACCURATE` already. Cleaver is fully capable of dismembering without any modifications, knife will never dismember due to it being `W_CLASS = 2`, and for a weapon to be able to dismember it has to be at least `W_CLASS = 3`.

Edit. Also @Cruix, the sharpener doesn't really "sharpen" the items in a way that would allow say, a baton to dismember. The weapon needs to already have sharpeness of 1 or 2 for it to work with the sharpener. All it really does is increase weapons force, thus increasing the chance of dismemberment occuring.

Cruix:
If knives and cleavers are already sharp then there is even less reason to make them sharp in this PR.

TheSadFox:
> Both are IS_SHARP_ACCURATE already. Cleaver is fully capable of dismembering without any modifications, knife will never dismember due to it being W_CLASS = 2, and for a weapon to be able to dismember it has to be at least W_CLASS = 3.

Time-Green:
I tried to cut of my legs with a knife in order to crawl at faster then run speed.
Might also need to make a PR for that.

TheSadFox:
> **knife will never dismember due to it being W_CLASS = 2, and for a weapon to be able to dismember it has to be at least W_CLASS = 3.**

Time-Green:
Are you sentient? I already read that, I just told my story behind it.

PR #419, opened 2016-07-26
Fixes a few shadowling bugs and invests growth into 4 events

Super3222:
Fixes https://github.com/yogstation13/yogstation/issues/339

Bugs:
- Abstract feet (like golem or chitin feet) could be discovered by examining bloody footprints.
- Immunity to shadowling conversion bug in https://github.com/yogstation13/yogstation/issues/339

Events:
- Alien Infestation, Brand Intelligence, Swarmer Shell, and Meteor events have had a weight drop from 5 to 3.
- Leaning closer towards chaos, their weights will slowly increase every passing hour.

:cl: Super3222
add: Alien Infestation, Brand Intelligence (rogue vendors), Swarmer shells, and meteor events now have all had a drop of weight. To compensate, their weights now grow every passing hour.
fix: Bloody footprints from blood tails can no longer be "identified".
fix: You can no longer get immune to shadowling conversions when two of them enthrall you at the same time.
/:cl:


ParagonPrime:
This seems bad, honestly.

SS13 rounds are, at most, two and a half hours. Thus the "growth" is redundant because it'll grow twice, if you're lucky, and still not happen.

Growth should be by half hour.

FluffySurvivor:
This is terrible. Very rarely rounds pass 1 hour. With less events there is less stuff to do which results in boredom.

Super3222:
Alright, don't say I didn't warn you if I make growth a half hour.

FluffySurvivor:
15 mins and we're good

Super3222:
15 minute sounds fun in theory, but in actual implementation it'll be a clusterfuck. Does anyone else have a differing opinion on 15mins?

FluffySurvivor:
Nuke ops, rev, gang, cults, sling, blob don't last more than 30 mins. 

The only difference 15 mins do is that it makes long traitor/ling rounds less boring.

Greatbigben:
Hell no, 15 mins is too short. 30 is good enough

X-TheDark:
Keep it at 30 for the first pass. It can always be adjusted as necessary.

Stealthkibbler:
Yeah make it 30, remember when they made space ninja a 1hr thing? It either happened too late to have an impact on the round or never happened at all.

PR #422, opened 2016-07-28
Misc Map Changes (Botany and Cargo)

Super3222:
- Removes that stock computer from the post office (the one that was accessible by people standing on the other end
- Adds another job locker in the back of botany since there's 3 botanist slots and we only have TWO lockers


FluffySurvivor:
More winter coats for botany pls

PR #432, opened 2016-07-30
This makes paramedic hurt so much less.

AsV9:
:cl:
rscadd: Rollerbeds fit in your backpack when collapsed now. Now they're actually useful!
/:cl:

### DIFF
--- a/code/game/objects/structures/beds_chairs/bed.dm
+++ b/code/game/objects/structures/beds_chairs/bed.dm
@@ -95,7 +95,7 @@
 	desc = "A collapsed roller bed that can be carried around."
 	icon = 'icons/obj/rollerbed.dmi'
 	icon_state = "folded"
-	w_class = 4 // Can't be put in backpacks.
+	w_class = 3 // QUALITY OF LIFE
 
 
 /obj/item/roller/attack_self(mob/user)


Super3222:
""""""""""""Critical""""""""""""

We should consider implementing a system where dragging a bleeding person across the floor would hurt them to some degree to make roller beds actually good (which would take time for the playerbase to get use to).

Then roller beds would be a counter to this, making them somewhat useful and the roller beds hanging out in medbay somewhat useful, and roller beds in general definitely more useful.

AsV9:
> We should consider implementing a system where dragging a bleeding person across the floor would hurt them to some degree to make roller beds actually good (which would take time for the playerbase to get use to).
> 
> baystation soon

Seriously though I'd rather hurt someone and heal them back up then having to waste a precious hand carrying a stupid _collapsed_ rollerbed.

X-TheDark:
Since you can no longer quickdrag (you can't, right?) and move at runspeed while dragging, I think rollerbeds are really good now. This is probably not too powercreep'y.

TheSadFox:
You can move at normal speed when dragging with passive grab still, Dark. Its essentially old pull now.

ParagonPrime:
asv can i marry you

X-TheDark:
> You can move at normal speed when dragging with passive grab still

If that's the case, there's really no reason to use rollerbeds (other than bloodtrails, which really don't do anything except wizard rounds).

They'd need to be made to be able to have several dead people buckled to it. A pocket-fitting "locker" for dead bodies at least sounds somewhat useful.

ParagonPrime:
i'm still in favour of the "dragging without bed = damage" thing, though maybe put a few rollerbeds in public, easy-to-reach areas just in case

Alek2ander:
public rollerbeds are called lockers

PR #449, opened 2016-08-02
Prevents StonedMC from freezing when minigun itself is spawned rather than the backpack

X-TheDark:
Also replaces "minigun" in CentComm (the gun itself) with the proper one (the backpack that is part of and has the minigun).

Also moves the `..()` call so it doesn't add 5k+ items (gun+ammo) to the deletion queue if it's spawned by itself just because it instantiates all the ammo even if the gun needs to be deleted.


Kmc2000:
:speedmerge:

PR #450, opened 2016-08-03
Nerfs detective's revolver brute damage from 15 to 10

X-TheDark:
As requested by @GrayRachnid

### DIFF
--- a/code/modules/projectiles/projectile/bullets.dm
+++ b/code/modules/projectiles/projectile/bullets.dm
@@ -12,7 +12,7 @@
 	stamina = 80
 
 /obj/item/projectile/bullet/weakbullet2 //detective revolver instastuns, but multiple shots are better for keeping punks down
-	damage = 15
+	damage = 10
 	weaken = 3
 	stamina = 50
 


Zostroll:
The revolver is suppose to be op...
Edit: I realised you were nerfing the dets gun not the syndies gun. Go ignore my comment.

X-TheDark:
Yeah, my bad for not clarifying that.

FluffySurvivor:
The damage isn't the problem. It's the double form of CC in it (50 stamina and a stun). 

You should reduce the stamina to 25 so it takes more effort for a full stamina stun.

X-TheDark:
Gray's concern was basically critting people in one clip (90 damage + bleed). I'd like more feedback about the stamina damage, if it is an issue.

george99g:
I don't see a problem with the stamina damage. I'm also okay with the damage reduction.

FluffySurvivor:
You can easily crit people with it because  one initial shot= no escaping

Super3222:
There's a reason why this gun was given to a non-security investigator.

Stamina damage is the last thing that needs a nerf, because it separates the detective's gun from other lethal revolvers and gives it some actual purpose. It's stun isn't that long, in fact it lasts for a bit. The reason it does stamina damage is to make sure the guy you're shooting is down.

In Fluffy's eyes, apparently I need to shoot a guy trying to attack me with an esword 4 times in order to handcuff them (as the detective). ( (15 + 25) + (15 + 25) + (15 + 25) + (15 + 25) Do the math yourself, man ).

And guess what? With this nerf, I'd need an extra shot too.

Brilliant idea, so now I need 5/6 of my entire clip to stun a guy.

Obviously I don't agree with this. :-1:

At least you aren't trying to replace their gun with a stun revolver.

X-TheDark:
Just for clarification sake, do you not agree with the brute damage nerf or Fluffy's suggestion of halving the stamina damage?

Super3222:
I don't like how it's nerfing the Detective's revolver and, I especially don't agree with Fluffy's stamina suggestion.

Getting shot up with 6 bullets directly should leave you at a very critical state. All guns in this game are dangerous (aside from maybe 1 or 2 silly ones).

Jc1469:
I was the one that originally brought up the detective's revolver being OP, seeing how often I play it.
That gun can end a wizard round, heck, end any antag, with 5 bullets.   If you even need to hit them with the 6th, the bleeding will have them in crit regardless.  Stamina and weakness don't need touching.  The goal is to make it less lethal, not make it so the detective actually needs to go get a fucking taser.

ShadowDeath6:
I never liked weapons that stunned and did damage in the same shot, it just felt too easy, especially since the det's revolver fucks you over 3 ways (long stun, damage, and stamina crit), but this is a pretty decent way of handling it

FluffySurvivor:
dets revolver = taser laser and disabler all in one

Super3222:
:+1:

darkwahn:
And here I was going to try to push for Det's gun not being a toy at all anymore.

PR #456, opened 2016-08-03
Organ Decay

Cruix:
Ports organ decay and freezer closets and crates cooling their contents. Since organ decay uses SSobj.processing instead of world.time, it is possible that dead bodies will last a little longer if the server is lagging heavily.

:cl:
rscadd: You can once again perform heart transplants to defibrillate people whose bodies have decayed.
rscadd: Organs once again decay when they are not inside a living body, unless the temperature is very low.
rscadd: Freezer crates and closets once again cool their contents down enough to preserve organs or dead bodies.
/:cl:

Incidentally, freezers now cool you down enough to make cryoxidone work at its least effectiveness. Cold showers or a dip in vacuum are still more effective, but it works for small heals.

### DIFF
--- a/code/game/gamemodes/miniantags/slaughter/slaughter.dm
+++ b/code/game/gamemodes/miniantags/slaughter/slaughter.dm
@@ -78,6 +78,7 @@
 	icon = 'icons/obj/surgery.dmi'
 	icon_state = "demon_heart-on"
 	origin_tech = "combat=5;biotech=7"
+	decay_time = 0
 
 /obj/item/organ/heart/demon/update_icon()
 	return //always beating visually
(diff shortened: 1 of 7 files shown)


Kmc2000:
So do organ coolers have a use now?

Super3222:
> So do organ coolers have a use now?

Currently, they aren't even in the game. At this time, they don't cool anything either.

Cruix:
The code that the coolers used to cool was broken, but it should be fixed now with a little modification.

TheSadFox:
> rscadd: You can once again perform heart transplants to defibrillate people whose bodies have decayed.

Noice.

Alek2ander:
If you lock someone in a freezer will they die?

Cruix:
No, it is not nearly cold enough to cause damage. Cold showers are much more extreme but do not work on organs because reasons.

Alek2ander:
Well that's unfortunate. Maybe make it colder? :smiling_imp:

Stealthkibbler:
Will transplanting a monkey heart into a human heart work? I mean these are SPACE MONKEYS we're talking about here.

Cruix:
Human and monkey hearts are exactly the same so yes.

PR #463, opened 2016-08-04
Airlock interface for silicons replaced with TGUI

X-TheDark:
Also includes quite a bit of refactoring, namely:
- Replaced a `sleep(10)` based door electrifcation coundown loop with a a simple `world.time`+timer based one.
- **New proc for doors - electrify(mode, what, time)** - used to standardize the electrification and stuff.
- Some other things I probably don't remember.
- Things that have procs for them are now used (mainly isElectrified(), also unelectrify(), instead of direct variable manipulation)

Mind you, the airlock code still can have some attention, especially things like the power being disrupted, it still uses a rather bad `sleep()` based countdown looping thing.

Also fixes a bug where the speed override would cause the door to never close (because the timer is created in the wrong place - it's set for 10 ticks, but there are enough sleeps and stuff that it never is actually ran).

:cl: X-TheDark
rcsadd: Airlock interface for silicons replaced with TGUI.
bugfix: Fixes speed override causing airlocks to never close.
/:cl:
untitled
untitled2


TheSadFox:
> TGUI

Please no, have mercy.

X-TheDark:
It's that bad, huh? Fine, I'll look into maybe-but-don't-get-your-hopes-up rewriting this to use the regular interface thing but make it more pretty.

Notamaniac:
More TGUI!

Make the peasants who can't run tgui suffer in pain and horror.

X-TheDark:
If people vote against this (as in, TGUI is more troublesome to people than I thought), I guess I'll just see if I can rewrite it as a better looking popup thing that doesn't use TGUI.

ShadowDeath6:
> be AI
> try to open door to save people
> Open tgui
> Crash
> Banned for ignoring harm and going braindead as AI

FluffySurvivor:
:+1: hail tgui

Alek2ander:
:no_entry_sign: :sweet_potato: :computer: :gun: :no_entry_sign:
Sometimes before we can usher in the new, the old must be put to rest.
:+1:

X-TheDark:
If TGUI causes problems for your computer, feel free to open an issue, if one does not exist, as to let us know to not TGUI'ify everything and use the standard Topic() system.

Time-Green:
changelog reeeee

oisin100:
To be honest, Not a fan.
TGUI is good and all, It might just be me, But sometimes it can take a heck of a long time to load a page (1-2 Seconds) which is super annoying if you don't know how to use hotkeys as an AI.

X-TheDark:
Then I don't mind this being closed. I can rewrite this into the regular browser/Topic() system.

killerx09:
Yes please.

The current AI interface for the airlocks is ugly and inconvienent to use, this would make them way better.

Alek2ander:
This should still happen. Seriously, I understand some players have issues with TGUI but the rest shouldn't have to be stuck with old static interfaces

PR #471, opened 2016-08-05
Adds a time estimate to cloning completion in cloning pods

X-TheDark:
Quality of Life, aka Power Creep

Also refactors cloning pod process code a bit, namely:
- Replaces the hacky way of putting reagents into mob to cure Oxyloss/Brute loss caused by crit with setBruteLoss/setOxyLoss procs;
- Health is updated only after we run the healing loop in process(), instead of each time we call adjustXLoss/setXLoss procs (not mentioning the updatehealth() called by the reagents that were put into the mob to heal oxyloss/bruteloss);
- NOBREATH mobs are handled better;

Additions to mob/living
- **Adds setBruteLoss/setFireLoss procs** to mob/living, since it already had setOxyLoss, etc.

untitled

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -27,6 +27,7 @@
 
 	var/datum/mind/clonemind
 	var/grab_ghost_when = CLONER_MATURE_CLONE
+	var/no_breath_mob = FALSE
 
 	var/obj/item/device/radio/radio
 	var/radio_key = /obj/item/device/encryptionkey/headset_med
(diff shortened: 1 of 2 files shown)


george99g:
Don't see anything wrong with this.

X-TheDark:
Estimated time is in real-time, not BYOND-time, by the way.

george99g:
Still estimated, so it shouldn't be an issue.

Cruix:
Do the setBrtueLoss and setFireLoss procs handle damage to the individual limbs?

EDIT: nope. Override the procs for humans to use adjustFireLoss and adjustBruteLoss so limb damage is handled.

X-TheDark:
So that's the caveat with why they weren't implemented...no, they do not currently. I will need to implement those to handle limbs for living/carbon+humans, definitely.

EDIT : I'm doing that right now as we speak.

Super3222:
> Quality of Life, aka Power Creep

Nice reference.

Also @X-TheDark, can you now see the percentage and ETA on cloning from a range?

X-TheDark:
Yes, this is added to the `examine()` of the cloning pod, so you can just do the usual examine from range shtick.

PR #447, opened 2016-08-02
Syndicate mob spawn event rework.

Zostroll:
Changes the minimum amount of people from 15 to 20.
Decreases the event weight from 2 to 1. 
Makes the syndicate energy sword mobs be able to be hit by bullets, a 50% chance of laser hits, and decreases there heath from 170 to 130.
Reduces the sound of the event from 100 to 75 to stop it from being so painful for headphone users.
Makes the gear from dead syndicate retrievable. 
Reasoning: https://github.com/yogstation13/yogstation/issues/441
:cl:
tweak: Portal storm event has been reworked and the energy syndicate sword mobs have been nerfed
/:cl:

PR #472, opened 2016-08-05
Removes portal storm event

Time-Green:
Alternative to: https://github.com/yogstation13/yogstation/pull/447/files

This just doesn't fit yogstation. It simply does not work and never will work.


Time-Green:
nvm something broke

PR #476, opened 2016-08-05
Cleans up some unnecessary junk in notes.

AsV9:
Having the server name (chopped up) on every single note is plain unnecessary when you've only got 1 server. I kinda understand it for /tg/ as they have two servers running, but still. So this pull requests removes the `[server]` portion which makes it so that only the timespan shows. The actual sorting by server is still a thing, so if we ever want to re-add it, all we have to do is to put the `[server]` back in it's place.

Also, this is what my notes look like: 
d66d438a958642e18b30f0aa97d8c881

Really messy.

### DIFF
--- a/code/modules/admin/sql_notes.dm
+++ b/code/modules/admin/sql_notes.dm
@@ -136,7 +136,7 @@
 			var/adminckey = query_get_notes.item[4]
 			var/last_editor = query_get_notes.item[5]
 			var/server = query_get_notes.item[6]
-			output += "<b>[timestamp] | [server] | [adminckey]</b>"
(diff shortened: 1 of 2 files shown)


Nirnael:
Purge all M E M E N O T E S while you're at it.

AndySok:
You called bekons UNROBUST???

b&

AsV9:
> Purge all M E M E N O T E S while you're at it.

good luck.

george99g:
There's a note export button so this really doesn't matter.

AsV9:
The note export button also used to contain `[server]`, which is why I had to edit the stuff in topic.dm.

PR #492, opened 2016-08-07
First attempt to make tgui optional

Cruix:
Hopefully a fix for all the people who crash using tgui. At the moment it is very hacked together, but it works and it does not runtime. Needs a lot of review and a lot of old menus ported. At the moment, it only has non-tgui menus for the traitor uplink and the crafting menu.

### DIFF
--- a/code/__HELPERS/type2type.dm
+++ b/code/__HELPERS/type2type.dm
@@ -591,3 +591,16 @@ for(var/t in test_times)
 /proc/list2string( l[], sep )
     for( var/x = 1 to l.len )
         . += ( l[x] + (x < l.len ? sep :null) )
+
+//Converts a string into a list by splitting the string at each delimiter found. (discarding the seperator)
+/proc/text2list(text, delimiter="\n")
+	var/delim_len = length(delimiter)
+	if(delim_len < 1) return list(text)
+	. = list()
+	var/last_found = 1
+	var/found
+	do
+		found = findtext(text, delimiter, last_found, 0)
+		. += copytext(text, last_found, found)
+		last_found = found + delim_len
+	while(found)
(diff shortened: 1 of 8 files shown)


ShadowDeath6:
nice

code/__HELPERS/type2type.dm
```diff
@@ -590,4 +590,17 @@ for(var/t in test_times)
 
 /proc/list2string( l[], sep )
     for( var/x = 1 to l.len )
-        . += ( l[x] + (x < l.len ? sep :null) )
\ No newline at end of file
+        . += ( l[x] + (x < l.len ? sep :null) )
+
+//Converts a string into a list by splitting the string at each delimiter found. (discarding the seperator)
+/proc/text2list(text, delimiter="\n")
```
X-TheDark: BYOND has `splittext(Text,Delimiter,Start=1,End=0,include_delimiters=0)` proc built in.

X-TheDark:
Welp, this is going to not be fun...(porting over all TGUI menus into non-TGUI versions, that is).

Super3222:
This... this will be something. Good luck with it.

Remember that tgui is pretty much a completely rewritten nanoui...
Meaning if you're giving the option of not using tgui, than it would have to use that older nanoui.
And that means going back through almost everything and re-adding nanoui in some fashion to make toggable.

Kmc2000:
CRUIX FOR HEAD CODER!!!!!

We need this

X-TheDark:
@Super3222 
Or just use the old browser/Topic() system, similar to how the Stock Market Computer is programmed (that is not using any NanoUI, if you can believe that, and it looks good).

@Cruix 
Once this is merged or something, I'll help out with writing the alternative versions for the current TGUI interfaces.

TheSadFox:
Oh baby.

ShadowDeath6:
you tried

PR #495, opened 2016-08-07
Fixed bugs with nuclear bombs, photograph descriptions, and the EMP kit description.

Cruix:
:cl:
bugfix: The EMP kit no longer claims to contain two EMP grenades even though it actually contains five.
bugfix: Photo descriptions will now show creatures as being hurt when they are below 75% health, instead of below 75 health.
bugfix: Fixed a bug where the nuke would sometimes not count down.
/:cl:

Fixed the nuke not working if its countdown got deleted. One round the ghost countdown got deleted somehow, and it stopped the nuke from working.

Photos were showing animals with less than 75 health as always being injured.

The EMP kit has way more than two grenades in it.

### DIFF
--- a/code/game/gamemodes/nuclear/nuclearbomb.dm
+++ b/code/game/gamemodes/nuclear/nuclearbomb.dm
@@ -210,7 +210,10 @@ var/bomb_set
 
 /obj/machinery/nuclearbomb/process()
 	if (timing > 0)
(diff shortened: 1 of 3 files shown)


X-TheDark:
5 EMP nades (+ implant) for 2 TC? Are you sure?

PR #498, opened 2016-08-07
Adds Automatic Arrival Announcement System and PDA logging.

Cruix:
Adds logging to the automatic arrival announcement system custom message. Admins also get messaged with a button to revert the message to its default value.
Re-added investigate PDA.


AndySok:
Readd NTSL too ples

PR #193, opened 2016-06-29
Ports lavaland lore and Explorer Gas Mask

X-TheDark:
The most complex thing I've ever attempted : Rebase the commits https://github.com/tgstation/tgstation/pull/18398 and https://github.com/tgstation/tgstation/pull/18432 onto our codebase.

Git is so wicked sick sometimes.

All credit to @Xhuis and @Core0verload

PR #512, opened 2016-08-08
re-adds the golem design disk and golem shells.

### LINKED ISSUE #508: Lavaland spawn names are fucked and the golem creation data disk is missing.
When you spawn as a escaped prisoner you don't get a name you just get "a free golem" or "an escaped prisoner" Also the disk to load the golem shells onto the autolathe is missing from the free golem ship.

Cruix:
This was lost in #193 
Fixes #508


Zostroll:
Did you fix the broken names?

PR #528, opened 2016-08-09
Balance tweaks to Zombie Implants

Super3222:
:cl: Super3222
tweak: (buff) You can no longer help someone lying down, up on their feet if they have capilletum in their blood steam (zombie implant chemical).
tweak: (buff) Zombie implant cost is dropped to 5 from 7. It is definitely in the same tier as freedom implants.
tweak: (nerf) Zombie implants drain a random amount of nutrients from the user while capilletum is in their blood.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -121,6 +121,10 @@
 		M << "<span class='warning'>You can't put them out with just your bare hands!"
 		return
 
+	if(reagents.has_reagent("capilletum") && lying)
+		M << "<span class='warning'>[src] is dead!</span>"
+		return
+
 	if(health >= 0)
 
 		if(lying)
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -129,6 +129,13 @@
 		else
 			M.visible_message("<span class='notice'>[M] hugs [src] to make them feel better!</span>", \
 						"<span class='notice'>You hug [src] to make them feel better!</span>")
+
+
+		playsound(loc, 'sound/weapons/thudswoosh.ogg', 50, 1, -1)
+
+		if(reagents.has_reagent("capilletum") && lying)
+			return
```
X-TheDark: This needs to be at the start of the block, otherwise you still get the problem of `M.visible_message()` being ran before you are `return`ed, resulting in you trying to help up a dead body still.

Super3222: Ah, I see what you mean. Actually, that's pretty silly of me. Wow.

Issue #518, opened 2016-08-09
Engineering East Camera View Missing


00aquir:
There is Engineering east camera.
There is no Engineering east camera on camera list.

SKR6DarkVoid:
@00aquir Can you show a screenshot of where it's missing.

From what I'm looking at, everything is connected to the proper network.

00aquir:
at the engineering lockers right side below the double doors

PR #533, opened 2016-08-10
DANK!  Fixes Camera issues + Janitor Issues.

SKR6DarkVoid:
Fixes Camera issues along with Janitorial issues.

:cl: 
rscadd: Two janitorial roles are now open, but we only have one set, I have added another set of lockers to accommodate for these issues.
bugfix: Fixes an Engineering Camera issue, apparently it was labeled wrong.
/:cl:

I don't know why it wants to merge 9 commits.. please help :( 
#518


X-TheDark:
Dank, indeed. The reason it wants to merge your merge commits is because instead of updating the master by fetch+merging/pulling from the remote repository, you merged into it locally and then uploaded it to your GitHub.

TL;DR : You'd probably need to refork or do something complicated like deleting the 'master' on GitHub/local and then getting it from the remote and pushing that to your GitHub repository.

PR #534, opened 2016-08-10
Fixes antag weights (again)

### LINKED ISSUE #505: Antag weight still not fixed
I was told it was fixed a while ago, but apparently it just never went through.

In case you're out of the loop: antag weight is broken, as it increases properly when you don't get antag but **doesn't decrease when you actually get antag.** So functionally it's pretty much the same as the old system except for being slightly worse for the newest players.

IMO this is a more critical fix than renaming goddamn meatballs.

X-TheDark:
fixes #505 

**Tested with 1/2 clients on traitor mode with 1/2 possible traitors. Everything works properly.**

**Cause:** pick_candidate() was ran as many times as we need candidates. Since this proc picks ONE candidate (decreasing the weight) and increases the weight of those it didn't pick, we ended up in a situation where, because this proc was called multiple times, it increased weights more than once (even if the candidate was picked later).
In fact, math works out in such a way that only the first candidate to get picked actually get their weight decreased (divided by 1.5), 2nd candidate weight didn't change (multiplied by 1.5 on first pick_candidate(), divided by 1.5 on second pick_candidate()), 3rd candidate got their weight multiplied by 1.5 (/1.5 \* 1.5 \* 1.5).

**Fix:** REWRITE FUCKING EVERYTHING. Not really, but still, the fix wasn't as simple as I'd like it to be. Had to add some parameters to make it loop and separate non-picked candidate weight update into another proc.

:cl: X-TheDark
bugfix: Antagonist weights should be working properly now.
/:cl:

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -179,8 +179,10 @@
 		if(istype(player.current,/mob/living/silicon/ai))
 			A = player.current
 
-	if(A.stat == 2 || !A.client || !A.mind) return
-	if(is_special_character(A)) return
+	if(!A || A.stat == DEAD || !A.client || !A.mind)
+		return
+	if(is_special_character(A))
+		return
 
 	ticker.mode.traitors += A.mind
 	A.mind.special_role = "traitor"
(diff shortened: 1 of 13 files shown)


X-TheDark:
Wonder if we should have a "To be reviewed" label...

code/game/gamemodes/game_mode.dm
```diff
@@ -179,8 +179,10 @@
 		if(istype(player.current,/mob/living/silicon/ai))
 			A = player.current
 
-	if(A.stat == 2 || !A.client || !A.mind) return
-	if(is_special_character(A)) return
+	if(!A || A.stat == DEAD || !A.client || !A.mind)
```
X-TheDark, line 182: This fixes a runtime

ParagonPrime:
paragon pls

X-TheDark:
I just pray it works now. It really should, considering I've identified the issue and fixed it, but I still pray.

PR #525, opened 2016-08-09
Allows suit storage's to put clothes on for you

Super3222:
:cl: Super3222
rscadd: Dragging yourself onto suit storage's will put on the equipment for you.
tweak: Emagging a suit storage will activate it's super UV. Letting suit storage's put on your equipment for you, will roast whoever's inside of it alive.
/:cl:

PR #537, opened 2016-08-10
Refactors SSU's to allow more robust extra item definition in storage

X-TheDark:
**This will probably conflict/cause conflicts with #525...**

First of all, shoes now get their own slot (because of how many of them have shoes), SSU's now have these slots:
- Helmet
- Mask
- Suit
- Shoes

In addition, as mentioned, instead of `storage` (which contained 1 extra item of any type), you can now define as many extra items as needed using `extra_types` typepath list. In addition, the SSU generates the max amount of storage slots based on this list (unless you manually override the `extra_items_max` var.

Examples: (left is how it looks when opened, right is how it looks emptied)
`extra_types = list(/obj/item/weapon/tank/jetpack/oxygen/captain, /obj/item/weapon/watertank/atmos)` results in:
untitled1

`extra_types = list(/obj/item/weapon/tank/jetpack/oxygen/captain = 3, /obj/item/weapon/watertank/atmos)` results in:
untitled3

Furthermore, all items are correctly moved around when you take them off and put them back on:
collage

Can you tell yet how proud I am in actually making this?


ParagonPrime:
Lovely. Lovelyyyyy.

Zostroll:
This is great!

PR #538, opened 2016-08-10
Fixes Toxin Testing Range sometimes returning the bomb

X-TheDark:
The bomb strikes back like a boomerang...no more. Apparently the bomb bounces off the indestructible wall and returns to the sender sometimes.

To fix this, I just replaced the tile under the camera with a indestructible one that I modified the air mix and dir variables.


Bekons:
Idunno, sounds like halarious hijinx. Keep the feature

PR #552, opened 2016-08-11
Cloth muzzle

Time-Green:
Adds a cloth muzzle, that can be crafted from a piece of cloth.

Also makes the original muzzle almost twice as fast in applying to make it easier to subdue an enemy.

ks
loos
Credits to Robotic for the idea

:cl:
rscadd: Adds a cloth muzzle, can be crafted from cloth.
:cl:


code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -6,7 +6,7 @@
 	flags_cover = MASKCOVERSMOUTH
 	w_class = 2
 	gas_transfer_coefficient = 0.90
-	put_on_delay = 20
+	put_on_delay = 12
```
george99g, line 9: You didn't mention anything about making the actual muzzle faster to put on.

Time-Green, line 9: Oh yeah, I did it to actually make it usefull in quick combat. And the actual muzzle should be better IMO.

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -16,6 +16,23 @@
 			return
 	..()
 
+
+
+/obj/item/clothing/mask/muzzle/cloth
+	name = "cloth muzzle"
+	desc = "A cloth gag for <strike>sex</strike> silencing your victim!"
```
X-TheDark: How about you don't? There is no reason (or, in my opinion, place) for these ERP memes.

Time-Green: Ok then, didn't think it was that big of a deal, since sex cuffs are litteraly in the game.

Super3222:
Interesting.

EMT321:
At least you can survive some kidnappings with a station bounced radio-making a ubiquitous items that inhibits even that seems a bit much.

I dont see what pressing issue this solves.

Time-Green:
In my eyes, it allows for actual kidnappings, and I hope for less murdering then.

And it'd be logical, but well, logic isn't always the best thing in 2dspessmen

PR #557, opened 2016-08-12
Makes the atmospheric hard-suit more resisted to lasers.

Zostroll:
The atmospheric hard suit has thermal shielding this should be reflected in its resistance to lasers.
:cl: (Zostroll)
tweak: The hard working scientists at centcomm figured out that the same technology that protects from the heat of a fire could when slightly modified offer increased protection from laser blasts.
:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/hardsuit.dm
+++ b/code/modules/clothing/spacesuits/hardsuit.dm
@@ -126,7 +126,7 @@
 	icon_state = "hardsuit0-atmospherics"
 	item_state = "atmo_helm"
 	item_color = "atmospherics"
-	armor = list(melee = 30, bullet = 5, laser = 10, energy = 5, bomb = 10, bio = 100, rad = 0)
+	armor = list(melee = 30, bullet = 5, laser = 30, energy = 5, bomb = 10, bio = 100, rad = 0)
 	heat_protection = HEAD												//Uncomment to enable firesuit protection
 	max_heat_protection_temperature = FIRE_IMMUNITY_HELM_MAX_TEMP_PROTECT
 
@@ -135,7 +135,7 @@
 	desc = "A special suit that protects against hazardous, low pressure environments. Has thermal shielding."
 	icon_state = "hardsuit-atmospherics"
 	item_state = "atmo_hardsuit"
-	armor = list(melee = 30, bullet = 5, laser = 10, energy = 5, bomb = 10, bio = 100, rad = 0)
+	armor = list(melee = 30, bullet = 5, laser = 30, energy = 5, bomb = 10, bio = 100, rad = 0)
 	heat_protection = CHEST|GROIN|LEGS|FEET|ARMS|HANDS					//Uncomment to enable firesuit protection
 	max_heat_protection_temperature = FIRE_IMMUNITY_SUIT_MAX_TEMP_PROTECT
 	helmettype = /obj/item/clothing/head/helmet/space/hardsuit/engine/atmos


ShadowDeath6:
lasers are beams of light not fire
don't buff fireproof hardsuits

Zostroll:
They damage by thermal means and the suit has thermal shielding...

Time-Green:
I mean, it makes sense, but I have no idea of this would be too OP.

Zostroll:
I mean its only 30; the wizard hard suit is 40 and the stock engineering one is 10. It needs to be at least double a engineering hard suit and I think 30 is a good number improved but not OP.

ShadowDeath6:
> it needs to be at least double an engineering hardsuit

realism is not a reason for a balance change like this, atmos hardsuit is already way better considering it makes you completely immune to plasma fire

Super3222:
What the fuck is this.
Lasers are highly concentrated BEAMS OF LIGHT which cannot be compared to a fire.
I'm not sure what to even call this? This PR is just bad. :-1:
Leave it at 10. Not even the CE's hardsuit can absorb more then 10% of laser damage.

TheSadFox:
> I mean its only 30

HoS jacket/sec vest has "only 30".

Zostroll:
The **standard** sec armor has 30 **bone armor has 25** is it unrealistic to have something made to resist heat resist the same amount of heat based damage as armor? No.  This also slows you down a fuckload and is instantly recognisable.

X-TheDark:
Sec Armor and Bone Armor don't make you impervious to plasma fires. There is no reason for this "buff". Hardsuits are what they are - spaceworthy, anti-fire gear. Not combat-ready (and unrobust, due to slowdown) suits of armor.

ShadowDeath6:
make nukeop hardsuits fireproof when?

X-TheDark:
Aren't they fireproof in space-mode?

ShadowDeath6:
no

Alek2ander:
Protection from heat and protection from lasers are entirely different. No need to buff civilian spacesuits armor.

FluffySurvivor:
is this some "muh immersions have to totally destroy balance" pr? 

:-1:

ghost:
Shadow, elite nukeop hardsuits are fireproof

Also.

I'm amazed. Do you legit think that thermal protection from fire = protection from a superconcentrated beam of light? No. That's not how physics works.

Do you want me to explain, bit by bit, how ridiculously powerful this laser must be to be able to actually injure someone? It's going to be several million times hotter than a plasmafire, and concentrated at a single point.

It'd slice through a thermal hardsuit like a knife through butter.

So this isn't a realism change. If anything, this makes it more unrealistic.

ParagonPrime:
i mean, quite technically, if we're talking realism:

-hyposprays (and bullets, and lasers) would permanently puncture the suit (albeit in small amounts) causing very slow decompression
-patches would simply be totally unusable
-space would kill really, _really_ quickly compared to ingame representation
-so on and so forth

And that's all just unfun. (I personally justify the hypospray/patch bit with there being some sort of special port on the suit that can accept them)

Fun fact: _all_ space suits leak air to some degree since it's impossible to 100% seal everything. It's just a very tiny, insignificant amount and space suits are designed to be able to decrease the effects of a puncture somewhat to give the astronaut time to maneuver to a pressurised environment.

aksuduud:
To beat a dead horse, properties which make a material a good thermal insulator do not make it protected from laser pulses, I don't think anyone is willing to argue that the firesuit should be laser protected? Balance wise, I feel this would just lead to a massive increase of the atmos suit being looted for being so robust.

GrayRachnid:
just shut this down fam

oisin100:
Although realism is nice, Balance is nicer.

PR #561, opened 2016-08-13
Fixed wizards not getting their equipment

Cruix:
There was a name collision.


X-TheDark:
Goddamnit...can't believe I missed this one...

PR #573, opened 2016-08-15
Pipboy

Cruix:
Added donator items to the preferences savefile.
Added a pipboy PDA for donators. It fits on the hand slot as well as the belt.
Refactored some clothing code to handle non-gloves fitting on the glove slot.
:cl:
rscadd: Donors now have the option to get a pipboy PDA.
/:cl:
pipboy


Daman453:
Yes

PR #581, opened 2016-08-16
Purrbation 4 donors

Time-Green:
Donors can now start-off with purrbations. (Currently humans only)

purrbation

Also makes purrbation into an easy to use proc that can easily be called.

:cl:
rscadd: Donors can now start-off with purrbation
:cl:


code/modules/client/preferences.dm
```diff
@@ -97,6 +97,7 @@ var/list/preferences_datums = list()
 	var/donor_hat = null
 	var/donor_pda = null
 	var/quiet_round = 0
+	var/purrbation = "No"
```
Cruix: This will work but it should start as a number or the TRUE define for consistency.
Also that string will evaluate to true not false.

TheSadFox:
Speedmerge.

Time-Green:
This is now merge ready.
Safe file works fine.

Sorry for the commit spam, I ended up messing up with the syncing, causing code to get pushed around.

Alek2ander:
On _normal_ servers this is a punishment...

Cruix:
Now you need to change the SAVEFILE_VERSION_MAX, and add sanitation when they are loaded.

EDIT: actually you really don't need to change the savefile version since the default is null, but you should still add sanitation.

ShadowDeath6:
is this removed by monkeying? if so then :+1:

Alek2ander:
You should sanitize yourself any time you touch a catbeast

TheSadFox:
> is this removed by monkeying? if so then :+1:

We should add purrbation gene to genetics as a counter that.

ShadowDeath6:
no

timkoster1:
The sanitation broke the safe. I most likely did it wrong, but I wont be doing any coding soon. Got some other stuff, it'll be fixed at some point within this few days.

Time-Green:
This is good to go. I was told sanitation was needed, but I only broke everything. And as of now, the safe works fine. If it really is needed, tell me.

Xantam-13:
Do it.

timkoster1:
;_; ok then, I'll go find out what sanitation is.

ghost:
BURN THE CATS

Time-Green:
It's done, cruix helped me with sanitation
pls merge

Time-Green:
Unbeknownst to me, this caused the fucking catgirl cancer

Ephemeralis:
I hope you've thought long and hard about what you've done.

PR #589, opened 2016-08-17
fixes the most important error in the antag weight verb

Super3222:
grammar

:cl: Super3222
fix: Grammar error fixed in the antag weight verb.
/:cl:


Time-Green:
It still doesn't actually work, at least for me the verb is still broken

Super3222:
@timkoster1 Eh, it was working for me. It was somewhat accurate but, that may have been because my weight was high compared to other people because I haven't been an antagonist for a loooong while.

The verb also only works when it's a mode that has an antagonist you signed up for.

PR #601, opened 2016-08-19
Fixes medborg hypo not bypassing hardsuits + Fixes incorrect chemical logging for borg hypospray

X-TheDark:
As per title.

### DIFF
--- a/code/modules/reagents/reagent_containers/borghydro.dm
+++ b/code/modules/reagents/reagent_containers/borghydro.dm
@@ -91,7 +91,7 @@ Borg Hypospray
 		return
 	if(!istype(M))
 		return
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
thank

X-TheDark:
Now, just to note, this means you can injects things that normally cannot be injected (golems with their skin, etc), instead of just bypassing thick materials.

The reason I did it this way (rather than fixing how `can_inject` is used, as it has a bypass parameter that wasn't passed correctly) - it's now completely consistent in it's behavior to the regular hypospray.

oisin100:
Well, Golems cant be injected anyway due to their DNA. But you should probably still make use of the can_inject proc.

Have you tried `can_inject(user, 1, penetrate_thick = bypass_protection)`(You will need to typecast to human)

PR #616, opened 2016-08-22
Fixes hub link

oisin100:


code/world.dm
```diff
@@ -158,7 +158,7 @@ var/last_irc_status = 0
 				msg += " (AFK)"
 			msg += "\n"
 		return msg
-	
+
```
oisin100, line 161: Inconsistent line endings *shrug

PR #633, opened 2016-08-24
Fixes Atmospherics Gas mix console

### LINKED ISSUE #559: Mix tank output valve can't be opened
Mix tank output valve can't be opened, nor can you set output pressure. Pressing the output toggle switches input instead. The set pressure button doesn't work at all.

AshCorr:
Fixes #559 

:cl:
bugfix: Atmospherics custom gas mix console now works!
/:cl:

Somehow the mix_out vent was tagged incorrectly.

### DIFF
--- a/_maps/map_files/TgStation/tgstation.2.1.3.dmm
+++ b/_maps/map_files/TgStation/tgstation.2.1.3.dmm
@@ -50468,7 +50468,7 @@
 	dir = 8;
 	external_pressure_bound = 0;
 	frequency = 1441;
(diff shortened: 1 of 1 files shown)


X-TheDark:
Weird, I must have made a mistake of some sort, as both of those were ported over from /tg/ (by me) exactly as they were there (they also had both of in/out as "mix_in") when I was converting the trinary pumps to the new format.

Alek2ander:
/tg/ probably changed it at some point, while our map stayed the same

PR #647, opened 2016-08-25
Ports the cid randomizer detector from tgstation.

AsV9:
Credit to @MrStonedOne at tgstation/tgstation#20058.

When a user's cid doesn't match their last cid, we open a new game window to the server, and close the old one, if it changed again, we disallow the connection.

They are free to try again once they remove the randomizer dll.

Admins are notified on first match for the user, and it will leave a note on the user as long as it hasn't left one recently.

It is a config, defaults on, but requires db before actually activating.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -72,6 +72,8 @@
 	var/forbid_singulo_possession = 0
 	var/useircbot = 0
 
+	var/check_randomizer = 0
+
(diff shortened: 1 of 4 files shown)


ghost:
What does do?

AsV9:
Prevents people from using the cid randomizer.

MrsTonedOne:
I should note, the config defaults to on in the default config(ie, new servers), but if it's missing (config not updated yet) it defaults to off.

PR #664, opened 2016-08-28
Fixed clockcult escape objective always requiring 1 cultist.

Cruix:


ShadowDeath6:
massive nerf, clockcult permanently gimped

PR #695, opened 2016-09-01
The rev flash flashes when it is on cooldown

Cruix:
The sprites were already in, I do not know why it was not doing this already.

:cl:
tweak: the rev flash now blinks when it is on cooldown
/:cl:


Super3222:
Not saying it's intentional or not but, it doesn't "work" while on cooldown. I don't see how it'd flash.

PR #696, opened 2016-09-01
misc borer changes, removes commented out influence shit

ShadowDeath6:
:cl: ShadowDeath6
rscadd: Healthscanners can now detect borers
/:cl:
also makes the revival force the host back into the body because catatonic corpses rising from the dead

some of my shit is out of date so I'll fix that, just PRing it now

EDIT: messed with the changelog.
-Cruix

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -139,6 +139,12 @@ MASS SPECTROMETER
 					if(hack.target == H)
 						user << "<span class='danger'>Unknown harmful microscopic machines detected in subject's bloodstream: Recommend treatment via Electro Magnetic Pulse or Strong Electric Shock immediately!</span>"
 						break
+
+	if(iscarbon(M))
+		var/mob/living/carbon/C = M
+		if(C.borer)
+			user << "<span class='danger'>Foreign organism detected in subject's cranium. Recommended treatment: Dosage of sucrose solution and removal of object via surgery.</span>"
+
 	user << "<span class='info'>Analyzing results for [M]:\n\tOverall status: [mob_status]</span>"
 
 	// Damage descriptions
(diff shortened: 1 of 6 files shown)


ShadowDeath6:
apparently it's actually up to date so

ShadowDeath6:
DNM I'm going to fix a bug tomorrow but I must slep

Super3222:
@oisin100 

Edit: A lot of these changes seem really bad in my eyes, can you make your reasoning more transparent. Some of those things were actually good features which could have been scavenged from the comment out

ShadowDeath6:
Influence was removed before the rebase because it forced players to play one way. As it's a midround spawn (midround antag technically) it shouldn't be forced to play pAI for 3 minutes waiting for its influence to hit 50% and if it drops below that by doing literally anything that isn't being nice the borer can't actually do things. It was a dumb mechanic meant to limit borers and didn't belong in the game.

ShadowDeath6:
also DNM x 2 because I suck ass at coding and it runtimes

Super3222:
It sounds like you don't like them starting as larvas but insist on having the entire hunter.
I don't believe they were suppose to be strong antagonists off the bat, which is pretty evident for the use of influence. Hell, Oisin was even starting an evolution system for them.
More or less, I do see why you wouldn't want it in the game. I'll wager it on how the game is affected by borers.

ShadowDeath6:
> starting as larvas

they can start life weak, but they shouldn't have to gain influence in one specific person to use any of their actual abilities

Super3222:
Yeah, a simple growth system which continually increases within the life() proc. Doubles, if they are inside of a host. Maybe used to control how much units of a chemical gets injected into the user?  For instance, right now it's 10 for each but, reduce it to 5 until they've been around for 2-3~ minutes at _least_. 

After 10 minutes of non-stop borer, they get a special verb to control the amount as if their secrete is a dropper that can inject 5-15 units of a chemical.

I also wouldn't mind increasing the borers chemical bank, specifically to reverse the damage done to punishments. For instance, the eye punishment now temporarily blinds. Too much would lead to complications, you get where I'm coming from?

ShadowDeath6:
r4m

PR #728, opened 2016-09-06
Fixed bullets hanging around when their bullet casing is destroyed.

### LINKED ISSUE #318: Bullets not properly deleting when the gun/magazine is destroyed in lava.
<s>Be a miner. Find prisoner ship. Get hit by 10~ bullets simultaneously, instantly killing you, from an invisible, non existant object according to an admin. Your death is called legittotallintendednotabugyouwonderwhydopeoplehatecurrentadminteam. You wake up on github writing an angry bug report.

Looking at the map in dream maker i cannot see any turrets or the like on ruins template and according to Cruix this isn't something that should happen. Looking around as a ghost i didn't see anything that could possible kill me there either. There was just a corpse of a legion or a prisoner next to me and not a single mob.</s>

See Kor's comment below.

Cruix:
Fixes #318

Reeeeeally hope this does not break anything.

### DIFF
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -26,6 +26,12 @@
 	dir = pick(alldirs)
 	update_icon()
 
+/obj/item/ammo_casing/Destroy()
+	if(BB)
+		qdel(BB)
+		BB = null
+	return ..()
+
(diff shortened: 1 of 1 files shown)


X-TheDark:
This should work.

PR #729, opened 2016-09-06
Added check-ruins verb for admins.

Cruix:
Requested by Adam.

### DIFF
--- a/code/modules/awaymissions/zlevel.dm
+++ b/code/modules/awaymissions/zlevel.dm
@@ -94,6 +94,8 @@ var/global/list/potentialRandomZlevels = generateMapList(filename = "config/away
 				continue
 
 			world.log << "Ruin \"[ruin.name]\" placed at ([T.x], [T.y], [T.z])"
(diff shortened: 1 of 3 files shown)


code/modules/awaymissions/zlevel.dm
```diff
@@ -94,6 +94,8 @@ var/global/list/potentialRandomZlevels = generateMapList(filename = "config/away
 				continue
 
 			world.log << "Ruin \"[ruin.name]\" placed at ([T.x], [T.y], [T.z])"
+			ruinAreas += null
+			ruinAreas[ruinAreas.len] = list(ruin.name, T.x, T.y, T.z)
```
X-TheDark, line 98: Can't this be simplified to
`ruinAreas += list(ruin.name, T.x, T.y, T.z)`?

Cruix: No, that will add each element to the ruinAreas list instead of the list itself.

PR #739, opened 2016-09-09
Summon Magic now Creates Antagonists

### LINKED ISSUE #656: more like summon bullshit
some magic shit you get is either grief or literally round ruining

horsehead- this doesn't do jack shit to wizards, it's just used to be a dumb asshole to your crewmates, don't even lie

staff of change / wand of polymorph- you get hit by this you're permafucked because you can't get cloned if you get griefed, and if you're turned into a xeno you're going to get killed by some dumb asshole, and if you roll drone or something you either beat the shit out of the guy who transformed you and turn yourself back or get bwoinked because admin discrimination says polymorphed drones can / can't retaliate against the dumb griefers who attacked them

it gives the wizard a spell point for making the crew grief each other

INB4 SOMEONE SAYS WIZ CAN GET FUCKED BY MAGIC

RISE! will counter literally all magic, including polymorph provided he doesn't roll something really really bad that deletes on death, I don't even know if that can happen

Cruix:
Possible fixes #656 

Summon magic now costs the same as summon guns and creates adepts instead of survivalists. Adepts have to collect magic artifacts and use them to kill other adepts. They get an antag hud so they know who to kill.
summonmagic

:cl:
tweak: summon magic now costs two points instead of granting an extra point, and creates antagonists with the goal is to kill each other.
:cl:


drderp3635:
nah man summon magic is a meme don't do this

Super3222:
> nah man summon magic is a meme don't do this

It'll forever be a griefers meme but, this is just to give some people flexibility to use their magic tools.
Because on Yog using your magic in self defense was seen as being griefy, which usually got people into arguments and banned.

EMT321:
Why make it cost two points?

X-TheDark:
That's the same cost Summon Guns has, I think, which makes sense, given that it now will act exactly the same way Summon Guns does.

ShadowDeath6:
does wiz see adepts?

Cruix:
No, only adepts see each other.

X-TheDark:
This is totally in accordance to how Summon Guns works, so I see no reason why not.

PR #743, opened 2016-09-10
Normal blob structures will block air

Super3222:

@AdamElTablawy requested this.
Since blobs spawn in maintenance, and almost always cause un-fightable breaches, forcing crewmembers to either risk their lives and die, or find a space suit, Adam's requesting that blob tiles should prevent block air from flowing in or out.

With this idea, blobs will be more easier to fight, and less people will die.

Though in my opinion, I think that blobs causing breaches is just a factor into the mode and, those who want to fight should do the best they can whether it's in maintenance or not.
#### Changelog

:cl:
tweak: Blob structures now block atmos.
tweak: Plasmafire is once again effective against blobs.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/core.dm
+++ b/code/game/gamemodes/blob/blobs/core.dm
@@ -8,6 +8,7 @@
 	explosion_block = 6
 	point_return = -1
 	atmosblock = 1
+	heatblock = 1
 	health_regen = 0 //we regen in Life() instead of when pulsed
 	var/core_regen = 2
 	var/overmind_get_delay = 0 //we don't want to constantly try to find an overmind, this var tracks when we'll try to get an overmind again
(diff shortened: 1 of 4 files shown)


ShadowDeath6:
bad change, this makes plasma flooding, however shitty it already is, borderline useless against blobs and will make the area twice as uninhabitable

Super3222:
@ShadowDeath6 I'll go ahead and do some tests with plasma flooding

Super3222:
Alright, after doing some digging and finding what I needed, I managed to figure out how to make plasma fire fairly effective against blobs.

Do note that it won't do anything against strong blobs like the shielded version or the core.

Supplementary Picture:


You see, I tried to circle (with red) the areas where the blob used to be. Due to the level of plasma fire, the normal blob tiles quickly died. However, the strong version managed to take the heat. Might have to make tweaks here and there in the future.

ShadowDeath6:
Without oxigen blob still evoluates

PR #751, opened 2016-09-11
Fix holodeck safeties href exploit.

george99g:
The holodeck safeties did not have a check if the src is a silicon, so anyone could disable the safeties with an href.
This includes any random griefing assistant.


oisin100:
It might be worth logging future exploits..

george99g:
Will update, one moment.

george99g:
@oisin100 Happy?

PR #752, opened 2016-09-11
Fix mech dominate href.

george99g:
Fixes AIs' ability to dominate mechs via an href exploit without being malf or having the right module ordered.


ShadowDeath6:
what does dominate mech do again

Cruix:
Puts the AI into a mech, kicking out the user if there is one. An alternative to shunting that is way cooler but never used because no one ever builds mechs.

PR #756, opened 2016-09-12
Nerfs Beepsky Smash

ghost:
And with this, the era of stun chemicals is over aside from traitors. 

I don't want any whiny bullshit about 'but bartender so fun xxxd'. Alcohol is already super lethal. If you jab someone full of beepsky smash they die anyway. No stuns needed.

It now heals sec officers a little.
#### Changelog

:cl:
rscadd: Beepsky smash no longer stuns. However, it now heals THE LAW
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -369,7 +369,7 @@ All effects don't start immediately, but rather get worse over time; the rate is
 	if(istype(M, /mob/living/carbon/human) && M.job in list("Security Officer", "Head of Security", "Detective", "Warden"))
 		M.heal_organ_damage(1,1, 0)
 		. = 1
-	..()
+	return . || ..()
 
 /datum/reagent/consumable/ethanol/irish_cream
 	name = "Irish Cream"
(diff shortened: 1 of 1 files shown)


Groudonmaster:
Boooo

EMT321:
I dont think this is the nerf chemistry needs.

ghost:
offer a reason

EMT321:
People are not dying because of stunchems.

They are dying because people can hypo them with deathmix without restraint, and speed past them at mach 5 to do so.

Making hyposprays more ubiquitous (mining medic when) and restricting them from processing certain chems (a la gooncode, which is publicly available in this regard) would solve keek rampages with poison instakilling.

Importing hippies meth nerfs would significantly alter chemistry from being fast and strong, to just strong. Which I feel is fair and balanced. Only additional change we might want is to make another chem for miners stimulants if we also nerfed ephedrine.

Alek2ander:
I prefer rare unrestricted hypos to common but limited.
Mining medic hyposprays were a bad idea

X-TheDark:
What exactly is the reason you want all instant-stuns from chems gone, Adam? What exactly is the problem?

Groudonmaster:
Yeah why are you removing them all?

Groudonmaster:
WHY

Groudonmaster:
THERE WAS CONTROVERSY

Cruix:
I was drunk on beepsky smash at the time

Cruix:
Reverting this so the PR gets 24 hours

ClearedObject:
I gotta agree with EMT, it's not the chems, it's how easy it is to load a cocktail of doom into a hypospray.

PR #770, opened 2016-09-12
Fixed certain things being able to put items in your hand when the limb is dismembered

### LINKED ISSUE #769: Actions that place an item into inactive hand do not check if the hand is missing or not
Steps to reproduce:
1. Lose exactly one arm
2. Find a camera and snap a picture
3. Photo appears in your disabled hand

Same thing happens if you for example craft metal rods.

Cruix:
Fixes #769

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -39,6 +39,15 @@
 		return 1
 	return 0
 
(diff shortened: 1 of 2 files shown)


McDonald072:
Damn that was fast

PR #784, opened 2016-09-14
Misc Fixes

### LINKED ISSUE #774: The laser gatling gun isn't no drop.
You can deploy the laser gatling gun from the back pack power supply and put it on a table. You can then take the backpack power supply off and operate the laser gatling gun without it. I understand its only used in the abuse room but it't kind of bothering me

### LINKED ISSUE #775: You can make legions with no bodys in them.
If you are pulling a body and it gets legioned and then you keep pulling the body will pull out of the legion and leave a legion that if you kill it it will have no body inside but it will drop a soul.

### LINKED ISSUE #777: Banana projectiles are blocked by single pane windows.
If you use the staff of honkmother on a single pane window it will not go through and will just keep spinning in the air.

Cruix:
Fixes #774
Fixes #775
Fixes #777

Fixed honkmother staff not going through single pane windows
Fixed laser gatling gun being able to be put on tables
Made laser gatling gun code a little less stupid
Fixed being able to pull people out of legions
Fixed minebots not eating ore right next to them if they were targeting a different ore further away while being dragged
Fixed animals wandering onto lava to their death


ShadowDeath6:
why is the honk staff an xray taser anyway

code/modules/projectiles/projectile/bullets.dm
```diff
@@ -134,6 +134,7 @@
 	weaken = 5
 	stun = 5
 	forcedodge = 1
+	pass_flags = PASSGLASS
```
Cruix, line 137: This does not fix the actual issue, which is that nodamage projectiles with forcedodge don't seem to be able to go through single-pane windows, but it works.

PR #797, opened 2016-09-15
Adds a 10 second cooldown between mining shuttle calls

X-TheDark:
fixes #nothing
As requested by @AdamElTablawy 

Prevents mining shuttle call spam to permastun/roadkill miners. Also prevents miners from spacing themselves all the time.

Total waiting time between calls is 15 seconds : 5 second transit time + 10 second cooldown
Map doesn't have the circuit anywhere in it placed and it can't be built, so I don't think I broke anything.
### Tests
- Moved shuttle, then tried moving it again instantly - needs to wait;
- Tried deconstructing/reconstructing the console - you'll just have to wait, powergamer-san;

:cl: X-TheDark
tweak: Mining Shuttle can now has a cooldown of 10 seconds between being moved using the Console.
tweak: No, disconnecting and reconnecting the Console's monitor or otherwise meddling with the machine will not help you.
/:cl:

### DIFF
--- a/code/game/machinery/computer/buildandrepair.dm
+++ b/code/game/machinery/computer/buildandrepair.dm
@@ -383,7 +383,7 @@
 /obj/item/weapon/circuitboard/computer/ferry/request
 	name = "circuit board (Transport Ferry Console)"
 	build_path = /obj/machinery/computer/shuttle/ferry/request
-/obj/item/weapon/circuitboard/computer/mining_shuttle
+/obj/item/weapon/circuitboard/computer/cooldown_holder/mining_shuttle
 	name = "circuit board (Mining Shuttle)"
 	build_path = /obj/machinery/computer/shuttle/mining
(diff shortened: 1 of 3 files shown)


Time-Green:
How am I supposed to assasinate that one miner that went alone on the shuttle now?

ghost:
hero

Super3222:
sounds good in honor of mac.

Super3222:
sounds good

Cruix:
I don't think this will fix the spacing problem. The problem with that is that miners fly to the base at the start and just as they arrive and open the door the shuttle gets called back, which causes them to get sucked out while stunned. To prevent this you would need to have a delay after the call button is pressed and before the shuttle moves, preferably with an alarm or other indicator that the shuttle is about to move.

X-TheDark:
I was afraid you were going to say that...I'll close and remake this when I get back then, as I also believe that is the "correct" solution that I was a tad "can't be arsed" to do...

PR #823, opened 2016-09-19
unironically buffs xenos

ghost:

xenos can now melt rwalls

which is fair since the only xeno type actually worth being afraid of, the hunter, has no acid and the rest are slow and weak enough to laser

also it pissed me off that viro hives were impossible
#### Changelog

:cl:
rscadd: Xenomorphs have been sighted using a stronger variety of their corrosive acid. Rwalls may not be enough to stop them.
/:cl:


ShadowDeath6:
sentinels are actually extremely beefy now because they can just mashspam acid. Not a huge fan of this pr

PR #826, opened 2016-09-19
Slows blob expansion, makes blob weaker to fire, lowers blob health slightly

ghost:
Blob now takes more damage from fire.

Blob takes double damage from burn.

Brute is slightly less bad against blob.

Blob pulse range and expand range decreased.

Blob expands and pulses less often.

Normal blob health lowered.

Aliens are more effective vs blobs (wut)

Reasoning? Well, blobs for one currently have a 90% winrate. Burn damage was just vastly ineffective versus blobs-Sure, you did more damage than with brute, but that's only because you did the intended amount of damage.

Normal blob health itself was just a bit too high to be fair.

The major issue with fighting a blob past the first 5 minutes is that it has so many nodes and is pulsing so often it's expansion is just ridiculous. It's covering everything constantly, and you can barely make a dent even with lasers, past like the 10 minute mark. This should change that.

Finally, fire was just more likely to kill crew than the blob. Now, fire should be pretty dang strong.

Of course, this is all experimental, and can be reverted or changed as requested. I'm not going to listen to any 'adam stop nerfing things' comments, 90% winrate is too high. But yes, I am willing to change or modify the PR given feedback.
#### Changelog

:cl:
rscadd: Blobs are significantly weaker to fire and burn damage.
rscadd: Blobs expand less quickly, to give crew a fighting chance.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/theblob.dm
+++ b/code/game/gamemodes/blob/theblob.dm
@@ -13,8 +13,8 @@
 	var/maxhealth = 30
 	var/health_regen = 2 //how much health this blob regens when pulsed
 	var/pulse_timestamp = 0 //we got pulsed/healed when?
-	var/brute_resist = 0.5 //multiplies brute damage by this
-	var/fire_resist = 1 //multiplies burn damage by this
+	var/brute_resist = 0.8 //multiplies brute damage by this
+	var/fire_resist = 2 //multiplies burn damage by this
 	var/atmosblock = 0 //if the blob blocks atmos
 	var/heatblock = 0 // if the blob blocks heatspread
 	var/mob/camera/blob/overmind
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/blob/theblob.dm
```diff
@@ -283,7 +283,7 @@
 	M.do_attack_animation(src)
 	playsound(src.loc, 'sound/effects/attackblob.ogg', 50, 1)
 	visible_message("<span class='danger'>[M] has slashed the [src.name]!</span>")
-	var/damage = rand(15, 30)
+	var/damage = rand(35,50)
```
ShadowDeath6, line 286: I get that aliens and blobs don't happen in the same round, but increasing this number means that aliens can literally mow through blobs stupidly easy since brute resist was lowered

ghost: yes i know

this is intendo

PR #827, opened 2016-09-19
Possible secondary blob nerf 

ghost:
Blob abilities cost more in general.

Blobbernauts cost more.

Blobs can attack less quickly.

Reasoning: Most of the major complaints with blob is how quickly it can just hammer you to dead. This changes that. If this is too severe or just nonfunctional I'll shut it down altogether, but this should give humans and borgs a fighting chance.
#### Changelog

:cl:
rscadd: Blobs can attack less quickly. Blob abilities are increased in cost.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/powers.dm
+++ b/code/game/gamemodes/blob/powers.dm
@@ -110,18 +110,18 @@
 
 /mob/camera/blob/verb/create_shield_power()
 	set category = "Blob"
-	set name = "Create Shield Blob (10)"
+	set name = "Create Shield Blob (5)"
 	set desc = "Create a shield blob, which will block fire and is hard to kill."
 	create_shield()
 
 /mob/camera/blob/proc/create_shield(turf/T)
-	createSpecial(10, /obj/effect/blob/shield, 0, 0, T)
+	createSpecial(5, /obj/effect/blob/shield, 0, 0, T)
 
 /mob/camera/blob/verb/create_resource()
 	set category = "Blob"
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
honestly the biggest problem with blobs is that you can slap down a factory that continually spams out spores that eat 2 lasers / 2 welder hits and has 200 HP, which is as much as a node. Decrease factory health to like 100 and keep blobbernaut cost the same, since you're pretty much sacrificing a factory if the blobbernaut is bad, and it can't be spammed that hard.

ghost:
Aight. I can do that.

code/game/gamemodes/blob/powers.dm
```diff
@@ -236,7 +236,7 @@
 		return
 	if(can_buy(5))
 		var/attacksuccess = FALSE
-		last_attack = world.time
+		last_attack = world.time * 2
```
Cruix, line 239: This will not work, instead look for where last_attack is referenced in /mob/camera/blob/proc/can_attack() to do what you want to do. As it is, the attack cooldown will be increased by however many seconds the since the last server restart.

killerx09:
I prefer a couple of nerfs to go through at a time, not three PRs in an instant.

code/game/gamemodes/blob/powers.dm
ghost: OOPS

ghost:
not needed anymore lmao

PR #859, opened 2016-09-24
buffs explosive holo trap

ShadowDeath6:
their traps last longer and they have a longer cooldown in between setting traps

as always vars adjusted are up for debate

:cl: ShadowDeath6
bugfix: Explosive holoparasite's bomb trap now actually works.
tweak: Increases the duration of the explosive holoparasite's bomb trap, as well as the cooldown between bomb traps set.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/guardian/types/explosive.dm
+++ b/code/modules/mob/living/simple_animal/guardian/types/explosive.dm
@@ -62,32 +62,41 @@
 	anchored = A.anchored
 	density = A.density
 	appearance = A.appearance
-	spawn(1000)
-		stored_obj.loc = get_turf(src.loc)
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
rused

code/modules/mob/living/simple_animal/guardian/types/explosive.dm
```diff
@@ -61,7 +62,7 @@
 	anchored = A.anchored
 	density = A.density
 	appearance = A.appearance
-	spawn(600)
+	spawn(1000)
```
Cruix: You should make this use a timer instead of a spawn.

ShadowDeath6:
DNM, actually, bomb holo traps are completely unresponsive. I'm just about finished with the fix, though

ShadowDeath6:
r4m @Cruix @X-TheDark

PR #863, opened 2016-09-24
Honkcans2

utahclock:
:cl:
rscadd: Honk! Industries has obtained financing to supply the clown and mime with a spraycan.
/:cl:


Time-Green:
What we really need is to give the clown colourful reagent grenades at roundstart.

Cruix:
The clown already starts with so much crap in his backpack, why does he need more?

Super3222:
Just make attacking a spraycan with a rainbow crayon, makes the crayon get "installed somewhere" resulting in a honk can.

And then attacking a spraycan with a silent crayon, does the same thing.

Super3222:
@utahclock 

Seems like clown already spawns with enough gear, and won't be getting anymore. I've already suggested an alternative route. Do you still want this PR up? If not, than I'm closing it.

PR #864, opened 2016-09-24
Medical rework EXTREME WIP

Time-Green:
I decided to PR it already so people could point things out that needs fixing.

The code right now is very broken, BLUNT and SHARP are a thing and do damage, but sharp does not cut nor dismember, and BLUNT does nothing special since I didn't get around to make the bone system.


code/__DEFINES/combat.dm
```diff
@@ -1,6 +1,7 @@
 //Damage things	//TODO: merge these down to reduce on defines
-//Way to waste perfectly good damagetype names (BRUTE) on this... If you were really worried about case sensitivity, you could have just used lowertext(damagetype) in the proc...
-#define BRUTE		"brute"
+//Way to waste perfectly good damagetype names on this... If you were really worried about case sensitivity, you could have just used lowertext(damagetype) in the proc...
+#define BLUNT       "blunt"
+#define SHARP       "sharp"
```
ghost, line 4: wh-

you just removed brute define-

wh-

that's not how this works 

that's not how anything works

Time-Green:
I have no idea why it failed, I am completely up-to-date and it compiles.

code/__DEFINES/combat.dm
Time-Green: You wanted blunt and sharp instead of brute.
Health scanners still display brute, it's sharp_dam + blunt_dam wich togheter is brute.
Havin it still exist in define is unneeded

code/game/gamemodes/blob/theblob.dm
```diff
@@ -203,7 +203,7 @@
 /obj/effect/blob/ex_act(severity, target)
 	..()
 	var/damage = 150 - 20 * severity
-	take_damage(damage, BRUTE)
+	take_damage(damage, BLUNT)
```
ghost, line 206: explosions would do BLUNT and SHARP but sharp is already a flag used for dismembering

Time-Green: I replaced BRUTE with BLUNT, since blunt would be most common, and changed most exceptions by hand. Adding an explosion type damage is extremely difficult. And pretty sure explosions do half blunt and half sharp now

code/datums/datumvars.dm
```diff
@@ -186,7 +186,8 @@
 			body += "<br><font size='1'><a href='?_src_=vars;datumedit=\ref[D];varnameedit=ckey'>[M.ckey ? M.ckey : "No ckey"]</a> / <a href='?_src_=vars;datumedit=\ref[D];varnameedit=real_name'>[M.real_name ? M.real_name : "No real name"]</a></font>"
 			body += {"
 			<br><font size='1'>
-			BRUTE:<font size='1'><a href='?_src_=vars;mobToDamage=\ref[D];adjustDamage=brute'>[M.getBruteLoss()]</a>
+			BLUNT:<font size='1'><a href='?_src_=vars;mobToDamage=\ref[D];adjustDamage=blunt'>[M.getBluntLoss()]</a>
+			SHARP:<font size='1'><a href='?_src_=vars;mobToDamage=\ref[D];adjustDamage=sharp'>[M.getSharpLoss()]</a>
```
ShadowDeath6, line 190: you're better off just having sharp and blunt all under one brute damage type because medicines and shit

Time-Green: I am keeping brute and sharp as it is.
Making brute flag do sharp and blunt is really easy compared to everything else

Super3222:
You might need to load Birdstation on your repo, there's a variable that has something to do with bruteloss on there.

george99g:
```
loading map_files/BirdStation/BirdStation.dmm
_maps/map_files/BirdStation/BirdStation.dmm:21288:error: invalid variable name: bruteloss
```

code/__DEFINES/combat.dm
X-TheDark: Adam wanted to leave BRUTE as it is, but introduce SHARP/BLUNT ways to apply it (think as weapon damage-type flags).

ghost:
You've removed brute damage, which trashes everything.

We already HAVE the sharp weapon flag which is used for dismemberment-it would be easy to also apply the damaging organ effects as well.

Otherwise, things not flagged as SHARP can be flagged as BLUNT, so that objects that cause brute and are BLUNT and not SHARP then proceed to check to fracture, et cetera. 

Instead of...this, where you've actually made 0 progress towards the intended idea. Brute damage remains a define-sharp and blunt are weapon flags for weapons that apply, and SHARP already exists.

ShadowDeath6:
if(sharpness == IS_SHARP)
   brutedamtype = SHARP
else
   brutedamtype = BLUNT

Time-Green:
I already made a similair proc shadow, wich makes all sharp weapons do sharp damage.

And adam, what are you saying? Do you mean that:
Blunt and sharp check for the total amount of brute damage and then apply effect.
Example: if(brute_dam >= 50 && W.sharp)
Then it would apply bleeding? So that it doesn't matter how much of each type you dealt, only the last hit counts?

Time-Green:
I am going to assume you meant the above.

Ephemeralis:
What's even the point in submitting a PR for broken code?

Time-Green:
Well, I PR'd it now and was told to do it in a different way, if I didn't I had put weeks off effort for something that was completely unintended

Adam, please tell me if the above is what you meant, so I can start over and do it that way.

ghost:
No.

I mean weapons themselves would be flagged as BLUNT or SHARP. If sharp, and damage brute requirements are met, apply the effects that SHARP was supposed to apply under a random chance.

If BLUNT, apply similarly.

BRUTE damage still exists. The WEAPONS are the things that are flagged as sharp or blunt and thus apply their respective effects.

ghost:
So basically 

# don't fucking undefine brute.

PR #874, opened 2016-09-25
Rezadone is easier to get

ShadowDeath6:
NOBODY EVER MAKES IT BECAUSE IT TAKES CARP OR KOIBEANS AND NOBODY EVER DOES THAT

it takes capsaicin to make it, which is in chili

:cl: ShadowDeath6
rscadd: Rezadone now needs capsaicin rather than carpotoxin to make.
/:cl:


ghost:
also available infinitely from the security pepperspray dispensers

ghost:
should also cause burn damage on OD

Cruix:
Pepperspray dispensers dispense condensed capsaicin, so that won't work.

PR #877, opened 2016-09-25
Epedededprheinrineldpephrine changes.

ghost:
Epedepehientirhele or however the FUCK you spell it metabolizes faster.

You can become addicted to it easier and overdose easier.

Overdose and addiction effects are more severe.

You become less resistant to stuns outright, but recover stamina damage far faster.

I might give it the IGNORESLOWDOWN flag, so that it does have that effect too, in compensation for some of the nerfs.

Also, all this means is you have to be careful with patch applications to avoid being addicted, but you'll be okay generally. Also, the stamina reduction is niiiiiice.
#### Changelog

:cl:
rscadd: Ephedrine (colloquially known as Epehdenieidpetiroekeeine) is now less of an equivalent to meth, but more effective at negating stamina damage.
/:cl:

EDIT: Fixed spelling in the changelog
-Cruix

### DIFF
--- a/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/medicine_reagents.dm
@@ -485,51 +485,54 @@
 	description = "Increases stun resistance and movement speed. Overdose deals toxin damage and inhibits breathing."
 	reagent_state = LIQUID
 	color = "#D2FFFA"
-	metabolization_rate = 0.5 * REAGENTS_METABOLISM
-	overdose_threshold = 45
-	addiction_threshold = 30
+	metabolization_rate = 0.25 * REAGENTS_METABOLISM
+	overdose_threshold = 20
+	addiction_threshold = 10
 
 /datum/reagent/medicine/ephedrine/on_mob_life(mob/living/M)
 	M.status_flags |= GOTTAGOFAST
-	M.AdjustParalysis(-1, 0)
-	M.AdjustStunned(-1, 0)
-	M.AdjustWeakened(-1, 0)
-	M.adjustStaminaLoss(-1*REM, 0)
+	M.AdjustParalysis(-0.5, 0)
+	M.AdjustStunned(-0.5, 0)
+	M.AdjustWeakened(-0.5, 0)
+	M.adjustStaminaLoss(-2*REM, 0)
 	..()
 	. = 1
 
 /datum/reagent/medicine/ephedrine/overdose_process(mob/living/M)
-	if(prob(33))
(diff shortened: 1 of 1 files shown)


BOBsHERE:
so you can take more shots from a disabler before going down?
and can we get some numbers on this? whats the OD threshhold and addiction threshhold now?

ghost:
read the code

BOBsHERE:
okay thats a HUGE nerf/ make the addiction threshhold 20, and OD 25

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -485,51 +485,54 @@
 	description = "Increases stun resistance and movement speed. Overdose deals toxin damage and inhibits breathing."
 	reagent_state = LIQUID
 	color = "#D2FFFA"
-	metabolization_rate = 0.5 * REAGENTS_METABOLISM
-	overdose_threshold = 45
-	addiction_threshold = 30
+	metabolization_rate = 0.25 * REAGENTS_METABOLISM
```
X-TheDark: Actually. this makes it metabolize twice as **slow**

ghost: FUCK

ghost: wait are you sure because longterm stims use a metab rate of 2

ghost:
no, not without a reason

BOBsHERE:
You also need to go change the miner stim pack, as it will now cause them to die a horrible death.

ghost:
right okay good point

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -485,53 +485,62 @@
...
+	overdose_threshold = 20
+	addiction_threshold = 10
```
ShadowDeath6: miner stimpacks inject 10u. Make it 20 as well

ghost: alright fine

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
X-TheDark: Yes, Chloral Hydrate has metabolization rate of 1.5 and that gets metabolized super fast, plus, longterm stim chempack thing has 300 units of the stuff, which is why it lasts so long.

ghost:
WHY

PR #872, opened 2016-09-25
Cryo no longer autoejects or shows reagents in the inserted beaker.

ghost:
I DON'T KNOW WHAT THE ACTUAL FUCK I'M DOING I THINK I BROKE IT BUT HOPEFULLY NOT
#### Changelog

:cl:
rscadd: Cryogenics machines have had the autoeject feature disabled, Nanotrasen accounting citing 'expenses' as the reason.
rescadd: Beakers no longer display their reagents inside the cryogenics tube.
/:cl:

PR #887, opened 2016-09-26
Revert "Cryo no longer autoejects or shows reagents in the inserted beaker."

Cruix:
Reverts yogstation13/yogstation#872


EMT321:
Thanks. Its currently possible to get stuck in cryo forever, as you are asleep due to temperature and cant wake up to resist out.

Cruix:
Also the cryo.ract file was not updated properly and I don't have time right now to fix it, install node.js, and compile it.

killerx09:
Keep the "not showing reagents in beaker", but put autoeject back in (possibly as a setting only for medical access)

PR #892, opened 2016-09-27
Adds the worst escape shuttle ever

Tenryumi:
This is not only the shittiest escape shuttle of all time, but is also my first attempt at both a PR _and_ a SS13 map. Please excuse any inevitable mistakes I make in both this PR and the shuttle itself.
:cl:
rscadd: After spending a redacted amount of resources on a party while an emergency shuttle was being constructed, Centcomm has been forced to salvage the shuttle's valuable components during its construction and replace them with more 'cost efficient' parts.
/:cl:


Time-Green:
Could you give us a picture?

_maps/shuttles/emergency_budget.dmm
```diff
@@ -0,0 +1,54 @@
...
+"u" = (/obj/structure/grille,/obj/effect/spawner/structure/window/reinforced,/turf/open/floor/plating,/area/shuttle/escape)
+"v" = (/obj/machinery/door/airlock/shuttle{name = "Emergency Shuttle Airlock"},/obj/docking_port/mobile/emergency,/turf/open/floor/plating,/area/shuttle/escape)
+"w" = (/obj/structure/chair{dir = 4},/turf/open/floor/plasteel/grimy,/area/shuttle/escape)
+"x" = (/obj/structure/chair{dir = 8},/turf/open/floor/plasteel/grimy,/area/shuttle/escape)
+"y" = (/obj/structure/chair{dir = 4},/obj/structure/window,/turf/open/floor/plasteel/grimy,/area/shuttle/escape)
+"z" = (/obj/structure/window,/turf/open/floor/plasteel/sepia,/area/shuttle/escape)
+"A" = (/obj/machinery/door/window/southright,/turf/open/floor/plasteel/sepia,/area/shuttle/escape)
+"B" = (/obj/structure/chair{dir = 8},/obj/structure/window,/turf/open/floor/plasteel/grimy,/area/shuttle/escape)
+"C" = (/obj/machinery/door/window/southleft,/turf/open/floor/plasteel/sepia,/area/shuttle/escape)
+"D" = (/obj/machinery/sleeper{icon_state = "sleeper-open"; dir = 8; step_x = 31},/turf/open/floor/sepia,/area/shuttle/escape)
```
X-TheDark: Replace `step_x = 31` with `pixel_x = 31` (just manually change the name in any text editing thing like Notepad++ or something)

The reason is that if there's anything at all on the map with `step_x` (or `step_y`) defined, it **globally** disables gliding (the transition you seen when you move from one tile to another), meaning that every move is instant (you don't have any animation, you just appear on the next tile you move to instantly). Needless to say, this is very, very, very bad, because it makes movement look incredibly choppy.

_maps/shuttles/emergency_budget.dmm
```diff
@@ -0,0 +1,54 @@
...
+"E" = (/turf/open/floor/sepia,/area/shuttle/escape)
+"F" = (/obj/machinery/sleeper{icon_state = "sleeper-open"; dir = 8; step_x = 29},/turf/open/floor/sepia,/area/shuttle/escape)
```
X-TheDark: Replace `step_x = 29` with `pixel_x = 29`

See the reason above.

Tenryumi:
dreammaker_2016-09-26_18-25-07
This is what I'm trying to add. 

WOW, I did not know that about that difference between 'step' and 'pixel'. Thank you, I'll change that at once.

X-TheDark:
Problem is that **you can't use Dream Maker's nudge function at all, since that modifies `step_x/y` parameters**. You need to manually change the variable using the right click menu thing.

ShadowDeath6:
I'd recommend making the bridge one tile higher, so you have room to move around instead of being locked in 2 seperate parts. Other than that, it looks good for what it's intended to be.

Tenryumi:
> I'd recommend making the bridge one tile higher, so you have room to move around instead of being locked in 2 seperate parts.

That's part of the idea! It's designed to be a terrible uncomfortable mess of a ship with extremely dated designs.

ShadowDeath6:
fair enough, but keeping that will make it less novel and more irritating to use. I see where you're coming from, though

Kmc2000:
Grozx

Karinrine:
Might be perfect for mini-events where the original emergency shuttle suffers a fatal issue and Centcom sends this shuttle as its last resort.

> Central Command Update: Sorry, looks like our shuttle rammed into an asteroid. We'll look into the problem. While we do, have this horrible shuttle.

Super3222:
That'd be(e) great.

oisin100:
Try adding a changelog to the OP

```
:cl:
rscadd: Nanotrasen added a new shit-tle
/:cl:
```

ShadowDeath6:
I think messing with your master is causing some problems

Tenryumi:
I'm just gonna redo this PR. I made it from my master and will be remaking it on a new branch like I should've.

PR #893, opened 2016-09-27
Revert "Revert "Cryo no longer autoejects or shows reagents in the inserted beaker.""

ghost:
Just resist out, you morons.

Easymode game can go die. If people have 'uniform disapproval' they can go comment on the github where I can address their concerns and fix it according to feedback.

if they're too lazy I don't give a fuck about any votes.

Resist time is fine, cryo used to be this way. It gives 'instant omnizine allheal tube' some risk in being used.

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -11,7 +11,7 @@
 	var/volume = 100
 
 	var/efficiency = 1
-	var/sleep_factor = 750
+	var/sleep_factor = 1000
 	var/paralyze_factor = 1000
 	var/heat_capacity = 20000
 	var/conduction_coefficient = 0.30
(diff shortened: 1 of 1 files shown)


EMT321:
You _literally cannot_ resist out. Your body temp stays low enough you are placed in permanent sleep so long as the cryo is hooked up.

Additionally, the playerbase has voted against it 3 times so far.

ghost:
then the playerbase can go comment on github because a vote without context is the same as saying 'no' with no reason to me

tell them to go comment and i can address their concerns or modify it according to their wishes

also, the resist shouldn't cancel if you're asleep. That sounds like a bug @EMT321. I'll try to fix that.

EMT321:
Please do _not_ speedmerge this change.

It is currently bugged, and the playerbase is vocally against it as evidenced by over half the active server pop voting against it on 3 separate occasions. This requires more dialogue then just you and I.

X-TheDark:
I would like to make Cryo less of a "pop-in-and-full-heal-lol" thing that absolutely anyone can do at any time, which completely and utterly obsoletes any Medical job we have on the server.

Also, no, any kind of `do_after` does actually check for your consciousness, so you cannot resist out while being unconscious and becoming so during the resist period will interrupt it.

In reality, I'm actually all for this revert of a revert. There is absolutely, positively, 100% no way to make Medical jobs useful/better without making it a pain-in-the-ass for most of the playerbase. You can't do anything about this. Anything to make Medical jobs better will make Medical worse for everyone else. That's just the way it is and always will be.

The only way to make Medical jobs better without making it worse for everyone else is to make Medics be able to do surgeries that apply timed buffs to people (faster speed/damage resistance/etc).

Cruix:
The problem is that the .ract file was not updated properly, so the interface claims that the beaker is empty when it is not. You also cannot resist out because cryo never turns off and keeps you asleep forever.

ghost:
yes i know that is a bug

@EMT321 speedmerging literally does not occur on yog

speedmerging is merging a PR like within a day of it being made

all of my prs sit for a week or so

ghost:
also i'm not doing this to make med better

i'm doing it to make cryo more risky so that people will kill you in it as a legit strat

being interrupted when asleep i am fixing

X-TheDark:
I mean, it makes absolutely no sense for you to be able to resist out of something while you are asleep.

ghost:
you should still be able to leave for balance concerns

ghost:
Fuck you

oisin100:
> Fuck you

Sorry adam, Im just using this PR as my testing zone.

ghost:
then go test how i can make you keep resisting when asleep only for this

ShadowDeath6:
maybe make it paralyze instead of sleep?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
you could add a check in the resist verb if the person is in cryo and let them out after some time

ghost:
working on that.

Issue #762, opened 2016-09-12
BZ isn't even fucking logged


AdamElTablawy:
1. release BZ

2. no logs at all even in atmos

00aquir:
and atmos scanner cannot see it in canister

PR #896, opened 2016-09-27
Added gas canister logging to BZ

Cruix:


Cruix:
Fixes #762

killerx09:
We should merge the "Removes BZ from station" pull that /tg/ has. Right now, using it as a non-antag gets you banned, using it as a team-antag gets your teammates cucked and you banned, and using it as a solo-antag is often not destructive enough to warrant using.

PR #900, opened 2016-09-27
Moves around some stuff in Engineering that gets spaced every round

Tenryumi:
When I play as an engineer, I notice every round that these objects get spaced because they're directly next to an airlock.
**# Old placement** (affected objects are circled in pink)
engineering stuff

So what I've done is move them elsewhere , where they won't be doomed to spend their time next to Lord Singuloth for eternity (and maybe see more use because of their more convenient location.)
**# New placement**
engi new


ShadowDeath6:
seems like some of #892 is in this PR

Tenryumi:
Oops. I thought I'd fixed that.
Sorry, I'm still learning how to use Git.

Super3222:
It's alright. Try using your `master` branch as the head next time.

PR #919, opened 2016-09-29
Revert "Made the singularity attracted to mass."

ghost:
It was reasonable to try.

It was less reasonable to keep. Singulo being released is an immediate, game ending GG that also makes the power beacon kinda useless all things considered.

And yes, sing being released should be a game ending GG-and it normally is-but singulo bullrushing the AI and the entire station chewing every particular corner is a bit much.

### DIFF
--- a/code/modules/power/singularity/singularity.dm
+++ b/code/modules/power/singularity/singularity.dm
@@ -24,7 +24,6 @@
 	var/event_chance = 15 //Prob for event each tick
 	var/target = null //its target. moves towards the target if it has one
 	var/last_failed_movement = 0//Will not move in the same dir if it couldnt before, will help with the getting stuck on fields thing
-	var/favored_dir = 0 //The direction with the most objects to eat. Due to the nature of get_dist(), this is heavily weighted toward diagonals.
 	var/last_warning
 	var/consumedSupermatter = 0 //If the singularity has eaten a supermatter shard and can go to stage six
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
:^)

BOBsHERE:
How about giving it a 1/4 chance of being attracted to mass when it moves? Then it may still eat the entire station, but it will be a slower death. Also, i wish there was a better way to stop a singularity. Right now the tech on BOH's is really high and is rarely achieved nowadays.

Cruix:
As it is it has a 60% chance of being attracted to mass per movement, so making it 25% would not change too much.

ShadowDeath6:
make it .00000000000000000000000000000000000000000000000001%

Super3222:
So I sorta predicted the outcome.

Ephemeralis:
Adjust the chance to 1/6 or 1/8 and it should be more reasonable. I can forsee the sing lingering around the fringes of the station that way, still causing serious damage while keeping the core of the station intact.

BOBsHERE:
just had the singularity released two rounds in a row. If it gets free right now the station is doomed.

EMT321:
Why not decrease the singularity speed by say, 1/4, and maintain current tracking? Make it a slow unstoppable force that can be ran from.

ghost:
@EMT321 that would a. Make eball and nar'sie weird and b. Still have sing beacon useless

Issue #909, opened 2016-09-28
Miners having QM Office Access


Dzlan:
A miner had access to the QM office, immediately after joining.
They should not. It was during a low population (8 ish people) round.

timkoster1:
Everyone has all acces at round start because the subsystem didn't start, everyone is invincible then aswell.

Dzlan:
No, this is throughout the entire round. Miners have access to QM office.

timkoster1:
Forget what I said then.
Likely because of the skeleton crew

AdamElTablawy:
@Dzlan 

Skeleton crew-AKA extremely low pop such as 8 or so-Gives people access they do not normally  have.

Dzlan:
I am aware of that Adam, but that does not mean they should be given access to their superiors office. There is absolutely no reason a miner requires access to the QM's office.

AdamElTablawy:
it does on lowpop.

Dzlan:
Can this be fixed? That shouldn't need to happen. There is nothing in the QM office for miners, at all.

EMT321:
It's meant for lowpop, when there might be just a single miner running cargo.

Feature, and a good one at that.

Dzlan:
So, why do they require QM office access?
Why don't they go to the HoP like Intended. A miner can run cargo with legitimate cargo access during low pop, which I agree is a nice feature. But they do not require access to the QM's office to do so.

EMT321:
Because when there are <15 people, and no HoP, that's an impossible task.

This is an edge case, only applicable on super-lowpop. The pros (increased access so 1 person departments dont stagnate) outweigh the cons (Possible sharing of space between multiple people).

Dzlan:
I disagree plainly because, it is not needed for the miner to have "QM Access to do the cargo job."
"Cargo access works perfectly fine."

EMT321:
Someone running cargo could also need the door remote and top level PDA cartridge. Both of which spawn in QM's locker.

Groudonmaster:
The real question is, is this really a problem? What's in the QM's office that is round altering? Absolutely nothing is what.

Time-Green:
It's a personal space.

Groudonmaster:
That's a retarded argument and you know that Time

Groudonmaster:
If somebody tries to break into your office with access just telebaton and beat them with said telebaton until they get the hint to fuck off

Time-Green:
How?
[...]

PR #929, opened 2016-10-02
Shaft Miners & Cargo Techs can no longer get QM during skeleton crew situations

Super3222:
Refer to issue https://github.com/yogstation13/yogstation/issues/909.

They might need it on /tg/, but we don't need it on Yogs.

Cargo Techs don't need the cargos door remote just like how Medical Doctors don't need medbays door remote.

If it doesn't matter, than don't bother saying anything.

As I said in the previous issue, I'm still skeptical of revoking Cargo Technician's access truthfully. But don't bother defending Shaft Miners.
#### Changelog

:cl:
rscdel: Shaft Miners and Cargo Techs can no longer obtain QM access when there's a skeleton crew situation.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/cargo_service.dm
+++ b/code/modules/jobs/job_types/cargo_service.dm
@@ -44,7 +44,7 @@ Cargo Technician
 
 	outfit = /datum/outfit/job/cargo_tech
 
-	access = list(access_maint_tunnels, access_mailsorting, access_cargo, access_cargo_bot, access_qm, access_mining, access_mining_station, access_mineral_storeroom)
+	access = list(access_maint_tunnels, access_mailsorting, access_cargo, access_cargo_bot, access_mining, access_mining_station, access_mineral_storeroom)
 	minimal_access = list(access_maint_tunnels, access_cargo, access_cargo_bot, access_mailsorting, access_mineral_storeroom)
 
 /datum/outfit/job/cargo_tech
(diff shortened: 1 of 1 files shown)


ghost:
what the fuck skeleton crew is 8 people

there's never a QM even on highpop

just because dzlan is whining doesn't mean you need to revert this, door remotes are useful if you're running cargo

Super3222:
Shaft Miners should never have access to the QM's office. They are suppose to be mining... and they can practically **take over** cargo with access to the bay by spending 5~ minutes mining, putting materials in a crate, and shipping it off. Above all, it's really just not their place.

I'm undecided with Cargo Techs because I CAN see them needing the remote with a skeleton crew. However, simply making a PR because of how I only feel would be bad. So, give pace to discussion.

EMT321:
People should learn to share in the edge case this happens on lowpop.

The door remote is a useful tool, as is the PDA cartridge.

If there is nothing of value in there, nobody will go in, and its a non-issue.

If there is something of value in there, it makes sense for any lone person operating cargo to have access, be they a CT or miner.

Super3222:
It's a consistency issue. Remotes are appointed to the people that run departments, and frankly the QM arguably runs cargo.

If Medical Doctor's don't need the medical door remote to operate Medbay, than Miners don't need the cargo door remote to operate Cargo.

EMT321:
I believe that the medical doctors do, in fact, get CMO access on ultralowpop.

Super3222:
They don't. But believe me, they do get more than normal access. Anyway, the point is delegates of each division are offered tools and items that represent their status. Just like the telescopic batons and bridge access that heads get, it's only trusted within the hands of people that volunteer to man that position. The door remotes are no exception.

PR #937, opened 2016-10-03
Fixes cult walls of all kinds

MacHac:
This PR tweaks the functionality of both blood and clockwork cult walls, because constructs spamming unremovable walls was getting annoying.
# NEW (and old) SHIT:
### Blood Cult

Blood Cult walls can now be removed with a standard welder.  Cultists do this twice as fast as a normal wall.  Girders can be sliced apart with a welder to drop runed metal.  As usual, cult walls can be hit with a tome to make them see-through and walk-through.
### Clock Cult

Clock Cult walls can be removed with a standard welder.  Clock cultists do this in half the time of a regular metal wall.  Non cultists take twice as much time.  Clockwork cultist girders can be unsecured and moved around, or smashed until they yield alloy fragments to remove them.
## WHAT THIS MEANS

You can now get rid of cult walls without blowing them up.  Without this, culties can just have the artyfacers wall off their summoning position and kill toxins, and you're forced to welderbomb in.  Also, flavortext.

:cl:
tweak: Cultists now remove cult walls faster.
tweak: Clockwork cultists now remove clockwork cult walls faster.
tweak: Non-cultists now take noticeably longer removing clockwork cult walls.
bugfix: Cult walls of all types can now be destroyed by anyone with a welder, although effectiveness varies based on cult membership.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/walls_misc.dm
+++ b/code/game/turfs/simulated/walls_misc.dm
@@ -75,6 +75,21 @@
 			alpha = initial(src.alpha)
 			density = 1
 			opacity = 1
+	else if(istype(W,/obj/item/weapon/weldingtool) && iscultist(user))
+		var/obj/item/weapon/weldingtool/WT = I
+		if(!WT.remove_fuel(0,user))
+			return 0
+		playsound(src, 'sound/items/Welder.ogg', 100, 1)
+		user.visible_message("<span class='notice'>[user] begins channeling energy into [src]...</span>", "<span class='notice'>You begin undoing the magics protecting [src]...</span>")
+		if(!do_after(user, 60 / WT.toolspeed, target = src))
+			return 0
+		if(!WT.remove_fuel(1, user))
+			return 0
+		user.visible_message("<span class='notice'>[user] breaks apart [src]!</span>", "<span class='notice'>You break apart [src]!</span>")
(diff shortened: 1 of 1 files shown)


Cruix:
This is an awful lot of copypasted code. Could you maybe have a proc in turf/closed/wall that takes the user and the tool and returns the time it takes to slice the wall apart, and another that gives the text for the visible_message, then only override those in the cult and clockwork versions?

ShadowDeath6:
make sure turf/closed/cult/artificer walls or whatever they are don't drop metal so artificers can't mass produce metal

Super3222:
> make sure turf/closed/cult/artificer walls or whatever they are don't drop metal so artificers can't mass produce metal

Artificer walls don't drop rune metal.
However, the feature you earlier made where Artificers could "corrupt" normal walls into cult walls isn't pathed for artificer walls. It's pathed for cult walls. Which in turn, can be broken down for cult metal.

Which is a bad idea... even if it's only popping out one or two rune walls. Mac'd need to add a few lines under `attack_animal` or something.

Super3222:
@MacHac Seems like you've gone inactive. If you're still working on this PR, than please answer. Otherwise, I'll be closing this PR **very shortly.**

PR #941, opened 2016-10-03
Minedrone Tweaks

Cruix:
:cl:
tweak: Minedrones now immediately pick up ore they move over, stop moving toward their target when you start dragging them, and can be moved through and nuzzle when they are on idle
/:cl:

Also added a generic on_pulledby() proc for when an object starts being pulled.


ClearedObject:
Finally, minedrones aren't useless now

PR #945, opened 2016-10-05
Added the ability for pais to accept radio encryption keys

Cruix:
:cl:
rscadd: Pais can now accept radio encryption keys.
tweak: Pais now have a notice in the radio software menu when their radio transmitting or receiving is disabled.
tweak: Pais can now set the range that their radio microphone and speaker reach.
/:cl:


Ephemeralis:
Nice.

Super3222:
This is good for traitors that have TC to spend for their evil pAI. :+1:

Cruix:
Added the ability for Pais to set their radio's microphone/speaker range so you don't get ousted when your pai starts spouting all the secure channels and the syndicate one.

PR #830, opened 2016-09-19
Updated PDAs and PDA cartridges.

Cruix:
:cl:
rscadd: Medical and security records on PDAs now have a search function.
rscadd: Quartermasters can now view but not buy or sell stocks from their PDAs.
rscadd: Cargo technicians can now use their PDA as an export scanner.
rscadd: Medical PDA cartridges now have a reagent scan option, identical to the health analyzer's reagent scan mode.
fix: The refresh button on PDAs now actually works for all menus.
fix: The RD's PDA cartridge can now use its signaler system.
fix: Fixed some PDA images that were not loaded properly.
/:cl:

I also added a few options for PDAs that I did not add to any cartridges except the admin one for balance reasons. There are now cartridge options to make the PDA into a forensic scanner and to receive motion alarms and burglar alarms.

PR #955, opened 2016-10-07
Made universal recorders remember languages and spans.

Cruix:
I was working on some language stuff and decided to refactor universal recorders. At the moment all the language remembering they do should have no significant effect, since they understand, speak, and translate all languages, but it opens up interesting possibilities like ruins that have recorded messages in a particular language.

:cl:
rscadd: Tape recorders now remember the font (but not the font size) of messages they record.
/:cl:

### DIFF
--- a/code/__HELPERS/lists.dm
+++ b/code/__HELPERS/lists.dm
@@ -393,3 +393,7 @@
 
 //Picks from the list, with some safeties, and returns the "default" arg if it fails
 #define DEFAULTPICK(L, default) ((istype(L, /list) && L:len) ? pick(L) : default)
+
+/proc/add_list_to_list(list/L, list/E)
+	L += null
+	L[L.len] = E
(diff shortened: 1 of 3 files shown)


Super3222:
Another refactor, another week added until the official release of cybermen.

code/__HELPERS/lists.dm
```diff
@@ -393,3 +393,7 @@
 
 //Picks from the list, with some safeties, and returns the "default" arg if it fails
 #define DEFAULTPICK(L, default) ((istype(L, /list) && L:len) ? pick(L) : default)
+
+/proc/add_list_to_list(list/L, list/E)
+	L += null
+	L[L.len] = E
```
X-TheDark, line 399: Why do it via a proc when you can simply do `list += another_list`?
Or `list = list | another_list`?

Cruix: Both of those will add all the elements of the second list to the first list, not add the actual list. If I did either of those instead of this (and I did, during testing), the parallel lists would go out of sync and there would be indexOutOfBounds exceptions.

X-TheDark: Ooooh, you're adding the actual list as an element, gotcha.

ShadowDeath6:
https://github.com/yogstation13/yogstation/pull/830#issuecomment-248154963

do you remember

X-TheDark (approved):

Super3222:
> do you remember

Never forget.

PR #968, opened 2016-10-10
Implements differently stocked YouTool vendor variant for public use

Tenryumi:
I've noticed many rounds now that Primary Tool Storage becomes completely empty very often nowadays. Since we started becoming more cool and hip to play on more people have been participating in rounds. 
That makes for less tools available (if any at all!) for late joiners or people whom don't make a mad dash for the room less than 5 minutes into the round.

So, I've moved around some tables to make way for the solution to that problem!
new tool
The premium coin items appear below "Engineering Gear."
_not pictured: the hacked items are an industrial welder (no cool upgraded welder here, sorry) and budget insulated gloves._
#### Changelog

:cl:
adds Public YouTool, stocked with the lame, mostly non-engineering tier tools
moves some tables around in Primary Tool Storage to make way for its placement
/:cl:

PR #969, opened 2016-10-10
Implements differently stocked YouTool vendor variant for public use

Tenryumi:
I've noticed many rounds now that Primary Tool Storage becomes completely empty very often nowadays. Since we started becoming more cool and hip to play on more people have been participating in rounds. 
That makes for less tools available (if any at all!) for late joiners or people whom don't make a mad dash for the room less than 5 minutes into the round.

So, I've moved around some tables to make way for the solution to that problem!
new tool
The premium coin items appear below "Engineering Gear."
3c317684-8ead-11e6-82f2-f98f30be635b
_not pictured: the hacked items are an industrial welder (no cool upgraded welder here, sorry) and budget insulated gloves._
_also not pictured: **CORRECT PLACEMENT OF THAT LIGHT FIXTURE** COUGHCOUGH_ #968 

Changelog

🆑
rscadd: Primary Tool Storage now has a Public YouTool, stocked with the lame, mostly non-engineering tier tools
tweak: moves some tables around in Primary Tool Storage to make way for the new vendor's placement
/🆑


ShadowDeath6:
why does the youtool need a coin that should be used on (redacted) for stuff that you can get normally

also you could just put the youtool on the tile directly below the vendomat instead of moving the tables

also also, there's a bugged light on top of the middle table, not a huge fan of the tables being moved either, it makes random empty space on the right side of the room and constrains space in the middle where people should have room

Tenryumi:
Nobody seems to care for this anyway.

PR #972, opened 2016-10-10
Adjusted blood transfusion chemicals and IV drips.

Cruix:
Tweaked saline glucose to make it less completely useless and iron to make it less overpowered as a free source of effectively O- blood.

:cl:
tweak: Saline glucose's ability to generate blood when injected has been adjusted. At 50% blood level, it now takes 1 unit of salglu to generate one unit of blood, and at 100% blood level it takes 2 units of salglu to generate one unit of blood. Effects are significantly reduced when injecting quantities larger than 2.5 at a time.
tweak: Iron's capacity to regenerate blood has been reduced from 1.25 units of blood per unit iron to 0.5 units of blood per unit iron.
/:cl:

Also adjusted how iv drips work, since they did this weird thing where they would transfer 5-10 units at a time but only have a small fraction of this, based on the total reagents left in the container they were draining from, do the reaction.

The salglu changes were requested by Puremanbird

### DIFF
--- a/code/game/machinery/iv_drip.dm
+++ b/code/game/machinery/iv_drip.dm
@@ -7,6 +7,8 @@
 	var/mob/living/carbon/attached = null
 	var/mode = 1 // 1 is injecting, 0 is taking blood.
 	var/obj/item/weapon/reagent_containers/beaker = null
+	var/transfer_amount = 2.5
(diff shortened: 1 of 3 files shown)


ShadowDeath6:
was iron really used so much that it was deserving of a nerf

PR #977, opened 2016-10-11
pop up

utahclock:
reenables admin popup

:cl:
rscadd: Admin popups are enabled again so they are no longer easily missed.
/:cl:

re enable admin pop up

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -158,7 +158,7 @@ BANAPPEALS http://forums.yogstation.net/index.php?forms/unban-appeal-form.1/resp
 
 ## Remove the # to show a popup 'reply to' window to every non-admin that recieves an adminPM.
 ## The intention is to make adminPMs more visible. (although I fnd popups annoying so this defaults to off)
-#POPUP_ADMIN_PM
+POPUP_ADMIN_PM
 
 ## Remove the # to allow special 'Easter-egg' events on special holidays such as seasonal holidays and stuff like 'Talk Like a Pirate Day' :3 YAARRR
 ALLOW_HOLIDAYS


Zostroll:
Admin pop up has killed me like 5 times. They boink is enough.

Super3222:
Don't. If someone's ignoring admin messages and claiming "ME NO SEE IT", than that's on that person. Not the admin.

In any case this should stay disabled. I'm 100% sure someone else tried enabling this before, and it got turned down.

At the VERY LEAST to integrate this into Tickets you could add a BUTTON for ADMINS to "force popup" on someone.

In all cases, I'd be common sense to be careful with it. So using it, and getting someone fucked over like in zostrolls case would be the admins fault. However, in almost every scenario a player should be able to hear that unique BWOINK sound and KNOW that an admin has PM'ed them. So really, it's going to be the players fault without the force popup.

So anyway :-1:

ghost:
People keep whining when I ban them because 'ME NO SEE MESSAGES WHAAAAAAA'

If you fuck up and get a message, that's on you. If the admin tickets you mid combat so you die, that's on them.

Nickvr628:
The issue becomes when someone is in the heat of the moment and does not hear the sound and see the chat message, or they walk away for a moment and when they sit back down the message is already buried in the logs.

I like the idea of a force popup though.

utahclock:
what nick said and the bwoink doesnt always fire if other sounds are.
On Oct 11, 2016 10:06 AM, "Nickvr628" notifications@github.com wrote:

The issue becomes when someone is in the heat of the moment and does not
hear the sound and see the chat message, or they walk away for a moment and
when they sit back down the message is already buried in the logs.

I like the idea of a force popup though.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
https://github.com/yogstation13/yogstation/pull/977#issuecomment-252944560,
or mute the thread

Zostroll:
Adam a lot of tickets are made for no rule breaking reasons. If you are saying a miner deserves to die to a Goliath cause he has asked a question then you are a bit of a cunt.

utahclock:
if someone asks a question admins can just a ghost to em before replying if
its a miner

Super3222:
Like I said earlier. You shouldn't force anything on the players or force the admins to popup players.

Simply make it a toggable option, where in a ticket you can click a button that makes every comment that YOU (the admin) send pop up on the player / ticketee.

Therefore, it's the admins decision whether they want to pop up or not. And ticket-based situations become more easy and flexible to deal with. There'd literally be no reason not have this system over pop ups unless you believe that every comment within a ticket is 100% important and NEEDS to be sent to a player immediately. Which is untrue.

BOBsHERE:
I liked when the admin chat popped op. Make it pop up to the side, so it doesnt cover your character or his immediate move space. Also make sure that it doesnt auto select, so we can still move around until we actually click on it.

Ephemeralis:
This is a configuration issue and lies within the scope of the Council, not the coders.

Closing.

PR #988, opened 2016-10-12
Adds in attack verbs to 4 items

Karinrine:
The station charter, cyborg endoskeleton, all ID cards and the grey
jumpsuit get unique attack verbs. The endoskeleton also plays the sound
when you borg someone ("I live.. again") when you hit someone or
yourself with it(this doesn't mean you go on a spree of hitting people
with the endoskeleton; it injures, though it would be a cheesy death if
you get beaten to death by an endoskeleton) as well. The attack verbs
for the ID cards varies on its type(example; hitting people with an
abductor ID may say alienated, abducted, etc.). The grey jumpsuit attack
verb is self-explanatory.


code/game/objects/items/weapons/cards_ids.dm
```diff
@@ -49,6 +49,7 @@
 /obj/item/weapon/card/emag
 	desc = "It's a card with a magnetic strip attached to some circuitry."
 	name = "cryptographic sequencer"
+	attack_verb = list("emagged", "hacked", "glitched")
```
ShadowDeath6: have you tested this to see if it does anything when used on borgs?

Karinrine: I haven't; it could be changed so that it'll only apply to humans and other races. I don't really know DM, so someone should change it to exclude emaggable objects (preferably someone with write access as they can edit the code).
I could also always remove the code for the emag's attack verbs, however.

Karinrine:
I could, but I'm mostly wondering if anyone would help me in making the attack verbs for the emag not apply for cyborgs.

PR #1004, opened 2016-10-14
Prevents 2 borers being spawned during the Strange Reaction

### LINKED ISSUE #999: Secret borer technique creates two borers, one braindead.
Title.

X-TheDark:
fixes #999.

As per title.
Also prevents mob-spawned-in-nullspace related runtimes during the chemical spawn thing by giving it a turf parameter;
Fixes some borer-related runtimes when the borer is spawned without a client;


ShadowDeath6:
thank

PR #1014, opened 2016-10-14
Fixed krav maga not respecting NODROP

Cruix:
You could disarm a changeling with krav maga and get your own armblade.
And the changeling would also have an armblade.
And you would have an armblade duel and get cremated even if you won because you can't remove the armblade.
Soon we'll have armblade sting for this, don't need it anymore. Totally a feature removal and not a fix.

:cl:
fix: The masters of Krav Maga are baffled by the fact that they can no longer use their martial art prowess to replicate a changeling's armblade.
/:cl:


ghost:
WHAT THE FUCK CRUIX THIS IS THE MOST HILARIOUS BUG I'VE EVER HEARD

Ephemeralis:
This is so hilarious that it should almost be functionality.

Can we do that instead?

PR #1015, opened 2016-10-14
Having one arm no longer makes you unable to be cuffed

ghost:
it was cute at first but then people started powergaming to cut off their own arms so they could never be cuffed and it's really irritating watching 'powergame-allowed' things like antags cut off their arms with eswords and run around murdering security because lol i cant be cuffed and i have no real downsides aside from being unable to use a dual esword
#### Changelog

:cl:
rscadd: Security officers have taken a seminar in advanced handcuffing, and are now able to handcuff one-armed suspects by wrapping the other end of the handcuff around the suspect's neck.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_helpers.dm
+++ b/code/modules/mob/living/carbon/human/human_helpers.dm
@@ -6,8 +6,6 @@
 /mob/living/carbon/human/canBeHandcuffed()
 	if(get_num_arms() >= 2)
 		return TRUE
-	else
-		return FALSE
 
 //gets assignment from ID or ID inside PDA or PDA itself
 //Useful when player do something with computers


Super3222:
advanced handcuffing

Cruix:
All procs automatically return null (which evaluates to false) by default, unless they change the . var, so technically this does nothing.

Cruix:
Did not mean to close it, it's still good for readability

Cruix:
Oh wait shit I misread everything this is doing the opposite of that

If you want this to work the proc has to always return TRUE

ghost:
it does return true

Cruix:
If they have less than two arms it will return null, which is false. To do what you want it to the proc would just have to return true with no if().

X-TheDark:
As this is right now, it literally doesn't change anything, as if this returns anything not true, then you can't cuff the guy. So returning null is as good as returning FALSE.

Super3222:
@AdamElTablawy are you still working on this?

PR #1028, opened 2016-10-15
Removes carp spawns near and around permabrig

Tenryumi:
yogstation 13 99 lagfree custom map west object chi 2016-10-11 164011
# THIS NEEDS TO **STOP**.

Why are there _SO MANY_ space carp spawns directly next to the area where only unarmed, defenseless prisoners reside? I've counted **7 CARP SPAWNS.**
dreammaker_2016-10-15_13-57-30
dreammaker_2016-10-15_13-57-51
dreammaker_2016-10-15_13-57-40

Removes all carp spawns near permabrig, where carps could easily happen to pass by windows and spot a tasty defenseless inmate.
# "Why tho? Prisoners can escape with this which is good and gets them involved with the round again."

It also renders the permabrig inhospitable and useless. The bloodthirsty space carps attacking inmates doesn't help, either. This changes the permabrig's use to contradict its own original intention. It's not meant to be a means of escape. It's a _prison,_ and should be a proper one. 
Unless a well-behaved inmate is content with being forced to sit in a 3x3 cell because a carp broke a single window, that is.

In addition, that method of escape requires _no effort_ on the player's part and looking at it objectively makes me realize just how lame that is.
_So, how'd you break out?
Who, me? Well, I used my wits to trick an officer to coming in with his baton out, where I then slipped him with soap, stunned him with his own baton, stripped him and beat him to death after stuffing a makeshift muzzle in his mouth, and then walked out of the Brig disguised as him. How about you?_
_Oh, I just sat on my ass and waited for a carp to open the window for me._
_..._

:cl:
rscadd: NT has covered the exterior of the permabrig with space carp repellent, after more than 10 instances of space carps breaking into and eating defenseless inmates.
/:cl:


Super3222:
You probably need to delete your master branch, make a new one, and than branch off of that without making any commits.

Tenryumi:
Given the myriad of commits that appear on all of my PRs now, I have to agree.
Wouldn't that delete all my branches and PRs, though?

Tenryumi:
Oh screw it. I'm gonna do that and backup my changes. Be back soon!

PR #1041, opened 2016-10-18
Floors in space between escape airlocks, adds reference to chapel mass driver

Tenryumi:
I've never understood why the space between the tunnels in escape was so wide and open. This 5x3 area of wide-open space creates a number of problems:
- It makes the way into the shuttle very claustraphobic. It's difficult to bring machinery onto the shuttle because of the narrow tunnels.
- It's unbelievably easy to block off escape entirely as a traitor AI or borg. Bolt + shock all the doors before the shuttle arrives, and you're gauranteed to force everybody but those with spaceproof equipment to hurt themselves greatly AND force a shuttle breach should they want to escape. That's not even taking into account all the people that will be shocked by the doors in the process.
- It just... looks tacky and impractical. Impractical because of the above, tacky because of the lattices and empty space.

dreammaker_2016-10-17_19-54-09
So, I've covered up this tacky space with something that simply looks more natural and offers those without EVA access a way onto the shuttle without nearly killing themselves (they'll probably still get shocked before they realize the route they'll need to take though!)

_ohmigosh, I found the proper way to merge this now! Now you can read the changes properly!_

**Let's be honest though, the change to the chaplain's mass driver I've made is easily the most important part of this PR.**


X-TheDark:
A reference is sorely decreased in it's value as a reference if you straight out tell everyone what it is a reference to. Subtlety is key with these things.

Tenryumi:
DAMN IT YOU'RE RIGHT!

X-TheDark:
Nice message telling that you changed the title.

Alek2ander:
> It's unbelievably easy to block off escape entirely as a traitor AI or borg. Bolt + shock all the doors before the shuttle arrives, and you're gauranteed to force everybody but those with spaceproof equipment to hurt themselves greatly AND force a shuttle breach should they want to escape.

That's kinda the point. Hijack is a thing.

Tenryumi:
@Alek2ander Yes, and hijack is deceptively easy because of it. 
- Order a borg to hold still (law 2 after all,) then emag it
- Bring borg to airlock tunnels seconds before the shuttle arives
- On the borg's side: Ctrl + click, then alt + click 8 doors
- emag the shuttle console

Hijack need to be made more challenging, and it needs to be easier to get onto the shuttle in the event someone DOES do the above (which happens all the time, by the way.)

ghost:
no, stupid change

it's intended to be like this because lack of efficiency etc and so taht hijack, the most difficult objective by far, can't be accomplished with hilarious ease

you can't order a borg to hold still and emag it, it'll bolt before you can, same with if it sees a flash, not to mention AI will scream it's head off and the crew will fireaxe in, etc

so no

it is intentionally difficult to get onto the shuttle if someone hijacks it, no fucking hugbox

booo 

bad

against

fully against

i'll write a more coherent response later but in general-this is stupid hugbox and also conflicts with the whole theme of the station.

ghost:
oh and also it's fun to make space diners there not to mention it makes no sense to have windows right next to shuttle windows

Tenryumi:
@AdamElTablawy You still have to smash through 2 sets of refined windows; that's more than enough time given to tatorman to emag the shuttle console or to shoot at everybody with lasers (through the windows!)

> "you can't order a borg to hold still and emag it, it'll bolt before you can, same with if it sees a flash not to mention AI will scream it's head off and the crew will fireaxe in, etc"

You're giving the players in this scenario far too much credit here. It's very easy to get a borg to hold still. A simple "hey, borg" will give you plenty of time so you can take out the flash (that you've been hiding in your pocket to immediately take out and use just for this purpose) so you can subvert it in less than 3 seconds in a way that is so subtle the AI will barely even be able to tell what just happened, and that's assuming that the AI is even looking in the first place (it has doors and obnoxious human commands to attend to after all.)

> "it is intentionally difficult to get onto the shuttle if someone hijacks it, no fucking hugbox"
> "it makes no sense to have windows right next to shuttle windows"

To both of these comments, and to address @Alek2ander 's comment as well...
image
Metastation does the same thing, only _those_ windows lead into the bridge. And this map is meant for _**highpop**_ servers. We'll also be using this map in the near future as well. Just saying.

> "oh and also it's fun to make space diners there"

You can still make the space diners though. The only difference is that they'd be 8 tiles down.
image

X-TheDark:
Excellent...this is why I hate map changes...they always conflict with each other, whenever you merge one...

Tenryumi:
I'll see to it that this gets fixed.

Tenryumi:
Since this PR's branch was deleted during my purge of messy master history, I've re-uploaded and tweaked this PR in #1147 .

PR #1042, opened 2016-10-18
Module Framework for /obj/items

X-TheDark:
fixes #headcoder_being_a_lazy_git
See https://github.com/X-TheDark/yogstation/pull/3 for a more complete history of commits/things I've done.

I'll keep this short and sweet, hopefully and try to keep the implementation details away, unless someone starts asking for those;

...this is basically what happens when a developer tries to implement something (pickpocket gloves) in an efficient way - an entirely new system is created without actually making the said something (yes, I actually did this entire system out of trying to implement pickpocket gloves in a non-ass way).

**Post questions/suggestions if you have any. This is currently not really used on much, as I don't have any ideas for what kind of modules would be balanced/work well**
### Summary and current modules

Consists of 2 items - module holder and the module itself;

Module holder can be placed into any obj/item (except storages), although this is more for testing purposes, and afterwards, modules can be inserted into the equipment.
#### What is the purpose of this?

This allows a object-independant use of things like reaction to being hit and whatnot. For example, with this system, you can define a "Reactive Armor Module" that teleports person when he's hit. This module could be made so it can be placed into ANY suit/uniform, so now ANY suit/uniform that accepts modules can act like Reactive Teleportation Armor, making it not necessary to have subtypes of armor with different hit reactions; And this is just scratching the surface...any on-attack application things can be defined as any on-being-attacked application things can, in addition, it's very granular in that it separates a lot of types of attacks (as does the regular attack code, so it's very consistent);
#### Where it hooks into code

First of all, **this is a separate system, not integrated with shields/anything else, it's entirely it's own thing so current shield checks/etc remain unchanged**.

When you click on stuff, depending on whether or not you have something in your hand and if it's adjacent, you get different assault resolution options:
- If adjacent+unarmed -> calls module's `on_unarmed_attack`
- If non-adjacent+unarmed ->calls module's `on_ranged_attack`
- If adj+armed -> calls module's `on_obj_melee_attack`
- If non-adj+armed -> calls module's `on_obj_ranged_attack`

When something hits/attacks you, the resolution is as follows (the details are in the Details section):
- Unarmed attack -> `on_attack_hand`
- Armed attack -> `on_attacked_by`
- Hit by projectile -> `on_bullet_act`
- Hit by thrown stuff -> `on_hitby`

**Current admin-spawn only modules**
- Shock Glove module - idea stolen from goon. Requires a nearby exposed power cable with power in it. Has a cooldown of 3 seconds between uses. Max range of 6, with a 1 second delay between clicking and actually shocking someone (unless melee range, in which case the shock is instant). You have to stand still during this one second and the target must remain in 6 tile range. Damage depends on power in the grid, much like electrified doors; Requires clicking with an empty hand;
- Gravity Manipulation module - basically portable sorium/that_other_thing. Click on a tile/mob and things get interesting, depending on mode selected: **Push/Pull** : pushes or pulls the mob, doesn't work on turfs, obviously. **Attract/Repel** : acts like the chems, makes a vortex that either pulls or pushes everything in 3 tile radius towards/away from the clicked mob/turf. Unlike chem counterparts, this always throws. The user is immune to it's effects, so go nuts. 3 second CD when used in Push/Pull mode/6 second CD when used in Attract/Repel mode.
#### How are these modules "used" (assault module specifics)?

The modules are used by checking them in a certain sequence, for defining which I've made a helper datum - `/datum/resolve_order`; This is resolved in specific sequence, where one objects modules are applied before the others;

Assault modules (anything that's defined as something that applies when you click on something) resolve as follows (and in the order specified by First_thing_to_resolve->Second_thing_to_resolve->...):
- If you **have** an object in hand, currently, **only this object's modules are used**;
- If you **don't have** an object in hand : the modules are applied by equipment you wear in the following sequence : **Helmet -> (Hard)Suit -> Uniform -> Gloves**
#### What about defense modules?

Defense modules have these rules (Thanks @Ephemeralis for helping me with ideas on how to make a more sensible system for this):
**Your body is divided into 3 zones for the purposes of resolution : Head / Torso+Arms / Legs**
1. First of all, if you have an item held in hand (either/both), **modules in your in-hand items are resolved first**
2. If Head is hit - only resolve the Head module (Hat/Helmet, but not Mask).
3. If Torso+Arms (any of these) - Suit->Uniform
4. If Legs - Suit->Uniform->Boots

Furthermore, there is support for 2 types of defense modules : global and local;
- Global defense modules - applied first, regardless of the zone you are hit in (as in, a module inside a helmet will work if your legs are attacked, if said module is defined as global); This might be, for example, a personal shield module kind of thing;
- Local defense modules - applied only for the zones that are hit - if your head is hit, modules from the suit/uniform are NOT applied, etc;
#### How do I interact with these modules (turn them on/see description/etc)?

If a holder is inserted (or created inside of) an item, that item receives a new **verb** - Modify Modules([name_of_equipment]) and looks something like this:
modules
It may not be the prettiest thing, but it's functional **and also not TGUI** (seeing as TGUI is not liked due to it's slowness).
#### What about stuff that happens after the modules (like hugging someone), do those actions still apply or do the modules override everything else?

Good question. This depends on what the hook procs return (the ones defined for modules in the "Where it hooks into code" section) : if the module procs **return TRUE** - all further actions are skipped (you don't hug anyone, you don't grab anyone, you don't attack anyone with the item in-hand, etc). If the module **returns FALSE**, we continue the resolution (as in, attempt to hug/grab/attack someone with thing in hand).
#### How do I get these?

Admin-spawn only right now, until more balanced modules are implemented;
#### Is there anything this CAN'T do?

Currently planned features that will be done separately:
- Weapon modules - currently there aren't any modules that use weapons (as in, "fire laser module", etc). The current system has defines for power costs and also holder has a power cell (currently unused), but I still need to hook up the power usage (which is why all current modules (all two of them) are cooldown based);
- Ability modules - anything that gives you an ability/action button that you activate whenever you want, rather than needing to click someone/something (spells, basically).
- Passive modules - there aren't any proper systems in place for passive modules (that use power each tick and enhance user/item in some way);
#### What do these look like?

clever_very_clever_indeed
Very clever, indeed. /s
#### Other questions

_This will be added as things are discussed in this PR_


X-TheDark:
Interesting Travis failure reason (the code is fine, seems like something on the host caused an error. I've compiled/rewritten parts of this so much that I'd kill myself if the actual code failed to compile, that's how much work and trust I've put into this).

code/modules/modular/module.dm
```diff
@@ -0,0 +1,156 @@
...
+	var/id = "module" //this should be thought of as a slot rather than id, you can only place 1 module with the same id into the holder
+	var/can_be_applied = TRUE //Does this module apply anything to anything or is it something passive that isn't checked
+	var/can_be_toggled = TRUE
+	var/can_be_removed = TRUE
+	var/has_charges = FALSE
+	var/module_type //bitflag for determining if this module should be resolved on attack or on defense or if it gives us any kind of ability to use, see code/__DEFINES/modules.dm
+	var/onhit_type
+	var/list/action_types //list of action types we are to be given
+	var/list/action_list
+	var/unique_resolution = TRUE //Only 1 module of this id can be applied on one action, processed here and in mob_procs.dm
```
X-TheDark, line 15: 13-15 line related things still need to be done, these are here just so I remember to actually work on this.

code/modules/modular/modules/gravmanip.dm
```diff
@@ -0,0 +1,63 @@
...
+	var/vortex_range = 3
+	mode_list = list("Push", "Pull", "Repel", "Attract")
+	var/const/USE_COOLDOWN_MOB = 30
+	var/const/USE_COOLDOWN_TURF = 60
+
+/obj/item/module/assault/gravmanip/can_be_applied(atom/A, mob/user)
+	if(!..())
+		return
+	if(mode in 1 to 2)
+		if(isturf(A)) //can't push pull turfs (can still do so with objects, shhhh)
```
X-TheDark: ...*sigh*, I'm so bad at remembering this. You actually can't, because `/obj/item` is not an `applicable_atom_types`...way to go, me!

oisin100:
[wow](https://www.youtube.com/watch?v=jUy9_0M3bVk)

ghost:
why though

X-TheDark:
@AdamElTablawy 
No reason other than to make it more interesting for people to try to code stuff, really. Unfortunately, as I said, currently, the only 2 uses are the glove modules - Power Glove (aka, ranged shocking thing) and Gravity Thing. It's a framework on which to expand upon at some point in the future...or something.

ShadowDeath6:
Please make powergloves its own traitor item I would love to see someone wire the singlo into the grid and fry people

Ephemeralis:
There's refactors, then there's _refactors_.

Daaaamn. Can do some cool shit with this.

X-TheDark:
@ShadowDeath6 I firmly believe that shock gloves are too good right now.

The electrocution_act **ALWAYS** stuns for 5 seconds/units/BYONDtime, even if there's nominal (more than 10000) power in there. If Adam's changes to shock damage are merged, this has the potential to do 150 (75 to someone with "regular" gloves) damage once every 3+1 seconds (3 sec CD + 1 sec charge, unless melee). So, basically, what I'm trying to say, shock gloves would be pretty damn good, considering it's ggnore, unless you have something that shortens stun times/insulated gloves.

X-TheDark:
I just need to do one small thing to simplify list processing, then this will be ready for merging.

X-TheDark:
I think this is as merge ready as I can make it on the first pass.

PR #1047, opened 2016-10-20
Buffs last Resort

ghost:
last resort is kind of a shitty power that's inferior to, well, lesser form or anything. you gib and become a shitty wimpy animal that can't revive or use any powers and has low health and can't really escape before getting your ass kicked and it depends on having a corpse and a long ass incubation time to come back again.

you can't even use it when dead or unconcious, where literally any other ability would be better.

now, Last Resort can be used when unconcious and the stun is long enough to escape into the vents, or, if you think you can, chomp the guy stunned to death and egg him-but since he's not unconcious likelyhood is he'll scream so it can backfire without proper prep.
#### Changelog

:cl:
rscadd: The Last Resort ability can now be used when dead or unconcious and is slightly more viable overall.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/headcrab.dm
+++ b/code/game/gamemodes/changeling/powers/headcrab.dm
@@ -22,7 +22,7 @@
 		H.blur_eyes(20)
 		H.adjust_eye_damage(5)
 		H.confused += 3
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
How to make it op- allow regen as the sluggo

PR #1048, opened 2016-10-20
puremanbird cannot into robust a changeling anymore-AKA, anatomic pancea buff

ghost:
Anatomic pancea was sort of underwhelming as an ability-Yeah, it treated some things, but it didn't really...treat toxin damage per se, nor some things it should have treated. Hopefully, this fixes part of that.

Specifically, it now treats deafness and blindness/eyedamage, won't treat harmless diseases, and will automatically set the user's toxin damage to 40 if their toxin damage is above 40-AKA, if they have 60 toxin, they instantly go to 40. But it won't say, set 20 tox to 40 tox. It's up to the pentetic acid to cure the toxin.

I'm not sure if the last bit, the toxin treatment, actually works-Might need to come up with a better solution for that.

It also fixes hallucinations. 

The cost and chem cost is higher tho as a result.
#### Changelog

:cl:
rscadd: Anatomic panacea is more efficient at it's intended job.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/panacea.dm
+++ b/code/game/gamemodes/changeling/powers/panacea.dm
@@ -18,23 +18,21 @@
 			C.vomit(0)
 		egg.loc = get_turf(user)
 
-	user.reagents.add_reagent("mutadone", 10)
-	user.reagents.add_reagent("pen_acid", 20)
-	user.reagents.add_reagent("antihol", 10)
-	user.reagents.add_reagent("mannitol", 25)
-	user.hallucination = max(0, M.hallucination - 50)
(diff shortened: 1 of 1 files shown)


Tenryumi:
Travis is displeased.

ShadowDeath6:
This is godawful. 

Remove the pentetic acid, give it basic antitoxin, and have it remove all chemicals before adding the curing medicines.

ghost:
if only i knew how to remove all chemicals, i would @ShadowDeath6

ShadowDeath6:
`user.reagents.clear_reagents()`

PR #1059, opened 2016-10-20
ports lzimann's eshuttle change

ghost:
get hit by the shuttle and gib motherfucker
#### Changelog

:cl:
rscadd: CHOO CHOO
/:cl:


X-TheDark:
Nuh-uh, not until I/someone_else partially ports the shuttle startup thing. As this would currently just instagib with no warning/chance of escape.

As fun as this change is (no, really, I'm all for this), we do need a shuttle leave/arrival cooldown before this will not cause mining shuttle to be death incarnate.

ShadowDeath6:
can we make anyone who plays miner instantly gib

Issue #1039, opened 2016-10-17
Lings can remove mindslave at will with horror form


puremanbird:
Is this a joke?

ShadowDeath6:
This isn't intended but lings are notoriously resistant to mind control

puremanbird:
Then make the lings resist the implant without destroying it or something. Right now they get the mind slave message like anyone else would, so a ling that removes their implant is like an ai requesting antimov. The mindslave costs too much TC to have it wasted.

vladlds:
>lings are notoriously resistant to mind control

@ShadowDeath6 can you link me where is that coming from?

ShadowDeath6:
I'm working on having them just push the implant off or something without consuming it

I also don't have a source, but it's the same reason that loyalty implants don't/didn't work on non-conversion based antags in the past. Mindslave shouldn't work on traitors for the same reason, but that's more of a balance pass than anything else considering that it's much harder for them to just bug it out of their body.

PR #1070, opened 2016-10-22
you can no longer mindslave implant changelings

ShadowDeath6:
Refer to issue #1039 .

inb4 remove don't improve

lings have too many ways to alter their form and adding checks for every old/new thing is kind of irritating considering that most antags are resistant to mind control and lings (as far as I've seen in code) are the most resistant


ghost:
yeah seems fine to me

PR #1076, opened 2016-10-22
ports goofball foamchange except makes it more unfair

ghost:
ports the goofball foamchange except makes it even more OP
#### Changelog

:cl:
rscadd: FOAM IS NOT ASS. Credit to goofball.
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/effects_foam.dm
+++ b/code/game/objects/effects/effect_system/effects_foam.dm
@@ -13,6 +13,7 @@
 	animate_movement = 0
 	var/metal = 0
 	var/lifetime = 40
+	var/reagent_divisor = 5
 
 
 /obj/effect/particle_effect/foam/metal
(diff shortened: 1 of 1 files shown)


AsV9:
So what does this do then?
Please explain more.

code/game/objects/effects/effect_system/effects_foam.dm
```diff
@@ -147,7 +155,7 @@
 		location = get_turf(loca)
 
 	amount = round(sqrt(amt / 2), 1)
-	carry.copy_to(chemholder, 4*carry.total_volume) //The foam holds 4 times the total reagents volume for balance purposes.
+	carry.copy_to(chemholder, 6*carry.total_volume) //The foam holds 4 times the total reagents volume for balance purposes.
```
Super3222, line 158: foam now creates 6 times the total reagents..

Super3222:
This seems like an odd change. What's the reason behind it?

ghost:
Goofball changed it so that foam made 5 times total reagents, i made it 6 instead

Reason-clothes absorb the majority of foam, even basic clothes, so gas is 4000x superior. This brings foam on par with gas because foam will actually apply it's reagents.

The issue with foam is that it divides it's reagents over the tiles it covers and then goes throgh clothes so a 50u acid foam will end uo just being 3u acid. You need to hse tiny amounts of foam to get an effect so smoke was superior

Now 50u acid will apply properly to people in it

BOBsHERE:
Awesome. Basically the reason I never bothered with foam grenades.

Super3222:
Went ahead and tested it using a grenade (on a test server).

Beakers:
1) 50u of that foam reagent stuff, and 50 CLF3
2) 50u of the foam reagent stuff, and 50 more CLF3

result;


husk was due to space damage.
the mob kept on lighting on fire - wasn't immediately killed or anything like that. it took time.
the mob was harmed by the burn damage extremely slowly, i was unable to tell whether they had reagents inside of them or not.
reagents moved extremely fast.

i'd say it's fairly powerful. I don't think we'll have any problems.

Cruix:
Let's give it a try.

PR #1077, opened 2016-10-23
Starts the first of many HoG changes

ghost:
Traps now can be instantly destroyed by hitting them with any object, much like rat'var sigils.

Traps take 20 seconds to recharge.

Null rods protect the user from traps.

Gives specific trap descs for cultists and noncultists.

Todo-Give traps placedby vars to track the deity that placed them, and thus track their affiliation with different sides and thus ignore cultists of the same side.
#### Changelog

:cl:
rscadd: HoG traps have been changed significantly. They can be instantly destroyed by hitting them.
/:cl:

### DIFF
--- a/code/game/gamemodes/handofgod/traps.dm
+++ b/code/game/gamemodes/handofgod/traps.dm
@@ -45,7 +45,7 @@
 		desc = initial(desc)
 	if((side == "red" && is_handofgod_redcultist(H)) || (side == "blue" && is_handofgod_bluecultist(H)))
 		desc = friendlycultdesc
-	if((side == "red" !is_handofgod_redcultist(H)) || (side == "blue" !is_handofgod_bluecultist(H)))
+	else
 		desc = enemycultdesc
 	alpha = 200
 	animate(src, alpha = 30, time = time_between_triggers)
--- a/code/game/gamemodes/handofgod/traps.dm
+++ b/code/game/gamemodes/handofgod/traps.dm
@@ -6,10 +6,14 @@
 	alpha = 30 //initially quite hidden when not "recharging"
 	health = 1
 	maxhealth = 1
+	desc = "A bug. Report this to a coder."
+	var/friendlycultdesc = "A bug. Report this to your god, and also a coder."
+	var/enemycultdesc = "An enemy bug. This probably shouldn't exist, ahelp it."
 	trap = TRUE
 	autocolours = FALSE
 	var/last_trigger = 0
 	var/time_between_triggers = 200 //20 seconds to recharge
+	var/side = "neutral" //"blue" or "red", also used for colouring structures when construction is started by a deity
 
 
 /obj/structure/divine/trap/Crossed(atom/movable/AM)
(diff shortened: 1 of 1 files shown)


Super3222:
> Todo - Rework the entire mode and keep up to date with it until the end of eternity

ShadowDeath6:
> traps

ghost:
imrpoved

ghost:
CHOKE ON A COCK TRAVIS

YEAH

NOBODY LIEKS YOU

Cruix:
Replaced an unsuccessful attempt at Demorgan's law with an else clause

code/game/gamemodes/handofgod/traps.dm
```diff
@@ -132,12 +132,12 @@
 	time_between_triggers = 1200 //Exists for 2 minutes
 	
 /obj/structure/divine/trap/ward/attack_hand(mob/user)
-	if(iscarbon(user) && side == "red" && is_handofgod_redcultist(H)) || (side == "blue" && is_handofgod_bluecultist(H))) && user.a_intent == "harm")))
+	if(iscarbon(user) && ( (side == "red" && is_handofgod_redcultist(H)) || (side == "blue" && is_handofgod_bluecultist(H)) ) && user.a_intent == "harm")
 		if(density == 1)
-			user.visible_message("You force the ward to become passable.</span>")
+			user.visible_message("<span class='notice'>You force the ward to become passable.</span>")
```
ShadowDeath6, line 137: why is this a visible message

code/game/gamemodes/handofgod/traps.dm
```diff
@@ -1,16 +1,19 @@
...
+	name = "A bug"
 	icon_state = "trap"
 	density = 0
 	alpha = 30 //initially quite hidden when not "recharging"
-	health = 20
-	maxhealth = 20
+	health = 1
+	maxhealth = 1
+	desc = "A bug. Report this to a coder."
+	var/friendlycultdesc = "A bug. Report this to your god, and also a coder."
```
Super3222, line 10: Alright, let me teach you something about vars.
Compare `health` to `var/friendlycultdesc`.
The difference is that health was already predefined in a parent type object (don't think too hard on that one, just remember there was once a basic object that had `var/health` and now that you've grown a divine version it doesn't need the `var/`.

Now the problem is, every child of this item (`/obj/structure/divine/trap`) has a `var/` as well.

So really, for every child of this type (`/obj/structure/divine/trap`), you only need `friendlycultdesc = blah blah blah`.

Hope that clears it up

ShadowDeath6: muh copypasto

Super3222:
downed version:

anything that's not `obj/structure/divine/trap`, but is a variant like `/obj/structure/divine/trap/fire`.

do not use `var/friendlycultdesc`.
use `friendlycultdesc` instead.

Super3222:
@adameltablawy correct thy mistakes

Bittern12:
sounds like stuff that needed to be added. Good job Adam

PR #1083, opened 2016-10-23
Fixes mobs not being able to be thrown

### LINKED ISSUE #1081: Old grab system was better
New grab is baffling, it doesn't seem to restrain anymore which reverts people who have legitimate reason to restrain people from preventing harm into critting to stop people.

 The hand icons about grabbing and killing and flashing when you are killing in a hold are gone too making new grab system confusing and easy to strangle someone.

And the one unforgiveable thing that is gone because of new grab is throwing people.

X-TheDark:
Error caused due to some leftover code from the old grab system (that was an item which was checked for), presumably from when conflicts were resolved during the rebase.

Tested stuff:
- Thrown a mechanical toolbox, works as expected;
- Thrown the fireball spell thing, works as expected;
- Thrown an aggressive grabbed human, now it works as expected;

:cl: Mob Throwing Enthusiast
rscadd: Mobs can now be thrown again, enjoy.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -236,13 +236,9 @@
 					var/start_T_descriptor = "<font color='#6b5d00'>tile at [start_T.x], [start_T.y], [start_T.z] in area [get_area(start_T)]</font>"
 					var/end_T_descriptor = "<font color='#6b4400'>tile at [end_T.x], [end_T.y], [end_T.z] in area [get_area(end_T)]</font>"
 					add_logs(src, throwable_mob, "thrown", addition="from [start_T_descriptor] with the target [end_T_descriptor]")
-
-	if(I && I.prethrow_at(target))
-		return
-
-	if(!(I && istype(I)))
-		return //Grab processing has a chance of returning null
 	else if(!(I.flags & (NODROP|ABSTRACT)))
+		if(I.prethrow_at(target))
+			return
 		thrown_thing = I
 		unEquip(I)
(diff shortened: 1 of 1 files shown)


Paragonid:
Do kill macrosses work as expected?

X-TheDark:
There's a delay between grabbing, upgrading that grab and then throwing.
You ALWAYS stay stunned for shorter than it takes to regrab and throw you, as:
- There's a delay after grabbing someone (1 second), before you can upgrade that grab;
- There's a delay after upgrading the grab, before you can throw (1 second again);
- You are `Weaken`'ed for 1 tick (half a second) if you are thrown against a wall;

So, you'd need to just let the person grab/upgrade/throw you for you to be killed by it. I've tested this for 5 minutes locally, and I've not been able to get an aggressive grab before they get up.

Unless I'm misunderstanding as to what a "kill macro" is in this scenario.

Ephemeralis:
Kill macros are retarded anyway and there should always be counterplay to death prep scenarios that don't involve "don't get grabbed".

Seems good to me.

X-TheDark:
No, seriously, what is this "kill macro" I keep hearing?
1. There's at least a 2 second delay between grabbing and being able to throw (since you need to upgrade grab, then throw);
2. Throw stuns you for 1 tick;
3. When you are thrown, the grab is dropped, so you need to be regrabbed (see point 1);
4. Passive grabs can be walked out of;

I guess the only problem are the aggressive grabs than have a 30% chance of being resisted out of?

PR #1085, opened 2016-10-23
buffs xeno hives further

ghost:
simplemob xenos were getting wiped too easily so here you go they're much worse
#### Changelog

:cl:
rscadd: Simplemob xenos are far deadlier, again
/:cl:


Time-Green:
I am fine with this, on one condition:
- Make them gib if you fire extinguish them while they just came from lava

ShadowDeath6:
tbh I think the robust pathing is too op for these, the projectiles are practically homing and you get your dick kicked if there's more than one spamming spit

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -12,6 +12,7 @@
 	response_disarm = "shoves"
 	response_harm = "slaps"
 	speed = 0
+	weather_immunities = list("lava","ash")
```
Super3222, line 15: I agree with ash. But lava? I feel like they should be affected by whenever a clever miner leads a hive into it.

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -113,7 +115,26 @@
...
+		alien_state = 0
+
+/mob/living/simple_animal/hostile/alien/sentinel/Life()
+	. = ..()
+	if(.)
+		if(alien_state == ALIEN_HAUL_ASS)
+			retreat_distance = 7
+			minimum_distance = 7
+			rapid = 1
+			projectiletype = /obj/item/projectile/neurotox/panic
```
Cruix, line 134: Buffing the projectile and making them shoot three at a time is a bit much. You also don't set these back to their initial in the else.

code/modules/mob/living/simple_animal/hostile/alien.dm
Cruix: It does not make much sense lore wise, but balance is more important.

PR #1086, opened 2016-10-23
Chasms no longer swallow dim light tiles.

Super3222:
:cl:
fix: Chasms no longer swallow dim light tiles.
/:cl:


drderp3635:
a hole so deep it swallows the absense of light. feature

Super3222:
Note for Coders: Dim lights are on the TURF layer, not the LIGHT layer, which is why it's interacting with the Entered() proc in the first place.

ShadowDeath6:
Does this fix the chasm slowly opening

Super3222:
I don't believe that this'll fix that, however it might if deleting the light tiles was slowing down the chasm.

PR #1093, opened 2016-10-24
changes crate health from a ridiculous 1000 to 150

ghost:
so you can shoot open crates

'b-b-but muh guncargo'

you can already open locked crates with lava which anyone in cargo can do now you can just shoot em too
#### Changelog

:cl:
rscadd: Crates can be shot open in a reasonable timeframe now.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates/secure.dm
+++ b/code/game/objects/structures/crates_lockers/crates/secure.dm
@@ -4,7 +4,7 @@
 	icon_state = "securecrate"
 	secure = 1
 	locked = 1
-	health = 1000
+	health = 150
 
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
Cargo is the cancer of ss13

Paragonid:
how about lockers
Hiding in the locker (even without lock) is terribly shitty in-combat tactics. They should take damage inside maybe, or at least locker should be destroyable.

ghost:
Lockers are destroyable

Super3222:
I'm fine with this. It looks good, and it wasn't a big problem the last time I changed it on the old repo.

PR #1116, opened 2016-10-29
Objectively makes IV Drips UNFUN (fixes the wild blood/reagent reactions)

### LINKED ISSUE #1103: Blood bags/IV drips are bugged like hell
When you attach an IV drip to someone with the correct blood type you're instantly filled up to over 300% blood.

Now if you attach an IV drip with the wrong blood type you instantly get fucked with 1000u of toxin.

Credit goes to I Can Hear Colors for this one

X-TheDark:
fixes #1103

Two things that are also fixed:
1. You actually get the same amount of blood/reagents as there was in the IV Drip reagent thing (previously, before Cruix tried to fix it, 200 units of reagents resulted in you only getting ~120 due to a bug in how the fraction was calculated);
2. This unbreaks the IV Drip reactions, by fixing the use of `reaction()` proc - it doesn't use amount, it uses a volume multiplier, to make it's reagents react before injecting them (this prevents duplication of reagent objects);
#### Changelog

:cl: CentComm committee of Slightly Less Red Cross 
bugfix: ID Drips no longer make their reagents react on a planetary scale.
/:cl:

### DIFF
--- a/code/game/machinery/iv_drip.dm
+++ b/code/game/machinery/iv_drip.dm
@@ -117,8 +117,8 @@
 				if(istype(beaker, /obj/item/weapon/reagent_containers/blood))
 					// speed up transfer on blood packs
 					transfers += transfers_per_inject
-				for(var/i in 1 to transfers)
-					beaker.reagents.reaction(attached, INJECT, transfer_amount, 0) //make reagents reacts, but don't spam messages
+				var/reaction_volume = min(transfers * transfer_amount / beaker.reagents.total_volume, 1)
+				beaker.reagents.reaction(attached, INJECT, reaction_volume, 0) //make reagents reacts, but don't spam messages
 				beaker.reagents.trans_to(attached, transfer_amount * transfers)
 				update_icon()
 


ShadowDeath6:
bluespace bloodpack that injects 1000u of any reagent when???

code/game/machinery/iv_drip.dm
```diff
@@ -117,8 +117,8 @@
 				if(istype(beaker, /obj/item/weapon/reagent_containers/blood))
 					// speed up transfer on blood packs
 					transfers += transfers_per_inject
-				for(var/i in 1 to transfers)
-					beaker.reagents.reaction(attached, INJECT, transfer_amount, 0) //make reagents reacts, but don't spam messages
```
X-TheDark, line 121: This line caused the IV Drip to react 2.5x of it's reagent volume (transfer_amount = 2.5, but that was used as a volume multiplier) with the victim...2/4 times...in one cycle...`reaction()` is really weird (also, reaction and reagent removal is separate, so you can freely react thousands of units of reagents, while only removing a fraction of that. Cool stuff).

Cruix: Removing the for loop defeats the whole purpose of the IV drip, which is to inject many times at a small volume for effects that require a small volume for reaction_mob(), like saline glucose.

X-TheDark: Yeah, I figured after making this that you did this for SalGlucose.

X-TheDark:
Cruix fixed this properly in his thing, seeing as this nerfs SalGlucose blood regen thing.

PR #1117, opened 2016-10-29
Fixed IV drips injecting far more blood than they should

### LINKED ISSUE #1103: Blood bags/IV drips are bugged like hell

Cruix:
fixes #1103

:cl:
fix: IV drips no longer inject way too much blood.
/:cl:

### DIFF
--- a/code/game/machinery/iv_drip.dm
+++ b/code/game/machinery/iv_drip.dm
@@ -112,13 +112,14 @@
 	if(beaker)
 		// Give blood
 		if(mode)
-			if(beaker.volume > 0)
+			if(beaker.reagents.total_volume > 0)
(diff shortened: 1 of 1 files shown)


utahclock:
Does this mean you can no longer overfill a persons blood to say 105% ?

Cruix:
No, you still can, you just actually need to use that much blood to do it instead of magically generating it from nowhere.

utahclock:
Ah i see. Cool.

PR #1120, opened 2016-10-29
Makes S.S. Shitbucket MK. 3 properly load

Tenryumi:
The ship would not correctly load into the world, and we learned that the issue pertained to the docking port. So I looked at the variables of the docking port...

the fix

was the tiniest 
most minor

thing, ever.
#### Changelog

:cl:
fix: The S.S. Shitbucket MK. 3 should now properly load onto the map.
/:cl:


X-TheDark:
Docking/ship related code is very specific, it's true. Tiniest of mistakes makes the whole thing not work.

PR #1138, opened 2016-10-30
Fixed icon error with ash walkers

Cruix:
I gave ash walkers a different ID than regular lizards so you can use the magic mirror or varedit to change a human's species to them, but apparently icons depend on that var.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types.dm
+++ b/code/modules/mob/living/carbon/human/species_types.dm
@@ -128,7 +128,7 @@ datum/species/lizard/before_equip_job(datum/job/J, mob/living/carbon/human/H)
 
 /datum/species/lizard/ashwalker
 	name = "Ash Walker"
-	id = "ashlizard"
+	id = "lizard"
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
Why would you wanna be ashwalker

Paragonid:
To bring singularity spawner to the station

Time-Green:
Absolutely unacceptable. This will completely fuck blobs even more than they already are, screw over revheads early and late game, contribute to the growing holoparasite problem, reinforce the emag-revolver meta, and destroy the viability of shadowlings. I don't know how you could even consider this change.

Did it ever occur to you how much this would affect traitors, abductors, changelings, nukeops, deathsquads, and megafauna? This game is a complex balancing act of interwoven checks and balances, maintained by some of the most highly trained and educated spaceman experts in the world, and this PR shows complete disregard for that haphazard yet workable tapestry which we have spent five years holding together with duct tape and bubblegum.

If by some miracle this is merged we may as well set the probability of all gamemodes to zero except extended so that the playerbase can at least eke out some meager entertainment from roleplay, since any resemblance of balance will be utterly and completely destroyed.

PR #1193, opened 2016-11-11
Fixed some stuff

Cruix:
Fixed changeling rounds not starting
Fixed adding "random AI law" admin verb not working,


X-TheDark:
I think I deliberately skipped this when I fixed the antag weights because I thought we didn't have pure changeling rounds. My bad.

PR #1200, opened 2016-11-12
[s] Made bluespace shelter capsules not work on the station.

Cruix:
Requested by GreatBigBen

Shelter capsules no longer work on the station z-level, because 90% of the time they are used to block hallways for no reason.

### DIFF
--- a/code/modules/mining/mine_items.dm
+++ b/code/modules/mining/mine_items.dm
@@ -264,10 +264,6 @@
 
 		playsound(get_turf(src), 'sound/effects/phasein.ogg', 100, 1)
 
-		var/turf/T = deploy_location
-		if(T.z != ZLEVEL_MINING && T.z != ZLEVEL_LAVALAND)//only report capsules away from the mining/lavaland level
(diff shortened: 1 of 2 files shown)


Stealthkibbler:
You may want to do the same with Ashwalker sinew tents as they are basically copy/paste code of the capsules.

Super3222:
Good change, but I feel like they should at least work in space.

EMT321:
I believe there are some limited uses to them on the station. Could you make them only trigger in the absence of oxygen, so they can be used in space and Lavaland?

Dagdammit:
From an in-game logistics standpoint, bluespace shelters in space don't make alot of sense; they're inflatable shelters that are supplying air by pumping in the atmosphere outside and refining it, or at least that's what I figured.

But that would be a shame from a meta standpoint, because there isn't much of an alternative. EMT's approach seems like a good solution.

Another could be if you were able to open a maintenance hatch on one of the machines in the back, & trigger a "repackage" option which caused the shelter to pack itself back up.

Dragonforcefive:
I applaud the person who fixed the capsule spawning on station crap. I always get annoyed when some random jackass spawns a shelter in the middle of Central Hallway when I'm Captain or HoS.

Cruix:
Now they always work in space.

PR #1208, opened 2016-11-13
New player rules now open into the browser.

oisin100:
We no longer use the rules file.
:cl:
bugfix: Rules now lead to the actual rule page.
/:cl:


ShadowDeath6:
But what if I don't want to switch to browser to look at rules

oisin100:
> But what if I don't want to switch to browser to look at rules

Tough luck.

PR #1212, opened 2016-11-15
adds a proc for when simplemobs attack themselves

ShadowDeath6:
fixes artificers also tell me if any other mob does this because I don't remember

:cl: ShadowDeath6
bugfix: Artificers can now heal themselves again.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/constructs.dm
+++ b/code/modules/mob/living/simple_animal/constructs.dm
@@ -67,11 +67,11 @@
 
 /mob/living/simple_animal/hostile/construct/AttackingSelf()
 	if(health < maxHealth)
-		adjusthealth(-5)
+		adjustHealth(-5)
 		visible_message("<span class='danger'>[src] repairs some of its own dents.</span>", \
 					"<span class='cult'>You repair some of your own dents, leaving you at <b>[health]/[maxHealth]</b> health.</span>")
 	else
-		M << "<span class='cult'>You cannot repair your own dents, as you have none!</span>"
+		src << "<span class='cult'>You cannot repair your own dents, as you have none!</span>"
 
 /mob/living/simple_animal/hostile/construct/Process_Spacemove(movement_dir = 0)
 	return 1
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/constructs.dm
```diff
@@ -59,17 +59,20 @@
...
-						   "<span class='cult'>You repair some of your own dents, leaving you at <b>[M.health]/[M.maxHealth]</b> health.</span>")
 		else
 			if(src != M)
 				M << "<span class='cult'>You cannot repair <b>[src]'s</b> dents, as it has none!</span>"
-			else
-				M << "<span class='cult'>You cannot repair your own dents, as you have none!</span>"
 	else if(src != M)
 		..()
 
+/mob/living/simple_animal/hostile/construct/AttackingSelf()
```
Super3222, line 68: You don't need this...

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -652,9 +652,12 @@
 	if(harness && harness.on_melee_attack(src, T, 1))
 		return
 	if(T == src)
-		return
+		AttackingSelf()
 	T.attack_animal(src)
```
Super3222: They're going to attack themselves twice... why not just get rid of the check?

code/modules/mob/living/simple_animal/constructs.dm
ShadowDeath6: that's the entire point of this pr you fuk

Super3222: You still don't need it.  You just need to remove the check in `AttackingTarget`.

ShadowDeath6: I added the check for a reason super

Super3222: So simple animals can't attack themselves? That shouldn't even be a feature. EVERYONE accidentally attacks themselves - it's just apart of the game. Exiling simple animals from it is snowflake-y.

ShadowDeath6: Simple animals can only ever harm themselves if they click themselves. You can claim snowflake or hugbox but it's an irritating feature with literally no upsides and this PR adds a way for special effects to happen and fixes the problem.

Super3222:
bad pr imo and we should get rid of that check

ShadowDeath6:
if you can provide a real reason to convince me why we should allow simple animals to hit themselves other than "because other people can hit themselves" or "hugbox" I'll close the pr

Time-Green:
I like it, alot of times I attack myself to see if there's something special

There isn't

Super3222:
If you're making a proc for attacking themselves, at least make them attack themselves on regular. Artificers can do their heal thing on their own time.

PR #1226, opened 2016-11-18
[s] Fixed being able to talk while unconscious

### LINKED ISSUE #1224: Ling chat.
Hello. I just wanted to notice that if you try to chat in changeling channel (say ":g*") while not a changeling makes one say something in a regular way. As found out later, it was for changelings only. The funny part of this is that one can still speak like any other generic crew member _while unconcious_.
In addition, I suspect all similiar non-radio channels will give the same result to unconcious human nonantags.
I guess that's enough.

I apologise for my bad knowledge of English.

Cruix:
Fixes #1224


ShadowDeath6:
oh

Time-Green:
Boooooo
I didn't even get the chance to abuse this

PR #1251, opened 2016-11-21
Adds Pretty Filters for IC chat methods

utahclock:
:cl:
tweak: tweaks pretty filter to include IC text.
/:cl:

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -9,10 +9,6 @@
 		return
 	usr.say(message)
 
-
-
-
-
 /mob/verb/whisper(message as text)
 	set name = "Whisper"
 	set category = "IC"
@@ -22,10 +18,6 @@
 		usr << "<span class='danger'>Speech is currently admin-disabled.</span>"
 		return
 	say(message) //only carbons actually whisper, everything else just talks
-
-
-
-
 
 /mob/verb/me_verb(message as text)
 	set name = "Me"
@@ -38,7 +30,6 @@
 	message = trim(copytext(sanitize(message), 1, MAX_MESSAGE_LEN))
 
 	usr.emote("me",1,message)
-
 
 /mob/proc/say_dead(var/message)
 	var/name = real_name
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -2,18 +2,30 @@
 /mob/verb/say_verb(message as text)
 	set name = "Say"
 	set category = "IC"
+	message = pretty_filter(message)
+	log_say("[name] : [message]")
 	if(say_disabled)	//This is here to try to identify lag problems
 		usr << "<span class='danger'>Speech is currently admin-disabled.</span>"
 		return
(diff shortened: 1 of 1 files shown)


Super3222:
admin abuse

Kmc2000:
For those that dont know, pretty filters stop you using bigotry

Jc1469:
It'd probably be better to just leave IC as it is.  Replacing IC text makes an IC mess if it's set for bigotry.
Other than that, it's just another niche meme admin tool for trolling at the end of a round.

oisin100:
If we do this we should alert admins when someone trips a pretty filter.

matskuman5:
"If we do this we should alert admins when someone trips a pretty filter."

^^^^^^^^

Otherwise logs are fucked and people can just say "oh it was a joke i just typed in BANMEADMINS and hit enter"

utahclock:
I understand this is controversial. I suppose I can an admin message trigger. Not sure how to get it to state x said n word which triggered filter. The n word part would be part i would need help with . if any one has ideas.. On how to do that pm me on discord.

code/modules/mob/say.dm
```diff
@@ -2,19 +2,31 @@
...
+	log_say("[name] : [message]")
 	if(say_disabled)	//This is here to try to identify lag problems
 		usr << "<span class='danger'>Speech is currently admin-disabled.</span>"
 		return
 	usr.say(message)
 
+
+
+
+
```
monster860: **screams**

AAAAAAAAAAAA! Too many blank lines!!!

code/modules/mob/say.dm
```diff
@@ -2,19 +2,31 @@
...
+	log_whisper("[name] : [message]")
 	if(say_disabled)	//This is here to try to identify lag problems
 		usr << "<span class='danger'>Speech is currently admin-disabled.</span>"
 		return
 	say(message) //only carbons actually whisper, everything else just talks
 
+
+
+
+
```
monster860: **screams**

code/modules/mob/say.dm
```diff
@@ -27,9 +39,12 @@
 
 	usr.emote("me",1,message)
 
+
```
monster860: **screams**

monster860 (changes requested):
You're adding blank lines for 4 no raisin.

Super3222:
tbh I don't like it.

Time-Green:
I see seriously no problem with this?
Do you guys want people to scream F----- out loud? Isn't that against the rules?
Just notify admins if they do.

Super3222:
IC chatter shouldn't be hindered. One of the many things to laugh about is someone using that word, and than getting banned for it. Filtering it so unfun, and while it may be against the rules we aren't Nazis, and nobody is going to get triggered over it. I am all for the dedication against slurs and other degrading profanities but if they do get triggered than they should stay off SS13.

At the very least an admin notification should be sent to ASAY. Nothing more

PR #1264, opened 2016-11-22
Tweaks borers to no longer be able to claim soulless bodies

utahclock:
Tweaks borers so that SSD and bodies left behind from cloning can no longer be infected.
:cl:
tweak: Tweaks borers so that SSD and bodies left behind from cloning can no longer be infected.
/:cl:


code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -184,6 +188,7 @@
 
 	src << "<span class='userdanger'>You begin disconnecting from [victim]'s synapses and prodding at their internal ear canal.</span>"
 
+
```
Super3222: what's up wtih this space though

PR #1267, opened 2016-11-22
Fixes borer huds not disappearing

utahclock:
Fixes borer huds not disappearing after borers leave host.

Note: This PR contains the pr from #1264 as well as the changes that allow the borer huds to disappear. 
I can change this if #1264 is not approved.
:cl:
bugfix: Fixes borer huds not disappearing after borers leave host.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer_powers.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer_powers.dm
@@ -207,9 +207,6 @@
 		if(victim.mind)
 			host << "<span class='danger'>Something slimy wiggles out of your ear and plops to the ground!</span>"
 			host << "<span class='danger'>As though waking from a dream, you shake off the insidious mind control of the brain worm. Your thoughts are your own again.</span>"
-		for(var/image/hud in victim.client.images)
-			if(hud.icon_state == "borer")
-				victim.client.images -= hud
 		leave_victim()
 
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -52,6 +52,10 @@
 		src << "<span class='warning'>[H] cannot be infected! Retreating!</span>"
 		return 0
 
+	if(!H.mind.active)
```
Super3222, line 55: this is from https://github.com/yogstation13/yogstation/pull/1264. leave a note in your original PR post that this pr requires that one to be merged.

Stealthkibbler:
Also I think there may be a bug where a borer cannot enter a host that originally had a borer in it, because when I tried to let a borer in after expelling an annoying one, it couldn't get in.

code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -203,10 +207,13 @@
 		if(victim.mind)
 			host << "<span class='danger'>Something slimy wiggles out of your ear and plops to the ground!</span>"
 			host << "<span class='danger'>As though waking from a dream, you shake off the insidious mind control of the brain worm. Your thoughts are your own again.</span>"
-
+		for(var/image/hud in victim.client.images)
```
Super3222: This should honestly get placed in `leave_victim` in case that proc is forced instead of this one.

PR #1274, opened 2016-11-23
Borer Vessel Tweaks

Super3222:
#### Changelog

:cl:
tweak: When you remove a borer vessel from someones brain, there'll be an extra check that ensures the borer will be removed.
tweak: Borer Vessels no longer qdel, so you can stare at their ugliness.
/:cl:


Time-Green:
Does it have wifi tho?

PR #1280, opened 2016-11-24
Cargo can talk on science channel, scientists can talk on Supply channel.

Stealthkibbler:
The first step to miner powercreep.

:cl: Stealthkibbler
tweak: Supply and Science staff can now speak on eachothers channels allowing for jolly co-operation between supply and science, this also means scientists will get to hear the horrifying bloodcurdlings screams of miners as asslizards gut them alive.
/:cl:


Time-Green:
I think this is a bad idea, I get where it's coming from and I can certainly see why you could use it...
But it's unrealistic, might aswell merge cargo and science chat then, wich is also a bad idea.

Super3222:
I do see some use for it. On the other hand in honesty, I didn't completely like it when I saw it on /tg/.

(so maybe now miners will take telesci beacons with them to lavaland)

ShadowDeath6:
Let's just move mining to science at this point

seriously what will this change other than giving miners more annoying radio noise except now it's purple

If a miner isn't going to do their job and mine then science bitching about it will do less than nothing

Stealthkibbler:
Well for a first it will allow them to actually communicate with eachother, so instead of say, having to individually pda each miner or talking in common hoping miners see it among the other chatter scientists only have to say in supply what they need and how much or simply have to ask if any miners are still alive. Miners can also say in science chat what kind of gadgets they need so science can make them in advance, this will mitigate miners having to wait around R&D for extended periods of time because some scientists work under the mentality of "I wont make mining gear until they specifically ask me for that stuff.

It rewards communication, I honestly dont see any problem with it. Also nobody cares about realism in a 2d spacemans game where carp are the number 1 nemesis and you can synthesize oxygen in water form. If you still think of this from a realism perspective then you're doing it wrong. We aren't rp, we dont need to adhere to the laws of physics.

Dragonforcefive:
I completely agree with stealth on this one. Whenever I'm a miner, I tire of the QM never asking science for cool new gadgets whenever we bring any ores up. As a matter of fact, once, I was going to simply keep the ores on Lavaland if science didn't "pay" is back.

Time-Green:
I know you can consider this useful, but it just shouldn't be. Might aswell remove their whole fucking channel then.

Why not put down a cargo encryption key on the RnD tables?

Stealthkibbler:
If a xenobiologist doesnt want to listen they can simply disable the supply channel, if an R&D scientist wants to listen he can keep it enabled, nobody is being FORCED to listen to supply there's a reason why you can toggle channels on your headset. Respectively miners are going to talk on science channel if they need something and R&D and robo are going to want to listen because these miners are bringing in minerals. If a miner wants to be a cunt to science and not communicate at all he can just turn off his radio.

Also that's a horrible idea time for the reason that if someone takes the key and dies with it or fucks off then anyone who actually wanted to use it can't.

I dont see any bad coming from this change, at all, like literally the worst you can come up with is radio chat clutter? That happens anyway just because of how byond is designed, just having common channel turned on causes radio clutter and supply and science channels are usually rarely talked into anyway.

You are fretting over nothing.

ShadowDeath6:
Then have science able to talk on supply, not the other way around

Slinky5890:
I'd be fine with the QM and RD having the ability to talk in both channels, but there's no real point in cluttering up scientist's screens with idle cargo chatter. The requests console is used to pester miners and it seems to work fine enough.

Stealthkibbler:
It doesnt work fine enough Slinky, because it requires the cargo staff to be aware that it exists, I've gone rounds where I've had to pda the cargo staff multiple times after sending in multiple requests to check up on the miners and to top it off cargo players are the laziest, scratch their ass instead of do anything type of people I've seen in all of ss13.

I've pda'd cargo staff to tell them to checkup with miners on supply channel, only to instantly be messaged back "I dunno" and nothing else because people cant be assed  to even talk in the supply channel.

Cargo shouldnt even have to checkup on the miners since they should be focused on supply and demand, but its become evident that all our good cargo players, Mason Miami, Dzlan and a few others dont play cargo anymore and its been like this for months, cargo has just been so unreliable due to laziness, I don't want to waste 10-20 minutes of my round maxing out R&D with no way of directly finding out if the miners have fell into a ditch and died.

Maybe we havent gotten this kind of complaint because ever since bekons and keek got banned we've lost the only two regulars who're dedicated to R&D and developed the best guides to doing it before the tg rework. I guarantee you if they were still in the community they would support this 100% of the way as R&D scientists.

And to add to everything else.

Its not going to fucking clutter their radios, if you've ever even played cargo, miner or sciencist you'd know that even with full staff on both departments people barely use the department radios. This change is to hopefully encourage people to use it more.

If the one thing you're worried about in ss13 is clutter, then I got news for you, you can expand, minimize and maximize the text that shows up using your mousewheel and by grabbing the box so that you can see more text. It literally takes like 5 seconds to do and its something I do EVERY round and I've never complained about radio clutter ONCE because of it.

AsV9:
I'm okay with this change, but I don't feel like it's necessary. It certainly doesn't break anything, and if cargo really doesn't want to talk to science they can just turn off the channel on their headsets.

I don't find this change necessary as I feel like when you play science or cargo you don't have to put in much effort to PDA a scientist or just ask on common.

/shrug

Either way doesn't really matter. Up for merging and seeing how it turns out.

ghost:
/tg/ did this.

IT works fine.

There's no issue. Miners on /tg/ can speak on science channel.

No conflicts.

No point to arguing.

PR #1281, opened 2016-11-24
Adds some more ore spawners to some of our ruins

Super3222:

One problem that we've had on lavaland is the fact that miners almost never come back to the station. That's either because they;
1. Are finding cooler ruins to hang out at.
2. Are hunting for tendrils.
3. Got killed by a random fuana.

So in this PR, I gave a few of the maps some ore spawners.

Here are the types of ore spawners:
1. Normal Version - drops a certain amount of ore, and can drop more than one type.
2. Generous Version - It's the normal version in BIGGER numbers. However I'm not insane, so don't expect to walk in on 50 diamond ores
3. Bananium Version - it only has bananium for our clown maps.

Here's what I added:
- 2 bananium spawners in the clown biodome map.
- Replaces ALL the gold in cugganscove with a random ore spawner, and a generous ore spawner.
- Gives the hermit an ore spawner for trading with miners
- Gives the miming drill an ore spawner despite the main attraction of the ruin is the fact that a giant invisible shield blocks it from ashstorms.
- Clown facility got bananium... somewhere!

#### Changelog

:cl:
rscadd: Ruins are now spawning with ore! This is for miners to take back with them to the station!
rscadd: Clown Biodome, Cugganscove, Hermit Hideout, Miming Drill Camp, and the Clown Faciltiy all spawn with ore!
/:cl:


ShadowDeath6:
What does this mean, you added special subtypes of ore pockets that drop more or less ore?

Super3222:
> you added special subtypes of ore pockets that drop more or less ore?

They're map spawners which spawn loot when the server initializes. It makes more sense if you read it from the code.

PR #1290, opened 2016-11-28
Gibtonite is now STEALTHTONITE

ghost:
gibtonite now looks like uranium on mining scanners

because i dislike miners and also it used to be disguised as diamond and thus a legitimate 'dig dig dig FUCK' moment

#### Changelog

:cl:
rscadd: Be careful when mining. The ore scanner is less than accurate.
/:cl:

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -160,7 +160,7 @@
 	mineralAmt = 1
 	spreadChance = 0
 	spread = 0
-	scan_state = "rock_Gibtonite"
+	scan_state = "rock_Uranium"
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
:^(

doesn't gibtonite explode on kinetic hit?

ghost:
is there an evil grin emote

MayhemSailor:
Worthwhile change. Adds some challenge back to mining.

Stealthkibbler:
I like this change and the fact that you made it uranium, since diamond would be too obvious but uranium is uncommon enough that it can easily be mistaken.

Good change.

Dragonforcefive:
God, please help all miners

AndySok:
they can use mining scanners if they aren't screubs

PR #1294, opened 2016-11-28
Toxin Mixing Piping Optimization.

MayhemSailor:
#### Optimization changes for piping in Toxins Mixing.

The use of hidden piping and a jumbled set up lead to confusion for some new to the advanced sciences of bomb manufacture. This change optimizes the layout, and changes pipe types from hidden to visible to ensure there is minimal confusion when working with gaseous plasma.

Old on the left, new on the right. The lighting has been fixed since screencap.

##### Changelog

:cl:
rscadd: NT Engineering Division has had their pipefitters in to optimize toxin mixing for the lowest common denominator among the science staff.
/:cl:


MayhemSailor:
From DM:

AsV9:
Looks OK. Have you tested it works? I mean, there is no reason to suspect it shouldn't but pipes are weird and it's good practice.

Easy to make mistakes..

MayhemSailor:
Yup, I tested it while I was grabbing the screencap, works like a charm, or at least as well as toxins mix normally works

Time-Green:
Never had problems with this, but I can imagine doing so.

I like it.
Could you place something down to kinda decorate?

Stealthkibbler:
The warning lines should coat all the way around the piping, it looks rather off with only one side having warninglines. You can rotate them by creating a new instance and editing the DIR variable, the DIR works from 1-10, some numbers in DIR will do nothing. Generally the even numbers will rotate warning lines properly, try 0, 2, 4, 6 and 8 to see if it rotates the lines to your desired direction.

Even though its designed around the intention that you're generally going to be working on the pipes in one direction, it just looks ugly to have the lines only curve around one side.

MayhemSailor:
@Stealthkibbler the tiling has already been fixed see my latest comment before this.

@Time-Green while I haven't had many issues, I've dealt with enough tickets at this point I think the change is warranted. What would you suggest for decor?

Alek2ander:
>ensure there is minimal confusion when working with gaseous plasma.

This is not the Nanotrasen way

PR #1295, opened 2016-11-29
Fix the corners on the emergency shuttle.

monster860:
:cl: monster860
fix: Emergency shuttle corners in transit no longer show normal space as opposed to transit space. Why this wasn't fixed before is beyond me.
/:cl:


MayhemSailor:
10/10

PR #1300, opened 2016-11-30
Fixes the Ninja mask

Stealthkibbler:
Ninja mask now uses the correct LLA one and not the old one.

:cl: Stealthkibbler
fix: Space Ninjas mask now shows correctly on its mob sprite.
/:cl:


ShadowDeath6:
:^)

PR #1304, opened 2016-11-30
fixes my old shitty marrow weaver code + small buffs

ghost:
gets rid of some useless shit

fixes a bug where weavers had two statements

gets rid of a useless proc and replaces it with vars instead

allows marrow weavers to calm down

angry weavers do more damage and inject more toxin, fair since they're literally almost dead and spore toxin is weak

weavers will liquefy the organs of their kills and drink them, healing themselves

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
@@ -821,7 +821,7 @@
 				"<span class='danger'>[src] drools some toxic goo [L]'s innards...</span>",
 				"<span class='danger'>Before sucking out the slurry of bone marrow and flesh, healing itself!</span>",
 				"<span class-'userdanger>You liquefy [L]'s innards with your venom and suck out the resulting slurry, revitalizing yourself.</span>")
-			adjustBruteLoss(-target.maxHealth/2)
+			adjustBruteLoss(-L.maxHealth/2)
 			if(ishuman(L))
 				var/mob/living/carbon/human/H = target
 				for(var/obj/item/bodypart/B in H.bodyparts)
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
@@ -784,18 +784,18 @@
 	var/poison_type = "????"
 	var/poison_per_bite = 5
 	var/buttmad = 0
- 	var/melee_damage_lower_angery0 = 13
- 	var/melee_damage_upper_angery0 = 16
- 	var/melee_damage_lower_angery1 = 15
- 	var/melee_damage_upper_angery1 = 20
+	var/melee_damage_lower_angery0 = 13
+	var/melee_damage_upper_angery0 = 16
+	var/melee_damage_lower_angery1 = 15
+	var/melee_damage_upper_angery1 = 20
 	
 /mob/living/simple_animal/hostile/asteroid/marrowweaver/adjustHealth(amount)
 	if(buttmad == 0)
 		if(health < maxHealth/3)
 			buttmad = 1
 			visible_message("<span class='danger'>[src] chitters in rage, baring its fangs!</span>")
-			melee_damage_lower = melee_damage_lower_rage1
-			melee_damage_upper = melee_damage_upper_rage1
+			melee_damage_lower = melee_damage_lower_angery1
+			melee_damage_upper = melee_damage_upper_angery1
 			move_to_delay = 8
 			speed = 3
 			poison_type = "spore"
@@ -804,8 +804,8 @@
 		if(health == maxHealth)
 			buttmad = 0
 			visible_message("<span class='notice'>[src] seems to have calmed down, but not by much.</span>")
-			melee_damage_lower = melee_damage_lower_rage0
-			melee_damage_upper = melee_damage_upper_rage0
+			melee_damage_lower = melee_damage_lower_angery0
+			melee_damage_upper = melee_damage_upper_angery0
 			poison_type = initial(poison_type)
 			poison_per_bite = initial(poison_per_bite)
 	..()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
-		melee_damage_lower = 15
-		melee_damage_upper = 20
-		poison_type = "spore"
-		poison_per_bite = 3
-		desc = "A big, angry, toxic spider. It looks really, REALLY unhappy. It's badly wounded."
-	else
-		poison_type = initial(poison_type)
-		poison_per_bite = initial(poison_per_bite)
-		speak_emote = list("chitters")
+	if(buttmad == 0)
```
Time-Green, line 793: This looks awesome, you're improving. One thing tho, could you change this to:
if(!buttmad)

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
+		if(health < maxHealth/3)
+			buttmad = 1
+			visible_message("<span class='danger'>[src] chitters in rage, baring its fangs!</span>")
+			melee_damage_lower = melee_damage_lower_rage1
+			melee_damage_upper = melee_damage_upper_rage1
+			move_to_delay = 8
+			speed = 3
+			poison_type = "spore"
+			poison_per_bite = 5
+	else if(buttmad == 1)
```
Time-Green, line 803: Else if is not needed, since it basicly checks if the previous one was false. Unless quantum physics are into play, this is not needed.
Also, change it to 
if(buttmad)

if(buttmad) checks if the variable is TRUE
if(!buttmad) checks if it's FALSE
Everything but 0, null and FALSE is considered positive.

All not needed, but looks alot better.

Cruix: Don't say negative and positive, say true and false.
Negative numbers evaluate to true.

Time-Green: I meant the same thing, just with the wrong words.
I corrected it

ghost: There is a reason it's not an if else.

I plan to add variations of weavers that will get more agressive as they're hit, and possibly repath weavers altogether to put them in the giant spider file so I can use them as a branching point for more spider types since i'm going to be tackling simplemob spiders fairly soon.

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
+		if(health == maxHealth)
+			buttmad = 0
+			visible_message("<span class='notice'>[src] seems to have calmed down, but not by much.</span>")
+			melee_damage_lower = melee_damage_lower_angery0
+			melee_damage_upper = melee_damage_upper_angery0
+			poison_type = initial(poison_type)
+			poison_per_bite = initial(poison_per_bite)
+	..()
 
 /mob/living/simple_animal/hostile/asteroid/marrowweaver/AttackingTarget()
```
Cruix: Right now sentient weavers are able to eat corpses forever for infinite health. Replace this proc with this to make it so they can only eat bodies that still have organs:

/mob/living/simple_animal/hostile/asteroid/marrowweaver/AttackingTarget()
	..()
	if(isliving(target))
		var/mob/living/L = target
		if(target.reagents)
			L.reagents.add_reagent(poison_type, poison_per_bite)
		if((L.stat == DEAD) && (health < maxHealth) && ishuman(L)))
			var/mob/living/carbon/human/H = L
			var/foundorgans = 0
			for(var/obj/item/organ/O in H.internal_organs)
				if(O.zone == "chest")
					foundorgans++
					qdel(O)
			if(foundorgans)
				src.visible_message(
					"<span class='danger'>[src] drools some toxic goo [L]'s innards...</span>",
					"<span class='danger'>Before sucking out the slurry of bone marrow and flesh, healing itself!</span>",
					"<span class-'userdanger>You liquefy [L]'s innards with your venom and suck out the resulting slurry, revitalizing yourself.</span>")
				adjustBruteLoss(-L.maxHealth/2)
				for(var/obj/item/bodypart/B in H.bodyparts)
					if(B.body_zone == "chest")
						B.dismember()
			else
				src << "<span class='warning'>There are no organs left in this corpse.</span>"

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
-		poison_type = "spore"
-		poison_per_bite = 3
-		desc = "A big, angry, toxic spider. It looks really, REALLY unhappy. It's badly wounded."
-	else
-		poison_type = initial(poison_type)
-		poison_per_bite = initial(poison_per_bite)
-		speak_emote = list("chitters")
+	if(buttmad == 0)
+		if(health < maxHealth/3)
+			buttmad = 1
```
Super3222, line 795: 1's can be traded for TRUE, and 0's that you aren't using for counting can be traded with FALSE. 

For instance, if you wanted buttmad off you can toggle it like this.
`buttmad = FALSE`.

But using it to set a timer, like a clock that winds up every second, would be clumsy.

**In this instance,** you want `buttmad = TRUE`

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
-		melee_damage_lower = 15
-		melee_damage_upper = 20
-		poison_type = "spore"
-		poison_per_bite = 3
-		desc = "A big, angry, toxic spider. It looks really, REALLY unhappy. It's badly wounded."
-	else
-		poison_type = initial(poison_type)
-		poison_per_bite = initial(poison_per_bite)
-		speak_emote = list("chitters")
+	if(buttmad == 0)
```
Super3222, line 793: buttmad == FALSE

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
+			visible_message("<span class='danger'>[src] chitters in rage, baring its fangs!</span>")
+			melee_damage_lower = melee_damage_lower_angery1
+			melee_damage_upper = melee_damage_upper_angery1
+			move_to_delay = 8
+			speed = 3
+			poison_type = "spore"
+			poison_per_bite = 5
+	else if(buttmad == 1)
+		if(health == maxHealth)
+			buttmad = 0
```
Super3222, line 805: buttmad = FALSE

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
```diff
@@ -783,41 +780,55 @@
...
 		var/mob/living/L = target
-		if(L.reagents)
+		if(target.reagents)
 			L.reagents.add_reagent(poison_type, poison_per_bite)
+		if(L.stat == DEAD)
+			src.visible_message(
+				"<span class='danger'>[src] drools some toxic goo [L]'s innards...</span>",
+				"<span class='danger'>Before sucking out the slurry of bone marrow and flesh, healing itself!</span>",
+				"<span class-'userdanger>You liquefy [L]'s innards with your venom and suck out the resulting slurry, revitalizing yourself.</span>")
+			adjustBruteLoss(-L.maxHealth/2)
```
Super3222, line 824: You're going to want to round this value.
`adjustBruteLoss(-(round(L.maxHealth/2)))`

Super3222 (changes requested):
Needs updating where TRUE and FALSE statements can come in hand. Needs to round a value at the end. Probably needs to multiply an integer by negative 1 in case it doesn't turn out negative in the end.

code/modules/mob/living/simple_animal/hostile/mining_mobs.dm
Cruix: He has already stated that he plans to add more levels of buttmadness, so TURE and FALSE are not appropriate here.

Super3222: Oh so it's not Boolean

PR #1306, opened 2016-11-30
adds an admin message if a pretty filter is applied.

utahclock:
adds an admin message if a pretty filter is applied.

:cl:
rscadd: adds an admin message if a pretty filter is applied.
/:cl:

### DIFF
--- a/code/modules/admin/prettyfilter/prettyfilter.dm
+++ b/code/modules/admin/prettyfilter/prettyfilter.dm
@@ -79,6 +79,7 @@
 
 	usr << "\"[msg]\" becomes: \"[pretty_filter(msg)]\"."
 
+
 // Enter a piece of text and have it tested against the filter list
 /client/proc/add_pretty_filter(line as text)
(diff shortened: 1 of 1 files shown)


X-TheDark:
I'm concerned about admin message spam, but I'm not exactly sure how often something like this pops up.

utahclock:
It would only pop up when someone says a filtered word.

utahclock:
And its rare that someone says filtered words but it does happen.

This is what happens admin log wise if filtered word is said multiple times,

Super3222:
It still shouldn't replace IC text.

utahclock:
Tell cruix to revert ic then, this message alert works on ooc.

AsV9:
Feels a bit excessive to have logging on OOC filters imo.

Cruix:
I did not test this before merging, and apparently you did not either. Every single thing a non-admin says in OOC will spam adminlogs with messages.

PR #1307, opened 2016-12-01
Toxins Mixing Piping (new)

MayhemSailor:
#### Optimization changes for piping in Toxins Mixing.

Now that Utah has taught me about Map Merge, re-PR'd from [https://github.com/yogstation13/yogstation/pull/1294](PR#1294)

The use of hidden piping and a jumbled set up lead to confusion for some new to the advanced sciences of bomb manufacture. This change optimizes the layout, and changes pipe types from hidden to visible to ensure there is minimal confusion when working with gaseous plasma.

Old on the left, Pay no attention to the right... Updated image here:

##### Changelog

:cl:
rscadd: NT Engineering Division has had their pipefitters in to optimize toxin mixing for the lowest common denominator among the science staff.
/:cl:

### DIFF
--- a/_maps/map_files/TgStation/tgstation.2.1.3.dmm
+++ b/_maps/map_files/TgStation/tgstation.2.1.3.dmm
@@ -41111,6 +41111,8 @@
 /obj/machinery/atmospherics/components/unary/vent_pump{
 	dir = 2;
 	external_pressure_bound = 0;
+	frequency = 1441;
+	id_tag = "burn_out";
 	initialize_directions = 1;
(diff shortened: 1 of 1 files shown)


monster860:
Can you also add a control console and a gas sensor so that we can turn off the pumps

MayhemSailor:
@monster860 shouldn't be too much of a problem to do. It would be able to pinpoint the mix as well, although I question its value in a space that can be vented like that.

MayhemSailor:
@monster860 it has been added in.

PR #1317, opened 2016-12-02
[s] Makes IC pretty-filters silent

AsV9:

Now whenever you trip a pretty filter in whisper or say, instead of BANMEADMINS it just goes silent instead. The player is notified with a "You fumble over your words." and admins get a message stating what was changed.

Muh immershuns.

It has been tested, it works fine.
Also I wanted to test how adding `[s]` works but this doesn't actually have to be hidden away from players though.

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -2,7 +2,12 @@
 /mob/verb/say_verb(message as text)
 	set name = "Say"
 	set category = "IC"
+	var/oldmsg = message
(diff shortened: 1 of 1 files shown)


matskuman5:
YESYESYESYESYE

Much better than the immersion breaking shit

However they should be informed that it breaks the rules.

Maybe put in "You fumble over you words. (See rule 0.1)"

utahclock:
Make it alert admins for ooc or there can be conflicts with #1306

PR #1327, opened 2016-12-04
[s] Makes it even more harder than the last to cheat as AI

monster860:
I removed the "dim" icon state from the file so that you can't use that as a reference point. It's never used anyways.


monster860:
FML, I fucked up the secret tag.

supernoodles13:
good job. totally a secret now

monster860:
Eh, the discription is pretty vague.

Earthykiller127:
brb finding a workaround

monster860:
So have you found that workaround yet?

PR #1343, opened 2016-12-06
updates power terminals to have yellow cables.

utahclock:
Since Gowst, made changes to make cables yellow in maint, power terminals were not color matched. This matches cables.

:cl:
rscadd: Power terminals now have yellow cables.
/:cl:


Time-Green:
Could you perhaps make them the color of wire wich was used to construct it? Then you set the base color to yellow.

PR #1362, opened 2016-12-08
Makes the box of beakers hold large beakers

ghost:
fucking this is really needed already 
##### Changelog

:cl:
rscadd: The box of beakers now holds large instead of normal beakers.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/boxes.dm
+++ b/code/game/objects/items/weapons/storage/boxes.dm
@@ -166,7 +166,7 @@
 /obj/item/weapon/storage/box/beakers/New()
 	..()
 	for(var/i in 1 to 7)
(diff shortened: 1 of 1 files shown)


matskuman5:
H U G B O X

matskuman5:
No but seriously what will normal beakers ever be used for again?

Super3222:
This is a pretty hugboxy PR. IMO this doesn't need to happen given that the games not set up so you can access every little feature so easily. 50/50 beakers are fine, and chemists start with enough beakers to sustain themselves as long as they aren't going into grenades.

Cargo can also massprint large beakers with enough glass (which isn't hard to get).

matskuman5:
Exactly what Super said ^^^^

Cruix:
Leave a comment

PR #1367, opened 2016-12-09
Fixed tcomm sat APC

Cruix:


Nickvr628:
This fixes the weird floating APC that was not offset properly? Do the Signal Techs have access to it yet or is it still engineer only?

Cruix:
It does fix that. Engineers have never had access to that room and signal techs don't either, that room requires teleporter access, which I think only the captain and maybe RD have.

PR #1372, opened 2016-12-10
Ports Projectile Dismemberment

utahclock:
Per request of Super3222, adds projectile dismemberment (though at the time only the plasma cutter will, can add var to other projectiles as needed or seen fit.)

Ported from TG.

:cl:
rscadd: Projectiles can now dismember. 
/:cl:


ShadowDeath6:
How does this pass travis? We don't have spellblades.

code/modules/projectiles/guns/magic/staff.dm
```diff
@@ -59,3 +59,9 @@
 	item_state = "honker"
 	max_charges = 4
 	recharge_rate = 8
+
+/obj/item/weapon/gun/magic/staff/spellblade/hit_reaction(mob/living/carbon/human/owner, attack_text, final_block_chance, damage, attack_type)
```
ShadowDeath6: I don't think we have this?

utahclock: ah whoops forgot that to remove that line.

PR #1377, opened 2016-12-10
Fixes emojis

Super3222:

Whenever you had ":" in your ooc message an annoying box would come up asking if it was meant for ooc.
Well, we don't use that for radio communication anyway. So this PR will disable that only for ":" so you can send emojis straight to OOC without pause.

### DIFF
--- a/code/modules/client/verbs/ooc.dm
+++ b/code/modules/client/verbs/ooc.dm
@@ -55,7 +55,7 @@
 	log_ooc("[mob.name]/[key] : [msg]")
 
 
-	if((copytext(msg, 1, 2) in list(".",";",":","#")) || (findtext(lowertext(copytext(msg, 1, 5)), "say")))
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
are you SURE we don't use : for anything

aka :h :b :n :u :literallyanychannel

Super3222:
Oh, I tried doing that in character just see if anything would happen. Nothing came up...
Anyway, maybe there's a way around it to see if an emoji CAN be created within the OOC message.

Kmc2000:
i prefer it like that, it only pops up when you nistype the emote, so it tells you that its wrong before you make an ass of yourself in OOC

Super3222:
I see your point, KMC2000.

ShadowDeath6:
It doesn't only pop up when you mistype the emote tho

Super3222:
I know, but that situation can apply to both cases. You're able to see what you exactly typed up, and whether or not it's the actual emoji name.

PR #1379, opened 2016-12-10
Gives miners a free webbing + adds a crafted version

Super3222:
We are at the point of deciding whether miners should start with webbing or not. Before you go off on your own experience, I want to take a moment of your time and ask you this. What is an "explorers webbing"?

It's a belt for stashing all of your loot, equipment, and whatever falls into your hands. You can only specifically place mining equipment inside of it, and it's not oriented for anything off of lavaland.

What a lot of people don't get is that it's the perfect tool for lavaland. What I mean by this is simple, miners are suppose to have them. There's not a doubt in my mind that at one point one of the coders on /tg/ who worked on lavaland though that it'd be a good idea to actually go through with that? But what happened?

A lot of people like to point at explorer webbings and call them OP. Fine, I'm okay with that. It's an extremely helpful tool that goes a long way with helping miners out.

What I want to reach out to you is that all it's doing is just helping out miners carry around their equipment and a little loot on the side. Is that really that awful? Of course not. Than what's making it so bad? I can't seem to understand. We've built lavaland up to be this massive place, with lots of ruins, artifacts, and all of these tools that you need, but here we are denying the one prospect that miners should actually care about?

If you want to blame the "Survival Capsule + Explorer Webbing" combo going mainstream you might as well look at the code.

What this PR is doing is giving miners a free explorer webbing.

The explorer Webbing will still have 5 storage slots. All of it's pretty vital.
The one put up in the mining vendor will still be up there for non miners.

Likewise, you can craft a primitive explorer belt for 2 sinew and 1 bone.

##### Changelog

:cl:
rscadd: Every miner starts with an explorer webbing.
rscadd: You can now craft a primitive explorer belt out of 1 stack of bone and 2 sinew.
/:cl:

I should also find a way to port Kor/tgs kinetic accelerator rework.


Super3222:
@Cruix Merge conflicts fixed.

PR #1381, opened 2016-12-11
Added spawn zones for signal technicians.

Cruix:


Genesis720:
I approve of this change. No more spawning randomly in the RD office now!

PR #1401, opened 2016-12-14
Explorer hoods will delete now to prevent undroppable hoods. Also allows laz injector to be emagged. 

ghost:
you're all fat

##### Changelog

:cl:
fix: The explorer hood will delete when dropped now to prevent undroppable hoods.
rscadd: Emag the laz injector to make it malfunction. Have fun.
/:cl:

### DIFF
--- a/code/modules/mining/equipment.dm
+++ b/code/modules/mining/equipment.dm
@@ -20,7 +20,7 @@
 	desc = "An armoured hood for exploring harsh environments."
 	icon_state = "explorer"
 	body_parts_covered = HEAD
-	flags = NODROP
+	flags = NODROP | DROPDEL
 	flags_inv = HIDEHAIR|HIDEFACE|HIDEEARS
 	min_cold_protection_temperature = FIRE_HELM_MIN_TEMP_PROTECT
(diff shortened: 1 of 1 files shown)


Cruix:
This will delete the hood whenever it gets retracted back into the cloak.
Inventory code is generally shit.

ShadowDeath6:
How does the TED work? Use that

ghost:
It works the same. There's no predefined way to do this. 

I can't exactly make it store the armor values either. The only thing I can think of is some really gimmick ass shitty thing of making applying goliath plates to the helmet qdel it and turn it into another repathed helmet and change the suit's retracted helmet to something else but that's awful.

I guess miners will just have to deal with 'don't retract the hood.'

ShadowDeath6:
I'd rather it bug out when decapitated than bug out when retracted, you toggle the hood a lot more often than you get your head cut off

Super3222:
Fixing a bug by adding in another bug obviously isn't going to pass.
There are also ways to do this. Like checking for whether the hood is on the ground in the first place, and qdeling it from there. HELMETS already do this.

ghost:
then if there's a way to do it you do it

PR #1406, opened 2016-12-15
Fixes powernet and zipties sprite

Time-Green:
Let's act like none of this ever happened...
Thanks to @oisin100 and @skull132 for doing everything.


code/modules/power/terminal.dm
```diff
@@ -92,3 +92,4 @@
 		var/image/I = new('icons/obj/power.dmi', "term_wire")
 		I.color = color2code(wiring_color)
 		overlays += I
+		return
```
oisin100: Why do you need this?

oisin100 (changes requested):

oisin100 (approved):

Issue #24, opened 2016-06-02
Ports : Everything by X-TheDark


X-TheDark:
### Any PR that was made by me will be looked at an ported over the next few days. After that I'll help out with everything else

PR : https://github.com/yogstation13/-tg-station-yog-rebase/pull/27

Includes things like:
- Drones passing through doors;
- Odysseus stuff;
- Syndiborg (Assault) stuff - recharging gren launchers, the works;
- **Job Whitelist system**
- etc

I'll probably **NOT** be porting over Pickpocket Gloves for now as it's a clusterfest of shitcode and hacky fixes to make it work. I'll see if I can rewrite it to be better.

X-TheDark:
And I'm done with porting my stuff over. As I said, Pickpocket Gloves will be looked at at a later date, I think I would really like to rewrite them (and include some extras).

PR #1412, opened 2016-12-16
[s]Bugfix #24

utahclock:
Bugfix #24 :)


AsV9:
totallyinconspicuoustitle

Cruix:
Thank god that's fixed it's been around forever

utahclock:
Dont merge yet i need to fix a couple things with floors.

Time-Green:
What even is this? Did you delete the whole map?

utahclock:
Things are fixed

Super3222:
This needs to be PR'd to the special branch or turned around, and added implemented into the code without deleting the old map. Take the latter, and you'll need to edit which map zlevel 1 should generate. I'll show you where that can be arranged.

PR #1414, opened 2016-12-17
Dual Unrelated Tweaks

Cruix:
Non-cheiftan ashwalkers only speak and understand ashwalker
AIs can now use vending machines

Technically we're not supposed to mix features and bugfixes, but I got permission from a headcoder so it's fine.


ShadowDeath6:
I'm not against it, but what was your reasoning for giving AI's vendo access

Cruix:
It just seems like something an AI should be able to do, especially if one type of Borg is capable of doing it.

Kmc2000:
> I got permission from a headcoder so it's fine.


X, D

Vendor access would be great to shit out all the candies

PR #1416, opened 2016-12-18
Lets clean up this sticky mess...

oisin100:
This has gone on long enough

:cl:
rscdel: Admins no longer have the stickyban verb, Rejoice!
/:cl:


MayhemSailor:
Bad puns...

Time-Green:
Are there any other ways to make sure people don't just use different accounts/IP's?

oisin100:
> Are there any other ways to make sure people don't just use different accounts/IP's?

Ban their CID and IP, Which we do anyways.

AsV9:
it's not like shitlers need to know that admins can't sticky them

PR #1238, opened 2016-11-19
ID fluff

Time-Green:
Been a little inactive lately, school has been consuming me. I hope this makes up for it.

Gives every single job a completely unique ID card. 

The colors do mean something, the big lines in the middle give away the department, and the smaller ones the job in that department.

For example, Janitor has a green line, because he is civilian, and the other thing purple, because janitors are purple. I tried to make this as convenient as possible, but there's a limit.

Some examples: (Not all jobs included)
id s and stuff
Oh yeah and this uses ovelays and stuff. And I know QM isn't a head, but it looks cool.
Custom job names will look like the classic ID's

:cl: 
rscadd: Gives every job a unique ID. It's very pretty and the clowns ID will probably give you an epileptic attack.
/:cl:

PR #1448, opened 2016-12-22
Refactors ID_fluff

Time-Green:
1. Changes the switch into a list, it looks prettier now and safes space and memory
2. Making custom ID's will not change the colors on it, unless it previously didn't have them
3. Adds Signal Technician and Mining Medic to the mix

Original PR: https://github.com/yogstation13/yogstation/pull/1238
:cl:
tweak: Optimizes ID_fluff
/:cl:
Thanks to @Cruix , @oisin100 and @skull132 for showing me how Assoc lists work.

### DIFF
--- a/code/game/objects/items/weapons/cards_ids.dm
+++ b/code/game/objects/items/weapons/cards_ids.dm
@@ -319,6 +319,7 @@ update_label("John Doe", "Clowny")
 	"Quartermaster" = list("cargo","silver"),
 	"Cargo Technician" = list("cargo","brown"),
 	"Shaft Miner" = list("cargo","black"),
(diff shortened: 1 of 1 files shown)


Cruix:
Actually the list uses more memory than a switch statement, it's just that the memory it takes up is negligible and more than made up for by the speed improvement and code maintainability.

Time-Green:
That means that lists take more memory, but the space it saved makes up for it?

Cruix:
Oh, I thought the list was static, but it's not. As it is it does take up memory, but only for the length of the function call, which does not matter in the grand scheme of things.

It is still better than the switch because less code means less chance of a problem occurring, and the easier problems are to fix.

PR #1466, opened 2016-12-25
dual energy sword can no longer dismember

ghost:
100% guarantee to dismember, really

>target leg
>click spaceman
>they are stunned

murderbone tool, goodbye

also i fixed a stupid warning for hulks b/c that was badly worded

##### Changelog

:cl:
rscadd: The dual energy sword can no longer dismember. 
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -275,7 +275,7 @@
 
 /obj/item/weapon/twohanded/dualsaber/attack_hulk(mob/living/carbon/human/user)  //In case thats just so happens that it is still activated on the groud, prevents hulk from picking it up
 	if(wielded)
-		user << "<span class='warning'>You can't pick up such dangerous item with your meaty hands without losing fingers, better not to!</span>"
+		user << "<span class='warning'>Your hands are too clumsy to pick this up safely.</span>"
 		return 1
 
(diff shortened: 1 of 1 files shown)


silicons:
10/10

ghost:
i mean tqbh maybe this would fly on /vg/ or /tg/ but here? nah.

i'm a shitty unrobust guy and i still killed 80% of the crew with it, as does anyone who buys this. people just keep trying to taze the guy and instant stun in melee range when the only way to stun the guy without a shotgun or slips is to be in melee range is meh

GrayRachnid:
this is a bad pr

Nickvr628:
I am torn on this. Can we just make it LESS likely to dismember instead of disabling it completely?

Super3222:
> I am torn on this. Can we just make it LESS likely to dismember instead of disabling it completely?

Not unless we want it to affect ALL weapons with the IS_SHARP flag.

Stealthkibbler:
flashbangs are a thing y'know? I've taken down double esworders many times, Seiga was one of the more recent victims of my flashbang spam. There's also the fact that pepperspray goes right through the esword and stuns the wielder if they arent wearing a mask. People just need to not be shit at combatting it. Flashes are also super useful when used in hand as you can blind the individual. Also its not like they can buy anything that combos extremely great with the esword aside from noslips. It's an all-in item which is hardcountered by 4 different items that security has access to.  Even fucking krav maga counters it for christ sake, dont even get me started on beepsky.

Nickvr628:
Also pepperspray takes them out quickly too. Unless they are wearing a mask that covers all of their face.

PR #1479, opened 2016-12-26
Xenomorph larva now give a 20 second prompt

Stealthkibbler:
Thanks to Super for basically spoonfeeding me this

:cl: Stealthkibbler
fix: Xenomorph Larva now give a prompt when attempting to grab a ghost for bursting. If the prompt gets intrusive on your ghastly observing consider disabling the xenomorph pref. Thanks to Super and Cruix for supplying the code on this one <3
/:cl:


Cruix:
You need to use the client of the observer from the list, not the ghost mob

Super3222:
Ah, whoops. The proc `pollCandidatesForMob()` is suppose to be used exactly as described in it's name. To use it correctly, you have to delete everything that talks about a client in the top half of that proc call (lines 70 to 84). Then replace line 91 with the proc call (it says `new_xeno.key = C.key`). Use a compiler though, and test it. I'm just speaking from what I see in the code.

Stealthkibbler:
Will anything bad happen if we dont do that? It worked perfectly fine when I tested it in local.

Time-Green:
What does this do to the random xeno event? And will we be able to tell the difference?

Stealthkibbler:
It just means that everytime a larva is ready to burst it will prompt players for 20 seconds if they want to be that larva or not. If you have the pref disabled tho it shouldnt prompt so it should be good.

Jc1469:
Issue with doing this is that both Xeno and Borer REQUIRE players.  That's why I imagine there was never any prompt.
Neither can really achieve anything without players.  Borers can't even be created without an observer with the preference enabled.

Ultimately, all this just to fix that one building that is still just a sadistic torture box for syndies, I don't think it's worth it.  People being randomly tossed into _actual_ xeno bodies is why Xeno is good, you don't need to be sat waiting for a noise to be one, it just gives you it.  I just really don't think something which has spawned in 3/4 rounds I've observed should have that ability at all.

Stealthkibbler:
Theres nothing wrong with having a prompt though, the prompt also prevents afks from being chosen for the role. This is also a problem with normal xeno, larva tends to burst in the most inopportune of moments and being afk for even a split second can get you killed.

It's also far from a sadistic torture box, the xenos can kill the syndies as easily as the syndies can kill them, both sides are at fairly equal combat prowess when it comes to fighting. It's more akin to an environment with the intention of simulating xeno on carbon human combat.

It's also worth noting that if you're seeing the ruin that much then its just rng, I've seen it not spawn for like  5 rounds in a row.

Super3222:
@Jc1469

Borers are getting the same treatment in https://github.com/yogstation13/yogstation/pull/1483.

Stealthkibbler:
@Cruix Talked it over with super and the fix you guys suggested has the problem of spawning the larva regardless of the whether it successfully gets a ghost or not. It'd be better to stick with the current var that super has given me.

ShadowDeath6:
you could have an examine for larva seeing if it's braindead and if there's no key you can just enter it

Cruix:
I made a PR to this branch.

Just look at the last commit for the changes, everything else is bringing it up to speed with master.

PR #1484, opened 2016-12-26
Minor Lavaland Ruin Fixes and Xenomorph Facility Update

Stealthkibbler:
Some changes to the Xenofacility that make it feel less dull and open and more compact.

:cl: Stealthkibbler
tweak: Xenofacility ruin has had some big additions, rooms have been moved around, the power grid is now powered by Pacmans instead of a backup SMES meaning scientists will have to occassionally check up on the pacmans every 20-30 minutes. The Flavor text has been dumbed down for the Scientists and new ways to interact with xenomorphs have been introduced and hinted to, more weapons added and weapons are now stored in one big area, every scientist gets a hardsuit now and due to a recent breakthrough in xenomorph communication 50 units of cardboard have been provided. There's a few other stuff as-well, scientists will have to explore the facility and get accustomed to the new stuff they have available to them.
fix: Removed more falsewalls that have managed to sit under the radar yet again.
tweak: Goliath Mound now has much better loot for miners willing to brave it
del: Removed the Mosin from the Russian Bunker
fix: Watcher Spiral has less watchers and the chest no-longer gets pushed into lava
fix: Xeno Facility Atmos Problems should stop happening even if the observation windows get broken, if you notice it happening at all report it on the github or discord coder chat with @Stealthkibbler included in your message.
/:cl:


Jc1469:
will the turrets pick a fight with megafauna?

Stealthkibbler:
If megafauna spawns nearby, yes. But that happens with just about any ruin.

Jc1469:
just making sure megafauna can come and destroy the entire thing because i still don't like the general idea of the ruin in general. :+1:

Mordokh:
One thing I'd add if you already didn't add it would be a proper surgery room, enough supplies to make a sec borg, maybe one or two syndicate minibombs in case things go wrong. It'd be also great if the syndies had a point based system similar to cargo where if they ship live xenomorph specimens, or parts to the main syndicate facility to in return get points that they can spend on tools like the sec borg I mentioned, or perhaps for an AI for the syndie facility that would make the job much easier. Because as it is, without the proper tools, syndies get flooded by aliens too fast, and by adding the point based system they'd have an incentive to go slow and build up their point count by shipping xenos off to the main syndie base. You could also add something like bomb collars for the xenos.

Slinky5890:
Ooh i like mordokh's idea there a lot. Might be a bit too advanced/complicated for a ghost role but i'm a fan

Slinky5890:
Might be a good idea to add a few indestructible cameras too, so the scientists can actually OBSERVE the xenos in their habitat. Just have a few thrown around, maybe in the main pockets.

Stealthkibbler:
@Mordokh 
There are now two surgery rooms, there won't be any silicons in the facility as the presence of a borg will allow the syndicates to escape since the borg will be able to unbolt the exit airlocks and tank the perimeter turrets. Any form of explosives will severely render parts of the base unusable and irrepairable since destroyed tiles revert to lava and you cannot build on lava without an RCD. The point base system sounds good in theory, me and Super wanted to do something similar for another ghost role that didnt make it into the rework, the only problem with a point based system is the syndies will just construct a temporary containment near the shipping area and consistently ship xenos. Interaction between xenos and the syndies should be more on gameplay aspects and less on numerical progression; which is what this facility aims to do. In-fact the only role in the entire game that ever needs to capture live xenos is the space ninja but it's irrelevant due to the fact xenos will succumb when they arrive to the capture den as there are no weeds to pull them out of crit. Bomb collars for xenos MIGHT work, however xenos can resist out of any restrain instantly, instead the syndicates have access to a couple mindslave implants which work on xenos.

@Slinky5890 
I've tried fiddling with a camera network, there's big problems with it such as consoles refusing to link up to the network and the ai being able to look through the cameras despite them being put on a seperate network from the station or mining.

PR #1489, opened 2016-12-27
Shuttle calls lengthened remake.

ghost:
The shuttle will take 20 minutes to arrive on blue, 30 on green, and 10 on red. 

Green shuttle calls are called 'crew transfer' shuttles.

Mostly to deal with 'theres an antag CALL IT' shuttles.
##### Changelog

:cl:
rscadd: Shuttle times have been lengthened.
/:cl:

### DIFF
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -13,7 +13,7 @@ var/datum/subsystem/shuttle/SSshuttle
 		//emergency shuttle stuff
 	var/obj/docking_port/mobile/emergency/emergency
 	var/obj/docking_port/mobile/emergency/backup/backup_shuttle
-	var/emergencyCallTime = 6000	//time taken for emergency shuttle to reach the station when called (in deciseconds)
+	var/emergencyCallTime = 15000	//time taken for emergency shuttle to reach the station when called (in deciseconds)
 	var/emergencyDockTime = 1800	//time taken for emergency shuttle to leave again once it has docked (in deciseconds)
 	var/emergencyEscapeTime = 1200	//time taken for emergency shuttle to reach a safe distance after leaving station (in deciseconds)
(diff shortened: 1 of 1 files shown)


GrayRachnid:
20 green, 10 blue, 5 red.
20 green, 15 blue, 10 red.

change or riot.

AndySok:
Yogstation becomes Tau Ceti confirmed

Bittern12:
So we just have to sit through rounds where everything is broken and nothing is going on except avoiding being spaced or burning to death for 20 minutes? 
 No please.

Jc1469:
still no.

Shuttle being called for dumb reason is an OOC issue. Admins should be dealing with it, not the salt fairy.

ghost:
Admins can't deal with it.

@aphielanis You know that's not the case.

matskuman5:
What in gods name

Red alert almost never gets activated due to heads being busy/incompetent/dead.

10 minutes is a good amount of time, 20 minutes is just silly

ghost:
nope matsu, nope

it's because people call the shuttle for no reason all the time

matskuman5:
Then we'll get people calling the shuttle for no reason, except it takes 10 minutes longer to arrive

ghost:
good

Notamaniac:
Really torn up on this PR, while there needs to be less incentive to immediately call the shuttle (which isn't adding more rules, adding more rules is bad and should be frowned upon), all this PR does is encourage you to call the shuttle sooner since you have to wait longer, and only really gives antags 10-5 extra minutes to complete their objecitves. And it makes recalls so much more painful to sit through.

ghost:
On the contrary, people will be reluctant to earlycall the shuttle before the antags are dealt with due to the risk of a painful recall. Additionally, earlycalling to the extent of calling as soon as it's available because 'we might need it' is powergame, so.

PR #1490, opened 2016-12-27
sparks trigger welder tanks

ghost:
don't change your RCDs near a fueltank people

additionally, this means you can use timer assemblies, etc, under a fueltank to set it off when the sparks generate
##### Changelog

:cl:
rscadd: Sparks will now trigger welding tanks. 
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/effects_sparks.dm
+++ b/code/game/objects/effects/effect_system/effects_sparks.dm
@@ -35,7 +35,7 @@
 
 /obj/effect/particle_effect/sparks/Bump(A)
 	if(A)
-		if(istype(A, /obj/structure/reagent_dispensers/fueltank))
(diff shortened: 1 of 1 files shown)


Cruix:
This is such a bad idea

Jc1469:
I agree, but this is the one PR adam will ever make that I wont disagree with.  I want to see that one clown who always make a proximity igniter to be annoying blow the fuck up.

ghost:
hey, we almost had the same thing when igniter assemblies created hotspots

this just lets the AI in on the !!FUN!!

Silvaplex256:
oh my god yes

finally people can actually /use/ igniter assemblies

Ratvar:
Does it mean i can falsesend someone as abductor,so it create sparks?I already use it to draw dicks,but this is even better.

Frostben:
wooooh welder tank IEDS here we go!!

PR #1503, opened 2016-12-28
Fixed being able to strip abstract items

Cruix:
Even though you cannot see them in the strip panel.


ShadowDeath6:
so did the strip slots show up as empty, or were you opening the panel before they got the item?

Cruix:
The slots show empty

PR #1511, opened 2016-12-29
Changed the airlock AI interface window

### LINKED ISSUE #1468: Will the NanoUI for AI/Borg door control ever be re-added?
'We had such a thing?'

Cruix:
Old: 
door4

New: 
door1
door2
door3

It's not quite as fancy as the old nanoui one, but it's way better than what it looks like right now.

Fixes #1468
:cl:
rscadd: The AI door interface window has been updated to look more like the old nanoui one.
/:cl:


GrayRachnid:
THANK YOU

ShadowDeath6:
Bueno

Jc1469:
I kind of like the old one because it's more detailed, and actually explains what something being off or on does. More nub frenli.

But I do often have to re-click a door for the menu to actually update though, which is annoying.

PR #1521, opened 2016-12-31
Nerfed the loot and costs of some ruins

Cruix:
Increased the costs of some of the more loot-heavy ruins, and made some loot only drop some of the time. Two of the three spellbooks in the wizard den now only have a 10% chance of spawning, and all of the guns in the abductor terror ship only have a 40% chance of spawning.

Also ashwalkers only get one stack of mythril because they don't need three.


Stealthkibbler:
rip lavaland powercreep

PR #1537, opened 2017-01-01
removes a reference

ShadowDeath6:
this is a reference that i'm not ok with.


GrayRachnid:
wow shadow i didn't think you'd be so petty

ShadowDeath6:
you thought wrong then >:^)

Super3222:
pettydeath6
but eh, when I added it I liked the guy. Not so much anymore though.

PR #1549, opened 2017-01-02
Blob buff

Time-Green:
Now takes 2 welding attacks for a normal blob thing to die.

Blob was extremely weak due to people insta-killing the normal blobs. This makes it take 2 hits. I tested it.

:cl:
tweak: Normal blob tiles now take 2 welder hits to kill
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/theblob.dm
+++ b/code/game/gamemodes/blob/theblob.dm
@@ -341,6 +341,7 @@
 	maxhealth = 20
 	health_regen = 1
 	brute_resist = 0.25
+	fire_resist = 0.75
 	atmosblock = 1
 
 /obj/effect/blob/normal/scannerreport()


Stealthkibbler:
BLOB POWERCREEP

ShadowDeath6:
cool

Notamaniac:
It used to be two hits for most blobs, when did that change?

Time-Green:
No clue, but it did.

Jc1469:
#826 

Das what nerfed blob so much.

ghost:
This was already fixed @Time-Green.

It already takes two welder hits to kill a normal blob tile and one laser hit to kill a normal blob tile.

The reason welders were oneshotting the blob is because of the chem type.

ShadowDeath6:
All blobs except those weak to burn should die in more than 1 hit

Currently only those that resist burn will take 2 welder hits

ShadowDeath6:
I just checked the code, regular blobs have 30 health and the burn multi is 1.5x, doing exactly 30 damage to it

Time-Green:
I tested it with one type and assumed it worked. I'll test with a few more to make sure it really takes 2 hits with all of them

Time-Green:
I checked, and yes, this should apply to all of them. I could also find nothing of different reagents having different resistences

ShadowDeath6:
nice

PR #1557, opened 2017-01-03
Comments out drone language from NTSL

Jc1469:
People can still use translators, pAIs, recorders, all sorts of crap.  But intercomms and station bounced radios will hopefully not do it now, with this and Cruix' changes.

:cl:
tweak: Drone language has been removed from NTSL.  Maybe now people will have a harder time speaking to humans as drones and breaking their laws. 
/:cl:


ghost:
that's a shame.
I should make ratvar a translatable language

Jc1469:
intercomms and handhelds actually translated all languages, which was dumb.  Cruix fixed that already, this is just to fix the easiest of the like.. 6 ways a drone has to be understood by humans.

Cruix:
This just removes the variable with the language's bitflag, someone who knows what they are doing could still make NTSL translate drone

Jc1469:
> someone who knows what they are doing could still make NTSL translate drone

so nobody will ever be able to translate drone. sounds good to me.

Cruix:
Leave a comment

PR #1567, opened 2017-01-03
Makes xenobiology more secure

matskuman5:
image

The double airlock system works exactly like Virology.

Also added some r-walls to make sure the whole place is secure.

Had to reroute disposals a bit but I tested it and it should work.

This should hopefully prevent hordes of slimes from eating the station every round. This should also make the tator objective of stealing slime extract a bit harder.

:cl:
rscadd: Xenobiology has been made more secure by adding a proper airlock like in virology, and removing the maintenance airlock.
/:cl:


matskuman5:
What the fuck is this merge conflict? Hold on lemme try and figure this out

Time-Green:
Aaaaah, it's even harder now to kill those darn xenobiologists

matskuman5:
Huh? It works?

matskuman5:
WWWWWWWWWWWWWWEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE

Notamaniac:
Neutral on this, this makes xeno a good base for cults , makes it harder for the steal slime objective since now you have to hack through via science, then two more doors

Time-Green:
Not really a fan of this. It's already hard enough to kill a xenobiologist.
But that aside it's pretty good

GrayRachnid:
where did apc go?

Kmc2000:
against really, xeno needs to be breakable innable and outable really we already have one death fort (viro) on station.


Imho, keep the maint door, so you can still hack in and have the airlock stuff as well

Mordokh:
Great addition

Stealthkibbler:
@Kmc2000 you can legit just cut through the wall

Dagdammit:
@Stealthkibbler which one? I see r-walls in every direction.  

I'd support a small buff to xenobio's main entry, maybe two airlocks and the inner one is reinforced windows with electrified grilles, but this looks more like a deliberate effort to remove all access vulnerabilities from the workspace- which misses a fundamental point point of map design in SS13, I'd say.

Also do we really have significant problems with wild xenobio slimes escaping to flood the station? In my experience, if crew elsewhere in the station are going down to slimes it's probably because A: xenobio released them deliberately (usually sentient ones), B: the slime abductor gland, or C: that red slime which gets spawned by the pyro anomaly.

matskuman5:
I'm closing this, I can't be bothered to fix the merge conflict

PR #1572, opened 2017-01-04
removes the description of megafauna from the admin log popup of megafauna being lead to places. because why.. why would it be there?

Jc1469:
why the fuck


Stealthkibbler:
GUARDIANS OF NECROPOLIS

Jc1469:
I want to know every time a monstrous creature protected by heavy shielding is lured to nowhere.
I definitely do.  Every single fucking time.

Time-Green:
John, use correct commit names dammit.
"why was this a thing" is a bad name
If someone were to look in the code and see the last changes, they'd see that.

Jc1469:
and they'll see something that should not have been a thing. and not bother checking it.  it's the perfect name.

ShadowDeath6:
The only reason I can think is if the creature was named something else, such as the renamed ash drake in one of the ruins.

Jc1469:
im sure the admins know about that one ash drake by now.  if not I can re-add desc, add type, add health, god.  is fin.

PR #1580, opened 2017-01-05
Admins are now alerted when an anomaly is created.

matskuman5:
Tested it, works perfectly.

:cl:
rscadd: Admins are now alerted when an anomaly is created.
/:cl:


ShadowDeath6:
doesn't it tell you when anomalies are made over announcements tho

matskuman5:
Yeah but it only tells the general location

Plus anomalies are generated a few moments before the announcement comes in

PR #1589, opened 2017-01-06
[s] Fixes invisible TTVs

X-TheDark:
Made on behalf of @Super3222 

Funny, but let's not.


McDonald072:
Tag as fun removal and its good to go

PR #1592, opened 2017-01-07
Fixes an error in which mecha did not have relevant narsie and ratvar act procs

ghost:
le coffin of doom


ShadowDeath6:
make ratvar just force eject you so you don't get immediately obliterated then converted while dead

ghost:
true, but i don't want to have him spam eject people in mechs

ShadowDeath6:
then have him break the mech when it ejects the person

PR #1594, opened 2017-01-07
Blob tweak

Time-Green:
In my previous PR, I forgot lasers.
Now lasers instadestroy blob tiles and welding tools 2 hit them.
:cl:
tweak: Normal blob tiles can be instashot with lasers again
/:cl:


Slinky5890:
Damn, I was liking the superbuffed blob

ShadowDeath6:
bad pr this makes xray lasers even more op against them grr

Jc1469:
I don't think you're going to find a place to _balance_ the blob.  Blob tiles dying in 1 laser is pathetically weak.  Blob tiles dying in 2 lasers means you can only kill 5 blob tiles per gun.  Imo all nerfing should be done to the reagents and maybe spore/naut damage.  That said, the naut is only as strong as the reagent.

Time-Green:
This is what ot originally was anyway

ghost:
no it wasn't @Time-Green 

originally it was 2 lasers to kill 1 tile

PR #1599, opened 2017-01-08
Shapechange now transfers the health of the user

Cruix:
This means that mice will have 20 times their normal health and won't die in one second to vacuum, but juggernauts will only have 40% of their normal health. Of course, none of this matters, since the only form anyone ever chooses is the infinite taser stunbot, which has the same health as humans anyway.

It also means that hurting a shapechanged wizard actually means something.

:cl:
tweak: Using shapechange will transfer the health and maximum health of the original mob to the creature, and taking damage while transformed will affect the user when they revert to human.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/shapeshift.dm
+++ b/code/modules/spells/spell_types/shapeshift.dm
@@ -45,6 +45,9 @@
 	var/mob/living/shape = new shapeshift_type(caster.loc)
 	caster.loc = shape
 	caster.status_flags |= GODMODE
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
what happens if a wiz has artificers heal him as jugg then reverts

Cruix:
He will be healed

Jc1469:
I still love the wizard that shapeshifted into a mouse and got killed by a mousetrap at roundstart.

PR #1601, opened 2017-01-08
AO now has access to monitor-all-tickets so they can see tickets.

Jc1469:
kill me


Slinky5890:
Will this still let them reply?

Jc1469:
No, it wont. :smile:

Slinky5890:
Well that's all good then! Don't see why you'd have a problem with them just viewing tickets.

Jc1469:
I forgot the verb existed because I always had it enabled, and it auto-enabled when I was given a rank again. I wanted to them to see _tickets_ just not the ticket menu.

ShadowDeath6:
commit name

PR #1612, opened 2017-01-10
No more c4 insta-dead bullshit

Jc1469:
still Fixes #bullshit
also when someone copypasted the plastic file for c4 they left out a few things like c4 being put on a locker instead of in it. but thats not what this pr is for
:cl:
tweak: C4 and X4 are heavy as fuck and weigh down things they are placed on, making them unable to effectively embed inside of a person. They also don't go on bolas because fuck that.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/explosives.dm
+++ b/code/game/objects/items/weapons/explosives.dm
@@ -77,6 +77,8 @@
 		if(W.sharpness)
 			//user << "<span class='notice'>You should really kill yourself.</span>"
 			return
+	if(istype(AM,/obj/item/weapon/restraints/legcuffs/bola))
+		return
 	if((istype(AM, /obj/item/weapon/storage/)) && !((istype(AM, /obj/item/weapon/storage/secure)) || (istype(AM, /obj/item/weapon/storage/lockbox)))) //If its storage but not secure storage OR a lockbox, then place it inside.
 		return
 	if((istype(AM,/obj/item/weapon/storage/secure)) || (istype(AM, /obj/item/weapon/storage/lockbox)))
--- a/code/game/objects/items/weapons/explosives.dm
+++ b/code/game/objects/items/weapons/explosives.dm
@@ -72,11 +72,6 @@
 		return
 	if(loc == AM)
 		return
-	if(istype(AM,/obj/item))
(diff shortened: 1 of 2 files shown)


Super3222:
You should check for 'embed_chance' in conjunction with sharp.

Jc1469:
From the testing I was doing, every object has an embed_chance. 45, I believe.  Sharpness is what determines if it can actually embed or not.

Super3222:
Alright, I checked and yeah everything has a 45% chance to embed. IS_SHARP doesn't have to do with it though. Anywho, can you implement a similar change for bolas.

Jc1469:
I know of a few ways around this, but if it gets to the point where people figure out ways to still do it, more power to them I guess.

Ratvar:
smthngsmthngspeargrenadesblackpowderchl3mines,chemist can freely make something 1000% better then C4 bolas in half a minute.No C4 sharpitems means you cannot C4 captain rapier?No C4 for some summonitem miner weapons,nu,nu,bad. Nerf C4 bombing people into gibs with ease,maybe?

enailz:
Removes fun

ghost:
@Ratvar what did you even such say

also 'this is op so this other op thing should stay' isn't an argument

enailz:
Nerf the c4 throwing star throw distance and we good.

Jc1469:
Melee range unsurvivable death? No.

ShadowDeath6:
Bombs are bad good pr

Cruix:
For an object to imbed on throw the throw_speed has to be at least 4.

drderp3635:
C4 throwing star is a meme that rarely even works, because c4 takes ages to plat and if you miss you are royally fucked. Bad PR.

Ephemeralis:
Just make C4 TD -1 while attached to an embedded object. Trivial fix, keeps fun, nerfs unfun.

As in, actually -1, not 0 or whatever.

Alek2ander:
This is literal snowflakery.
Make c4-ed items lose their embed chance instead, or forbid attaching c4 to items of weight class 3 or lower or something like that.

Cruix:
It is snowflaky but so are legcuffs in general.

PR #1617, opened 2017-01-10
Stops megaphone spam.

Jc1469:
works and shit


Slinky5890:
Where's the removes fun tag?

PR #1621, opened 2017-01-10
Fixes the mining medic airlock

matskuman5:
Adds in a new name and ID so whenever the main medbay door is opened the mining airlock doesn't get opened as well

:cl:
bugfix: Fixed a bug where the mining medic airlock would be opened whenever the medbay airlock was opened.
/:cl:


Cruix:
The old garage door opener prank

PR #1623, opened 2017-01-10
Adds the unacidable var to turfs

Time-Green:
Adds the unacidable var to turfs. It's used in alien acid, clf3 and thermite.
:cl:
rscadd: Adds the unacidable var to turfs
/:cl:

Conflicts with: https://github.com/yogstation13/yogstation/pull/1618


ShadowDeath6:
Won't that nerf clf3 nades to not cook through tiles or am I wrong

if it does I'm not complaining tho

Time-Green:
It should go straight through.
The only on-station exception is the reinforced floor.

matskuman5:
Good work Time

ghost:
Time it shouldn't affect CLF3 or thermite.

It should affect the alien corrosive acid. That's all.

Time-Green:
Both of them were acid based, so I figured I should grab it along.
If the general consensus is not to include CLF3 and thermite, I'll change it.

Issue #1624, opened 2017-01-10
"High capacity water tank ? #372" No.2


00aquir:
""A specialised high-pressure water tank for holding large amounts of water."
but it holds the same amount of water as a normal one..."

AdamElTablawy: "This should be fixed now. Re open if not the case."

It still contains 1000 units... nothing even changed..

Time-Green:
I just checked, and yep, it has only 1000

ParagonPrime:
I got an email telling me the old issue was closed and I just spent two minutes laughing as soon as I saw this posted.

Stay classy, yogs :')

ToGWtF:
it just holds water specially :^)

Jc1469:
https://github.com/yogstation13/yogstation/blob/f589a9ca72e7273ba8d85d7d115da0ad4de6ce75/code/modules/reagents/reagent_dispenser.dm#L36

Damn, sure was hard to find where this mistake was made :rareparrot:
Max reagents allowed in them is 1000.  Trying to add 100000 to a container able to hold 1000 adds 1000.
Why do people just not test their code, even if it's simple.

PR #1629, opened 2017-01-11
Adds two zeroes to high-cap water tank capacity.

Jc1469:
#1624 

I question  why nobody else ever did this

:cl:
fix: High capacity water tanks are now actually high capacity water tanks.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_dispenser.dm
+++ b/code/modules/reagents/reagent_dispenser.dm
@@ -83,6 +83,7 @@
 
 /obj/structure/reagent_dispensers/watertank/high/New()
 	..()
+	create_reagents(100000)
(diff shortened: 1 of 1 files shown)


Stealthkibbler:
BOTANY POWERCREEP REEEEEE

Time-Green:
100000 is a little much. What about 2000?

Jc1469:
How bout no.  It's already set to 100000, it's just that some jackass forgot to change the max reagents, effectively making the code not work.

Time-Green:
No, it's already rare to empty a normal tank. Having it at fucking 100000 is insane, as it would never get emptied. Put it at 2000-5000

Cruix:
This is a fix, not a rebalance

PR #1631, opened 2017-01-11
Makes lavaland babymode; stuff no longer spawns within a 5x5 radius of labor or mining

Jc1469:
no longer will we be spawnfucked by ash ketchem, wrigley and avion.
:cl:
tweak: New anti-bullshit shielding has been installed near the mining station and labor camp. Now dangerous dangerocities will not.. I aint got an IC word for spawn, so close to them.
/:cl:

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -343,6 +343,10 @@
 		if(istype(loc, /area/mine/explored))
 			return
 		for(var/atom/A in urange(12,T))//Lowers chance of mob clumps
+			if(T.x < 75)
+				return
+			if(T.y < 55)
+				return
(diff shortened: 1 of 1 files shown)


Alek2ander:
Hardcoded coordinates are bad. What if lavaland layout is changed?
Use a hidden landmark that sets the protected zone or something like that instead.

Jc1469:
I doubt the layout is going to be an issue any time soon, considering how much of an area was already covered by default in the south west corner, which has nothing in it other than lava.  If anything, this allows better for coders to add things. 
And it's just values that people can change at the end of the day, and the commit is pretty clearly named so people can change it if they have to.

ShadowDeath6:
ban cruix for making quality of life label

monster860:
IC word for spawn: "Suddenly appear"

PR #1650, opened 2017-01-13
Cult ports and changes. 

ghost:
Added-

Offer rune. Replaces sacrifice/convert rune by fufilling the duties of both. Port from /tg/. Also heals the newly converted for some of their brute and burn. 

Extra help with learning how to cult.

Void torch. Allows cultists to warp items to other cultists. Port from /tg/.

Mind datum to sacrifices. Currently used for raise dead. Port from /tg/.

Removes-

Sacrifice rune. For the above reasons.

Convert rune. For the above reasons.

Leftover code.

Tweaks-

Cult sword. Now has blockchance, making it less useless. Also can be thrown to embed. Might remove that, if it's too strong, but to be fair you just chucked your only weapon at the guy, you're probably dead.

Flagellant robes. Instead of -100 to every stat, -50 instead. Might see some use from them now.

Veil shifter. Now has 4 charges instead of 2.

Revive dead rune. Port from /tg/. Instead of requiring two corpses (and having no cult checks and also not even finishing the fucking check to revive the guy) requires you to have more sacrifices than you have revived people. If this is true, the 'revived' counter goes up by one, and the person is revived. 

All constructs now have a standard speed of 1. This is about the same as a human.

Wraiths have a speed of 0. They're still fast. However, they are much more fragile. 2 lasers or a single slug should kill it.

Artificers have a speed of 2. This is slower than a human, but faster than a jugg. This cements their purpose as a support role, not a 'run around and click every wall to become cult then ram people constantly while spamming MM' role.

Harvesters have the best traits of all of them combined, minus artificer self repair. This is because why the fuck are the end-game nar'sie constructs weak ffs.

To be done if people think it's a good idea-Disabling rune use on non-roundstart locations or locations in space, anywhere. Lavaland mining station would count. Lavaland would not. This is to disable space base cheese cults.

Manifesting constructs still deducting health from the user as long as they are alive the same as manifested human.

##### Changelog

:cl:
rscadd: The revive dead rune has been changed. Now, each sacrifice will 'charge' it. If the number of sacrifice charges is greater than the number of cultists revived by any of the runes, a cultist can be placed on the rune to be revived.
rscadd: The sacrifice and convert runes have been placed by the 'offer' rune, which fulfills the purpose of both. 2 cultists will convert. 3 cultists will sacrifice a living target. If a target cannot be converted the rune will sacrifice them.
rscadd: The void torch has been added. This item, attainable through archives, can be used to warp items to other cultists.
rscadd: The flagellant robes now convey a 50% increase in damage taken instead of a 100% increase in damage taken. They still provide the same speed boost. Which is massive. Use them.
rscadd: The cult sword can now block incoming projectiles or melee hits, and can be thrown to embed. If this is too powerful it will be changed or reverted accordingly.
rscadd: Veil shifter now has 4 charges, up from 2.
rscadd: All constructs except for the wraith are about the same speed as a human, or slower. The artificer in particular is slower. Wraiths are far more fragile than before. Harvesters, spawned by nar'sie, are much much much stronger.
rscadd: As usual, if you have any questions or complaints ping me and i'll respond. 
/:cl:


code/game/gamemodes/cult/cult_items.dm
```diff
@@ -7,7 +7,9 @@
 	sharpness = IS_SHARP
 	w_class = 4
 	force = 30
-	throwforce = 10
+	block_chance = 50
```
ShadowDeath6: imo this is a bit too high, I think 25 or 30 would be better

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -281,7 +283,7 @@
 	if(!iscultist(user))
 		user.unEquip(src, 1)
 		user.Weaken(5)
-		user << "<span class='warning'>A powerful force shoves you away from [src]!</span>"
```
ShadowDeath6, line 284: I'd keep both these lines for feedback purposes.

ShadowDeath6:
not a big fan of nerfing wraiths as they're hardly ever used besides 'lol go kill AI then grief sec xD'

I need to rework the wall phasing mechanic sometime, it's pretty shitty

ShadowDeath6:
also I'm assuming that cultists can still use one person to sac a dead body rather than trying to convert said dead body?

code/modules/mob/living/simple_animal/constructs.dm
```diff
@@ -140,8 +140,8 @@
 	desc = "A wicked, clawed shell constructed to assassinate enemies and sow chaos behind enemy lines."
 	icon_state = "floating"
 	icon_living = "floating"
-	maxHealth = 75
-	health = 75
+	maxHealth = 40
```
ShadowDeath6, line 168: imo this should be ~55 so they can take a bit more damage rather than dying in 2 hits

code/modules/mob/living/simple_animal/constructs.dm
```diff
@@ -252,21 +253,22 @@
...
-	melee_damage_upper = 5
-	retreat_distance = 2 //AI harvesters will move in and out of combat, like wraiths, but shittier
+	maxHealth = 250
+	health = 250
+	melee_damage_lower = 25
+	melee_damage_upper = 25
+	speed = 0
 	attacktext = "prods"
 	environment_smash = 3
 	attack_sound = 'sound/weapons/tap.ogg'
```
ShadowDeath6: change this, it seems kind of shitty that they're so strong but has such a weak attack noise

code/game/gamemodes/cult/cult_items.dm
ghost: this is about the same as a single esword

code/modules/mob/living/simple_animal/constructs.dm
ghost: construct manifest spam?

code/game/gamemodes/cult/cult_items.dm
ShadowDeath6: still too much

ghost: it's literally a 'taze me on sight' sword that does the same damage as an esword, not sure how it's 2much

Time-Green:
You tested this right?

Super3222:
If the next PR you make is made using an online editor, I'm closing it.

code/game/gamemodes/cult/cult_items.dm
Super3222: you overestimate a 50% chance to block.

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -281,7 +283,7 @@
...
+		user << "<span class='cultlarge'>Get your grubby hands off me, you dirty ape.</span>"
```
Super3222: rise of the planet of the

code/game/gamemodes/cult/cult_structures.dm
```diff
@@ -173,15 +173,16 @@
...
 	var/pickedtype
 	switch(choice)
 		if("Supply Talisman")
 			pickedtype = /obj/item/weapon/paper/talisman/supply/weak
 		if("Shuttle Curse")
 			pickedtype = /obj/item/device/shuttle_curse
-		if("Veil Shifter")
+		if("Veil Walker Set")
 			pickedtype = /obj/item/device/cult_shift
+			pickedtype = /obj/item/device/flashlight/flare/culttorch
```
Super3222: This does not work that way. Instead, it will set the variable `pickedtype` to the cult shift device first, and than override it in this line with the cult torch.

code/modules/mob/living/simple_animal/constructs.dm
Super3222: since the punches are going to deal 25 brute i don't see why not. just use the juggernaught attack_sound

code/modules/mob/living/simple_animal/constructs.dm
Super3222: 40 might be tipping the boat a bit, but construct spam is a thing considered "a feature". so yeah i'd like to see it in action, but i have my doubts.

Super3222 (changes requested):
Code error

code/game/gamemodes/cult/cult_structures.dm
ghost: This is the direct code from /tg/. It should function properly.

ShadowDeath6:
Looks good :+1:

Time-Green:
You did test this right? Because this could fuck up alot of shit.

Jc1469:
sadam. can you make it so barnyard masks don't fuck up cock cult so horrendously.  A single guy with the barnyard spell from lavaland managed to make an entire clock cult unable to do anything.

ghost:
The reason that was done was to prevent people disguising their clock spells as NEIIIIIGH

code/game/gamemodes/cult/cult_structures.dm
Super3222: /tg/ puts them in a list, and spawns all of them. this puts it in a variable that's already written, and spawns that specific variable. /tg/ does it differently.

code/game/gamemodes/cult/cult_items.dm
```diff
@@ -227,7 +229,7 @@
 	body_parts_covered = HEAD
 	flags = NODROP
 	flags_inv = HIDEHAIR|HIDEFACE|HIDEEARS
-	armor = list(melee = -100, bullet = -100, laser = -100,energy = -100, bomb = -100, bio = -100, rad = -100)
+	armor = list(melee = -50, bullet = -50, laser = -100,energy = -50, bomb = -50, bio = -50, rad = -50)
```
ShadowDeath6, line 232: can we just not have a hood for it at this point

code/game/gamemodes/cult/cult_structures.dm
Cruix: I fixed

PR #1656, opened 2017-01-14
[s]Fixed an exploit where you could use boxing gloves to shoot guns if you knew sleeping carp

Cruix:


Jc1469:
really? that's pretty amazing.

PR #1683, opened 2017-01-19
Nerfed sentient ED-209s

Cruix:
Sentient ED-209s can now shoot one taser per second or two lasers per second instead of three per second, can no longer smash walls, can no longer shoot themselves, and can no longer stun and cuff if they have laser tag guns unless they are emagged.

This does not affect taser-using AI ED-209s because they will never shoot more than once per second and as far as I can tell never try to smash walls.


ToGWtF:
can you nerf them breaking walls?

McDonald072:
Why do y'all insist on removing fun

PR #1692, opened 2017-01-21
Fixed research disks magically staying in sync with the research they copied

Cruix:
Unlike designs, tech is mutable, so you cannot just copy the reference.


YogsMcree:
Make it so quantum disks stay synced, cause yknow, quantum.

PR #1697, opened 2017-01-22
Adds a wabbajack spawner

Time-Green:
everything

Creates a wabbajack statue spawner artefact.

:cl:
rscadd: It WaKeS sO yOu MaY sLeEp.
/:cl:


ghost:
kinda useless?

Time-Green:
Well, there's also the wizard spells that turns people into japanese schoolgirl neko's. So I don't see this being more useless then that

code/game/gamemodes/wizard/artefact.dm
```diff
@@ -613,3 +613,145 @@ var/global/list/multiverse = list()
...
+	icon = 'icons/obj/machines/magic_emitter.dmi'
+
+
+/obj/item/weapon/wabbajack/attack_self(mob/user)
+	var/T_loc = get_turf(src)
+	if(z == ZLEVEL_CENTCOM)
+		user << "<span class='notice'>The wabbajack does not wake here.</span>"
+		return
+	new /obj/machinery/power/emitter/energycannon/magical(T_loc)
+	var/obj/structure/table/abductor/wabbajack/left/L = new/obj/structure/table/abductor/wabbajack/left(T_loc)
```
Super3222, line 650: spawns same table

code/game/gamemodes/wizard/artefact.dm
```diff
@@ -613,3 +613,145 @@ var/global/list/multiverse = list()
...
+	L.x -= 3
+	var/obj/structure/table/abductor/wabbajack/right/R = new/obj/structure/table/abductor/wabbajack/left(T_loc)
```
Super3222: spawns same table

Super3222 (changes requested):
I went ahead, and tested it. It spawns the same two tables.

Super3222:
Also... what the heck. The concept is cool until you get to the part where it doesn't do anything at all.

Time-Green:
Uhm.. it does. Way it works is: 2 people need to get on the table and it starts shooting polymorphs. It seems I made a mistake with the tables, but that only affects the description. It works. I've tested it myself.

Super3222 (approved):

Super3222:
Requiring two people is pretty troublesome for a wizard... eh. Two living people at least.

Time-Green:
The wabbajack already exists on the special escape shuttle. And within 10 seconds there's 5-10 stacked onto those tables. I also think it MAY fire if it has one, just rarely.
It's kind of a fun thing to keep the crew busy

Super3222:
There's nothing wrong with letting it pass, but I doubt that wizards are going to use it compared to the other choices they have.

Honestly, if it was something where the wabbajack was more of a turret that activated every time a living body was forced onto their altar, than that'd make it a really useful weapon (as long as it sensed the wizard as a FRIEND). And at that point - why limit it to the a bolt of polymorph. It should fire bolts of chaos just for  amusement.

ghost:
This is useless

How are you going to get 2 people onto the tables while alive?

What's the point?

Why would anyone bother with this useless thing?

It'd be fine if it was under some sort of 'useless badassery' section but otherwise this is literally both pointless and probably going to kill the wizard trying to use it

Time-Green:
There's spells that turn people into fucking schoolgirl neko's. Is there a pointless badassery for wiz? Rituals was all I could find.

Super3222:
Like I said before, nothing wrong with letting it merge It's just a bad tool, and people aren't going to buy it after seeing it in use.

Also try adding more detail to your changelog message. No one's going to know what you're talking about.

EDIT: @Time-Green if you want to make it useful... in a way. I suggest that you go with my earlier suggestion

ghost:
@Time-Green There isn't a spell that turns someone into schoolgirl nekos

That's an admin command you dingus

Super3222:
If it was a spell, than it'd be apart of the event scroll. Which is useful.

Super3222:
Welp. Suit yourself. I've tested this, and it works.

PR #1485, opened 2016-12-26
Fixes

Cruix:
Fixed alien acid making space on lavaland and being unable to hold the CTF flags
Removed abductors automatically calling the shuttle when they complete their objective
Removed RCD's ability to make walls over lava
Added a way to resist out of abductor experiment machines (copied from https://github.com/tgstation/tgstation/pull/22280)

PR #1699, opened 2017-01-22
Traitorabd now requires 40 players instead of 1

Time-Green:
Someone should also remove traitorabdchan from the config, as it simply doesn't exist.
:cl:
tweak: Abductors dont spawn on lowpop anymore
/:cl:
I wasn't able to test this, for obvious reasons.


config/game_options.txt
```diff
@@ -106,6 +106,7 @@ PROBABILITY SANDBOX 0
 
 CONTINUOUS TRAITOR
 CONTINUOUS TRAITORCHAN
+CONTINUOUS TRAITORABD
```
Time-Green, line 109: I wasn't able to find any code related to the shuttle calling when a team has their objective complete, but I think this MIGHT have caused it. And even if not, this should still be here.
EDIT: This also HAS to be done on the host. It's just here for documenting reasons

Cruix: https://github.com/yogstation13/yogstation/pull/1485/files

Time-Green: Heh...
This should still go through tho, as the round WILL end if somehow all the abductors and traitors die.

Super3222:
The problem isn't just low pop. There's tons of other problems that pile up as well.
Honestly, it was worth a shot, but how people handle abductor roles is getting worse. It should be altogether disabled somewhere in the ACTUAL server's config files, but remain dormant in our code.

BOBsHERE:
super, its waay worse on low pop. if it doesnt get disabled im perfectly happy not seeing it late at night

Jc1469:
someone add revert/deletion tag. that's what this is.

Honestly, the gamemode just needs to fucking die. 

Traitorabd-

See obvious bad abductor, captain's ID is probably gone at roundstart because they take it instead of scanning it.
You now in your head go "but if is alien, is also teletubbies?"
Get abducted and given some meme organ that does something annoying to everyone else on the station, along with all of the traitors running around that your meta-alarm-clock is constantly beeping about.
Become gibs, become not gibs.
Get told all the bullshit is valid because it is.  It's like summon events all over again, except not as great to watch.

Super3222:
Can the next PR shut it down?

PR #1701, opened 2017-01-22
Fixes stock prices wildly skyrocketing

X-TheDark:
Fixed this on /tg/ ages ago and forgot to port it over...my bad.

|| and && have precedence issues in BYOND if you don't bracket them properly.


ShadowDeath6:
so does this totally fix the weird stock market exploit to get infinite money

X-TheDark:
It should. /tg/ haven't made any more Issues about stock prices after the fix.
This really just fixes a bug with operator precedence causing SUDDENLY BRIGHT AND SUDDEN exponential stock price increases.

Issue #1678, opened 2017-01-18
Communication Console should require being on station z-level to function.


Dagdammit:
Had a recent abductor round where scientist built a comm console on the abductor ship. He could recall the shuttle instantly with no way to prevent it. We used it and then had to immediately agree to voluntarily stop using it, it was just way too blatantly unfair.

Super3222:
The Abductors are getting smarter...
but seriously - I thought this was already a thing? Maybe it's only limited to z1 and z2? In any case, I agree with limiting it to z1.

nichlas0010:
WHY CAN'T YOU GUYS CLOSE THE ISSUES WHEN YOU FIX THINGSSSSSSSSSSSSSS

PR #1711, opened 2017-01-24
Comms consoles now only work on the station zlevel.

Jc1469:
Not technically a fix as opposed to an outright change, I guess.  but w/e

#1678

:cl:
tweak: We're shit and so are our communications consoles. They now only work on the station.
/:cl:

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -19,7 +19,7 @@ var/const/CALL_SHUTTLE_REASON_LENGTH = 12
 	var/message_cooldown = 0
 	var/ai_message_cooldown = 0
 	var/tmp_alertlevel = 0
-	var/z_lock = 1//Var that admins can change to 0 if they need to use a comms console for some crap, I dunno.
+	var/z_lock = TRUE//Var that admins can change to 0 if they need to use a comms console for some crap, I dunno, because it's basically like I didn't change it.
 	var/const/STATE_DEFAULT = 1
 	var/const/STATE_CALLSHUTTLE = 2
 	var/const/STATE_CANCELSHUTTLE = 3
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/communications.dm
```diff
@@ -50,9 +49,10 @@ var/const/CALL_SHUTTLE_REASON_LENGTH = 12
 /obj/machinery/computer/communications/Topic(href, href_list)
 	if(..())
 		return
-	if (src.z > ZLEVEL_CENTCOM) //Can only use on centcom and SS13
-		usr << "<span class='boldannounce'>Unable to establish a connection</span>: \black You're too far away from the station!"
-		return
+	if(z_lock) //Can only be used on SS13 unless z_lock is changed..
+		if(z != 1)
```
X-TheDark: `if(z != ZLEVEL_STATION)` is what you're looking for

X-TheDark (changes requested):
Please use the defines whenever possible. Not a big deal though.

code/game/machinery/computer/communications.dm
```diff
@@ -21,6 +19,7 @@ var/const/CALL_SHUTTLE_REASON_LENGTH = 12
 	var/message_cooldown = 0
 	var/ai_message_cooldown = 0
 	var/tmp_alertlevel = 0
+	var/z_lock = 1//Var that admins can change to 0 if they need to use a comms console for some crap, I dunno.
```
Super3222: if this is going to be Boolean use TRUE instead 1.

Super3222 (changes requested):
bool

X-TheDark (approved):

Super3222 (approved):

Super3222:
Looks good. I would've suggested letting emags break the z lock, but even that would bring up the same problems. In the future, we could make a centcomm version with a broken z lock and etc. Anyway it looks good.

@Cruix 

EDIT: No more memey commits or I'm closing your PR's.

Jc1469:
aylmao recall seems pretty detailed to me.  aylmaos and recalling.  it's a good commit name, it gets to the point instantly.

Super3222:
The first set of commit names aren't meant to be names. They are meant to get a clear point across to other coders in case they need to experiment, review, or debug earlier code.

Super3222:
The first set of commit names aren't meant to be memes. They're meant to get a clear point across to other coders in case they need to experiment, review, or debug earlier code.

PR #1721, opened 2017-01-26
Simple mob alien code cleanup + Couple nerfs to the queen

ghost:
Thanks for pointing out the issues, stealth. Queen 1v1ing a collosus shouldn't be possible.

Changes:

- Queen neuro is no longer hilariously deadly

- Queen no longer rapidfires, same with sentinel, whoops

- Queen is now properly slow

- Queen now summons backup when first spotting you

Code changes:

Fixes up shitcode, adds in proper var use, gets rid of a couple useless lines, fixes alien queens shrieking and spinning like tops nonstop when not supposed to

##### Changelog

:cl:
rscadd: Alien empress simplemobs have been made slightly less hilariously overpowered.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/alien.dm
+++ b/code/modules/mob/living/simple_animal/hostile/alien.dm
@@ -243,7 +243,7 @@
 			egg_cooldown = initial(egg_cooldown)
 			LayEggs()
 	if(AIStatus == AI_ON)
-		if((angry = 1) && prob(20))
+		if(angry && prob(20))
 			if(bitchslap.cast_check(0, src))
 				bitchslap.choose_targets(src)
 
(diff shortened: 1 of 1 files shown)


ghost:
whoops

ShadowDeath6:
whoops.dm

BOBsHERE:
are all xenos still op? when a normal xeno mob is nearly dead it does the rapid fire hit crit stun. while they cant kill megafauna i have seen them take them down to half health

Super3222:
I did warn you.

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -215,12 +215,12 @@
 /mob/living/simple_animal/hostile/alien/queen/large/adjustHealth(amount)
 	if(angry == 0)
 		if((health < maxHealth/2) && if(stat != DEAD))
```
X-TheDark: `if()` inside another `if()`-------------------------------^^^^^^^^^^^.

This is the actual problem that's causing issues...

PR #1722, opened 2017-01-26
Cutting off legs no longer stuns. You can handcuff one-armed people with extra time.

ghost:
Reason for leg change-Esword leg intent is bad, pls no. There's already knockdowns in the chest, no need for more free stuns. Losing a leg is bad enough. Infinite harm intent 34 damage stunbaton eshield, bad.

Yes there are counters to the esword. That's not relevant to the conversation. If you are being hit with the esword, you obviously did not employ any of those counters. It's not just an esword change too-Basically any sharp melee weapon is a lethal stunbaton and a death sentence on one hit because mister leg goes flying off.

Handcuffing one armed people is because having one arm has very very few downsides-clickdrag avoids dropping containers and only a few weapons need two hands, there's no slowdown, and as a bonus you're unhandcuffable.

Which usually means sec ends up killing one-armed people who they arrest and just can't keep down.

Now, with extra time to apply the handcuffs, you can handcuff a one-armed person's arm to their neck. This doesn't have any effect aside from fluff but at the least it allows a one-armed person to be cuffed.

##### Changelog

:cl:
rscadd: Removing legs no longer stuns.
rscadd: You can handcuff one-armed people, with some extra time.
/:cl:


ghost:
you know what with 1000 errrors

i'm gonna close this

X-TheDark:
**There's only one error, actually, the missing ending parenthesis on line 61**, the rest are just cascade that results from the parser shitting itself.

PR #1727, opened 2017-01-28
Adds colored constructs

Time-Green:
# Adds colored constructs! 
**-All constructs can now be given a color. No limits.** 
**-This color will be carried over to the new constructs/soulstones they make.
-Wizard constructs are blue now and have a wizard hud
-Chaplains shoulshard is green and so are his constructs
-All of this is also done with shades and soulstones
-Also fixed some weird bug I came along**
naamloos
(Construct in the upper left corner is the shapeshifted wizard
:cl:
rscadd: Adds colored constructs, yay.
/:cl:


Super3222:
cool

matskuman5:
Good work but the dark blue constructs look a bit hard to see

ShadowDeath6:
how does chaplain even get a soulstone

also this pr is good

Time-Green:
It's actually the metastation soulstone. Not really anyway to get it... Perhaps something for later.
And thanks, I am glad you guys like it.

GrayRachnid:
the yellow also looks a bit weird.

Super3222:
if you could include the information about the differentiation between the colors it'd be nice

ShadowDeath6:
>>>>all constructs, no limits!
>>>>>>>>>>>>>>no harvesters

close pr for lying

Time-Green:
Uhm, it uses a colored overlay. All colors that are included in color2hex can be used here. If a purple artificer makes a soulshard, it will be purple, and so will whatever construct it makes with it, etc.
About the harvester, I didn't include it since it's cult only.

The wizard gets blue soulstones, and once he makes a construct or shade with it, it'll have wiz hud and be blue. So there's a difference between cult and wiz constructs. It allowing other colors was done for fun, and is admin only.

PR #1732, opened 2017-01-29
adds modified brig

utahclock:
Per request of Super, updating/reloading a brig rework.

:cl:
rscadd: Nanotrasen engineers have made renovations to the the brig.
/:cl:


ShadowDeath6:
no windows to execution from space is a pretty big nerf to people trying to get into sec

Dagdammit:
Interesting vulnerabilites are a vital part of ss13 map design, and this would make perma a LOT less interesting. I don't think I can see a single change I'd consider positive.

Lightning38088:
why was the secure cell removed?

Jc1469:
Glass airlock on the execution room is a bad thing, if silicons are good and know how to follow laws.
Removal of the seperate room in permabrig makes it harder to slip people in that 1x1 gap they have to walk through, making escape harder, which is bad because it's already not easy.
As said above, isolation and the transfer chamber window are both gone, both being ways in.
Windows in the execution room or whatever it's called in this _update_ aren't electrified, as far as I can tell.

Nickvr628:
Ehh I don't like it. While the new execution room layout is nice, I don't like the changes to perma, and losing the isolation cell. Also, the lack of windows is a large nerf to antags trying to break in, as others have said.

ghost:
No iso cell. Perma is okay I guess. Why is there a metal sheet in the execution room? Also, needs iso window.

Ephemeralis:
Why?

Rafael450:
I think a plant bag at perma would be amazing

Rafael450:
The good plant trails are also amazing

Super3222:
Utah doesn't care about this anymore as a matter of fact. There WAS a map that she made with better modifications, so I'll probably go digging for it.

PR #1739, opened 2017-01-31
fixes queen constantnly being enraged

ghost:
apparently I forgot that check and so being hit made the queen angry every time they were hit

##### Changelog

:cl:
rscadd: More alien simplemob fixes.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/alien.dm
+++ b/code/modules/mob/living/simple_animal/hostile/alien.dm
@@ -215,12 +215,14 @@
 /mob/living/simple_animal/hostile/alien/queen/large/adjustHealth(amount)
 	. = ..()
 	if(stat != DEAD)
-		if(health < maxHealth/2)
-			angry = 1
-			visible_message("<span class='alertalien'>[src] screeches in rage!</span>")
-		else
-			angry = 0
-			visible_message("<span class='alertalien'>[src] appears to have calmed down.</span>")
+		if(angry == 0)
+			if(health < maxHealth/2)
+				angry = 1
+				visible_message("<span class='alertalien'>[src] screeches in rage!</span>")
+		elseif(angry == 1)
+			if(health > maxHealth/2)
+				angry = 0
+				visible_message("<span class='alertalien'>[src] appears to have calmed down.</span>")
 
 /mob/living/simple_animal/hostile/alien/queen/large/Aggro()
 	..()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -215,12 +215,14 @@
 /mob/living/simple_animal/hostile/alien/queen/large/adjustHealth(amount)
 	. = ..()
 	if(stat != DEAD)
-		if(health < maxHealth/2)
```
X-TheDark: This check is supposed to work, though, since it compares current queens health to half her max health...why wouldn't this be working?

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -215,12 +215,14 @@
...
-			angry = 1
-			visible_message("<span class='alertalien'>[src] screeches in rage!</span>")
-		else
-			angry = 0
-			visible_message("<span class='alertalien'>[src] appears to have calmed down.</span>")
+		if(angry == 0)
+			if(health < maxHealth/2)
```
GrayRachnid: There isn't an else for this

code/modules/mob/living/simple_animal/hostile/alien.dm
```diff
@@ -215,12 +215,14 @@
...
+				angry = 1
+				visible_message("<span class='alertalien'>[src] screeches in rage!</span>")
+		elseif(angry == 1)
+			if(health > maxHealth/2)
```
GrayRachnid: No else for this either

GrayRachnid (changes requested):

code/modules/mob/living/simple_animal/hostile/alien.dm
ShadowDeath6: he could just make this check an else

code/modules/mob/living/simple_animal/hostile/alien.dm
ghost: doesn't need to be

code/modules/mob/living/simple_animal/hostile/alien.dm
ghost: Because it fires off every time that the queen's health is adjusted leading to constant spam of 'calms down' or 'rage'

code/modules/mob/living/simple_animal/hostile/alien.dm
X-TheDark: As it stands, this will make queen aggro once and only once, since you only do this check if she's not angry. If she is, there are no checks in place to make her un-angry.

PR #1747, opened 2017-02-02
Teleport spell pulling

Time-Green:
This is about the wizard spell, not the scroll.

## -Pulling someone as wiz will take them with you in tele

This counts for anyone and anything being connected in a pull chain. They get teleported with the wiz. They end up NEAR the wiz, not next to him, usually. If you want me to change that, it's actually easier then it is now, but I like it this way.

:cl:
rscadd: Anyone connected in a pull chain will get teleported if the wiz uses his teleport SPELL
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/area_teleport.dm
+++ b/code/modules/spells/spell_types/area_teleport.dm
@@ -54,6 +54,24 @@
 		if(target && target.buckled)
 			target.buckled.unbuckle_mob(target, force=1)
 
+		var/list/plebs = list()
+		var/mob/living/P = target
+		var/obj_to_tele //In case we wanna teleport an object aswell
+		var/max_telees = 20// If it runs more then 20 times, we can assume it's stuck in a loop and we should do something about it
+		var/telees
+		while(P.pulledby)
+			plebs += P.pulledby
+			P = P.pulledby
+			telees += 1
+			if(telees >= max_telees)return
+		while(P.pulling)
+			if(isobj(P.pulling))
+				obj_to_tele = P.pulling
+				break
+			plebs += P.pulling
+			P = P.pulling
+			telees += 1
+			if(telees >= max_telees)return
 		var/list/tempL = L
 		var/attempt = null
 		var/success = 0
@@ -63,12 +81,22 @@
 			if(get_turf(target) == attempt)
 				success = 1
 				break
+
 			else
 				tempL.Remove(attempt)
+
 
 		if(!success)
 			target.loc = pick(L)
 			playsound(get_turf(user), sound2, 50,1)
+		else
+			if(obj_to_tele)
+				var/obj/O = obj_to_tele
+				O.forceMove(pick(tempL))
+			plebs -= target
+			for(var/mob/living/PLEB in plebs)
+				PLEB.forceMove(pick(tempL))
+
 
 	return
 
--- a/code/modules/spells/spell_types/area_teleport.dm
+++ b/code/modules/spells/spell_types/area_teleport.dm
@@ -56,22 +56,30 @@
 
 		var/list/plebs = list()
 		var/mob/living/P = target
-		var/obj_to_tele //In case we wanna teleport an object aswell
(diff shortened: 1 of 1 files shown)


ghost:
Interesting PR, but why?

Time-Green:
I think it gives it a fun twist.
The wiz can use this to teleport someone to a secluded area and kill them, aswell as people having a chance to hold onto the wiz and follow him.

And fuckery

ghost:
I don't think the wiz will need the ability to teleport people to secluded areas to kill them, but this does give the wizard the ability to yank someone out of danger if they're an apprentice or something.

Sure, that's cool.

ShadowDeath6:
This seems kinda OP since the wiz gets a guaranteed loot from someone he kills


But it's wiz so who cares

Super3222:
This is pretty good.

drderp3635:
Good pr to maybe boost the use of an underused ability

Ephemeralis:
Fucking spectacular. All my approves.

code/modules/spells/spell_types/area_teleport.dm
```diff
@@ -54,6 +54,24 @@
 		if(target && target.buckled)
 			target.buckled.unbuckle_mob(target, force=1)
 
+		var/list/plebs = list()
+		var/mob/living/P = target
+		var/obj_to_tele //In case we wanna teleport an object aswell
```
Cruix: This should be an obj so you do not have to cast it later

code/modules/spells/spell_types/area_teleport.dm
```diff
@@ -54,6 +54,24 @@
 		if(target && target.buckled)
 			target.buckled.unbuckle_mob(target, force=1)
 
+		var/list/plebs = list()
+		var/mob/living/P = target
+		var/obj_to_tele //In case we wanna teleport an object aswell
+		var/max_telees = 20// If it runs more then 20 times, we can assume it's stuck in a loop and we should do something about it
```
Cruix: It is reasonable to have a max number you can teleport both for balance and for good code, but that should not be your primary method of loop detection. In the first while loop below, break if you see the caster, and in the second one break if you see him or anyone in the telees list. This will prevent loops but allow small loops to be teleported in case people want to teleport while holding hands for some reason.

code/modules/spells/spell_types/area_teleport.dm
```diff
@@ -54,6 +54,24 @@
...
+		var/list/plebs = list()
+		var/mob/living/P = target
+		var/obj_to_tele //In case we wanna teleport an object aswell
+		var/max_telees = 20// If it runs more then 20 times, we can assume it's stuck in a loop and we should do something about it
+		var/telees
+		while(P.pulledby)
+			plebs += P.pulledby
+			P = P.pulledby
+			telees += 1
+			if(telees >= max_telees)return
```
Cruix: Give the wizard a message here and below telling him why the cast failed, and use revert_cast()

code/modules/spells/spell_types/area_teleport.dm
```diff
@@ -54,6 +54,32 @@
 		if(target && target.buckled)
 			target.buckled.unbuckle_mob(target, force=1)
 
+		var/list/plebs = list()
+		var/mob/living/P = target
+		var/obj/obj_to_tele //In case we wanna teleport an object aswell
+		var/max_telees = 20// If it runs more then 20 times, we can assume it's stuck in a loop and we should do something about it
+		var/telees
+		while(P.pulledby)
+			if(P.pulledby in plebs)break
```
Cruix, line 63: These should also break if they find the wizard, otherwise he could get placed in the plebs list and get teleported twice.

Time-Green: I am currently in the middle of nowhere, but I'll fix it once I can get some wifi
EDIT: This is already prevented at line 103, where the wiz gets removed from the list. (target being the wiz)

PR #1749, opened 2017-02-02
Adds surgery tips

Time-Green:
### Medicals, Roboticists, Abductors and the CMO will now be told wich step to take next if they're doing surgery. 

naamloos
(It's now in normal size and italics, to lazy to update the image)

:cl:
rscadd: All surgery related jobs will now be instructed wich steps to take
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_step.dm
+++ b/code/modules/surgery/surgery_step.dm
@@ -86,6 +86,6 @@
 		if(surgery.get_surgery_step())                                   													//Give jobs expertises
 			var/datum/surgery_step/step = surgery.get_surgery_step()
 			step = step.name
-			user << "<i>Now to [step] on their [target_zone]</i>"
+			user << "<i>Now to [step] on their [target_zone].</i>"
 			return 1
 	return 0
--- a/code/modules/surgery/surgery_step.dm
+++ b/code/modules/surgery/surgery_step.dm
@@ -6,7 +6,6 @@
 	var/time = 10					//how long does the step take?
 	var/name
 
-
 /datum/surgery_step/proc/try_op(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	var/success = 0
 	if(accept_hand)
@@ -55,6 +54,7 @@
 		if(prob(prob_chance) || isrobot(user))
 			if(success(user, target, target_zone, tool, surgery))
 				advance = 1
+				addtimer(src, "guide", 5, FALSE, user, target, target_zone, tool, surgery)
 		else
 			if(failure(user, target, target_zone, tool, surgery))
 				advance = 1
@@ -65,7 +65,6 @@
 				surgery.complete()
 
 	surgery.step_in_progress = 0
-
 
 /datum/surgery_step/proc/preop(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
 	user.visible_message("[user] begins to perform surgery on [target].", "<span class='notice'>You begin to perform surgery on [target]...</span>")
@@ -81,3 +80,12 @@
 
 /datum/surgery_step/proc/tool_check(mob/user, obj/item/tool)
 	return 1
+
+/datum/surgery_step/proc/guide(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	if(user.job == "Medical Doctor" || user.job == "Chief Medical Officer" || user.job == "Roboticist" || isabductor(user)) //TODO:
+		if(surgery.get_surgery_step())                                   													//Give jobs expertises
+			var/datum/surgery_step/step = surgery.get_surgery_step()
+			step = step.name
+			user << "<span class='boldnotice'>Now to [step] on their [target_zone]</span>"
+			return 1
+	return 0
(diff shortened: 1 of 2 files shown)


Super3222:
the beginning of the knowledge system

matskuman5:
^

BOBsHERE:
Ah man. So torn. I really like the idea of not being screwed when i lose a limb because there's no competent medical staff, but i also like the idea of people having to actually learn their rolls without having their hands held. I mean, the surgical computer already shows the next step so i guess its not a big deal to add it into the chat bar.

Super3222:
I understand how you feel puremanbird. Honestly, I feel the same. I'm awfully torn about the idea; as it opens up a decent feature it's also kind of spoonfeeding. What you said about the surgery computer is also true.

Jc1469:
maybe we should make examining people show every possible reagent and how to make them to fix their injuries.  

maybe we should make a turret that shoots manuals at people, because manuals fucking exist, and the surgical computer exists.

God no.  This PR is stupid.

Dagdammit:
I was against this until I noticed it was only for doctors, CMO, roboticists and abductors. In this specific case, that's actually fitting enough that it turned me around. We can still evaluate any other potential implementation of a knowledge system on a case-by-case basis; I can't think of many cases I would support, with the possible exception of engineers knowing some wire functions.

You could always have it only function while the surgical table's adjacent to a working surgical computer. (Abductors can get the benefit in general or while aboard their UFO.)

I really think we should give this a try.

Time-Green:
The reason I made it was because, they're supposed to know it. They're freaking qualified surgeons. They're supposed to know it.

Time-Green:
But yeah, I know it may be to much spoonfeeding, but they are supposed to know it. And non-surgery peeps wont get it, so at least surgeons have a build in advantage.

And yes the knowledge system is coming at some point.

ghost:
Put a period at the end of the sentences, please. And make it the same font type.

Also, you could just make it a clue instead of spoonfeeding. They have a surgery observation computer-They should use that if they want the direct step told to them.

'They are supposed to know it' isn't a reason to spoonfeed people things like this. Ss13 is intentionally obscure and there are lots of hidden things new players don't know. We're starting off the slippery slope here.

Am against PR for above reasons.

ShadowDeath6:
Have it be italic and regular font color, no spans. This is a cool pr, though.

Also seriously adam this isn't lifeweb we don't need to hide everything behind a wiki and trial and error, at least it only tells you when you're a doctor

Dagdammit:
I am still in favor of this PR, but I'm not on board with giving a broad green light to any kind of knowledge system. Time-Green, in ss13 "they're (jobname), they should know how to do X" generally means the PLAYER has an OBLIGATION to learn how to do X before they select that job. 

I like this instance because I feel that giving those players that info will enhance the overall game experience, because surgery is underutilized and seeing a surgeon prod someone with surgery tools until one turns out to be correct is just dumb. But I'd favor a version where the tips only appear when a working surgery computer is on hand, or at least in the same room.

Time-Green:
I'll fix the text somewhere tomorrow, probably.

ShadowDeath6:
cool

Alek2ander:
There's a surgery computer for a reason. Please no job-locked features.

Time-Green:
I think medicals should have a surgery advantage over a random assistant.
And indeed, there's already a surgery computer. So how much does it really affect it?
Not that much, it just speeds things up and gives medicals an advantage.

ghost:
If there's a surgery computer, no need to make things easier than they already are.

Pointless feature bloat.

Dagdammit:
I'm hardly gonna say surgery should be easier- but prodding/stabbing people with implements until one works just feels like the wrong *kind* of hard. It's not like you'd be staring at a ribcage IRL and think "do I use the retractor for this?" 

I'd prefer it require a working surgery computer in the room, but regardless I'd vote merge this PR and make surgery a bit more challenging via a mechanic like https://github.com/yogstation13/yogstation/issues/1769 instead.

Super3222:
alternatively make the operating computer BLEEP out the next step during surgery.

Keekenox:
Not more of this job specific knowledge based hoo hah, we get it, you don't like powergame.Stop trying to stop powergame with job specific IC knowledge based mechanics. Stahp. This is not a step in the right direction, it's a step in the WRONG direction.

code/modules/surgery/surgery_step.dm
```diff
@@ -81,3 +80,12 @@
 
 /datum/surgery_step/proc/tool_check(mob/user, obj/item/tool)
 	return 1
+
+/datum/surgery_step/proc/guide(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery)
+	if(user.job == "Medical Doctor" || user.job == "Chief Medical Officer" || user.job == "Roboticist" || isabductor(user)) //TODO:
+		if(surgery.get_surgery_step())                                   													//Give jobs expertises
```
Cruix: This causes a list index out of bounds error on the last surgery step

Time-Green:
I fixed it. I checked the original proc and it wont ever return 0.
This worked perfectly in-game, so I think this does it.

matskuman5:
>alternatively make the operating computer BLEEP out the next step during surgery.

This. So much this.

Dagdammit:
Yeah I also like Super3222's take.

Time-Green:
The idea is to give medicals a direct boost above some random assistant.
I could have it bleep the steps, but then everyone would have benefit to it.

Dagdammit:
Have it only do so if the person performs a step while wearing an ID with
the right access.

Super3222:
I think things have died down from the age where the knowledge system was more considered. Making the surgery computer audibly explain / hint the next step is probably a better step.

Time-Green:
I'd rather close it, if that's the general opinion

Time-Green:
J

PR #1775, opened 2017-02-06
Visible Examines

Super3222:

Bare with me here -
This feature will make examining objects or other people noticeable.

Baystation, I believe, has done something like this as well. And to be honest with you - it makes a really nice effect. I am willing to at least **experiment** with it, here on Yogs.

1. Examining an object you're holding/in your inventory / on you **IS NOT VISIBLE**.
2. Examining yourself **IS NOT VISIBLE**.
3. Examining other people can only be seen by people **ON AN ADJACENT TILE.**

If you still believe that the smallest font size is going to bother you then you'll be happy to know that I've included preference options.
- The first one lets you see visible examines all the time.
- The second one lets you see visible examines every 6 seconds. After an examine pops up, it'll take another 6 seconds until you can receive another one.
- The third one completely shuts them down. Your examines will still be noticable to other people who have this setting on.

##### Changelog

:cl:
experimental: Examining objects or people is noticeable.
experimental: Visible Examines also have preference options for spam reduction or completely blocking them.
/:cl:


Lightning38088:
maybe soon we can add the feature where removing a weapon out of your bag make its show "x pulls out y" or "x conceals y"

Super3222:
That's too far. Weapons are also extremely visible.

ShadowDeath6:
is this lifeweb

also on lifeweb it gets immensely spammy so uhhhhhh that's gonna be a no from me

McDonald072:
God no, I constantly examine things, this would get really spammy as shadow says

Jc1469:
Encourages NOT examining people because you might tip people off that you're about to try something.
:-1: 
On top of what Shadow said.

Dagdammit:
Seems neat. I wanna try it out for a bit and see.

Time-Green:
I like the idea, but I am kinda worried about spam

matskuman5:
^

Super3222:
Yeah, I just want to try it out for awhile sometime and see what happens. It's probably not going to be permanent, but it'll be fun to watch / experiment.

EDIT: Furthermore, we can make looking at people only noticeable from people within 1 tile of you (specifically adjacent - has to be in a tile next to yours).

Slinky5890:
Totally fine with a test run

Cruix:
Maybe also add a click cooldown on examine

ShadowDeath6:
>add a cooldown for an informational command

no thanks I think I just ded

Dagdammit:
No it makes sense. If you have no cooldown, I can hold shift and click to
spam with incredible ease. If you make the cooldown for notifications I can
just examine one thing and then another to avoid someone seeing. A
one-second cooldown on examination is fine, takes at least that long to
properly look something over anyway.

Super3222:
A cooldown for notifications would work

BOBsHERE:
god no. The spam. Also im not the only one who just spams shift click on things.

Dagdammit:
It is tricky because sometimes you walk in and someone in a hardsuit
guiltily flees the scene, giving you a limited window to click them for
examination. Any kind of actual cooldown on examining things would
definitely alter how that situation plays out- though I'm not saying the
result would be a worse game experience, maybe making it harder to get a
good look at someone who isn't holding still would actually be a minor
improvement.

Also the potential for INTENTIONAL spam would not be increased by this
pull, because you can already spam just as quickly with the hotkey to point.

Still think we should try this. There could be unappreciated upsides that
outweigh the potential downsides we can spot in advance. It's one of those
things where experimentation is what gives you the real answer.

ShadowDeath6:
Not being able to get all the information you need is never an improvement, especially important info like finding out if they have a .357 on their belt

Super3222:
I'm not crazy enough to try a cooldown on examines. Interesting appeal like always Dag, but it takes away too much from the game. I was also thinking about making examines that happen behind heavy masks such as hardsuit or space helmets unnoticeable... maybe that could happen?

I'll find time to make a cooldown on notifications.

Dagdammit:
Could do it for masks, eyewear and helmets that mask your identity and/or
gives some measure of flash protection. Latter's always gonna be concealing
your eyes behind some kind of tinted lens, after all.

Notamaniac:
Or you could try the best of both worlds, let the person examine however much they want, but only show the "x looks at y" every one or two seconds

Dagdammit:
That's the option people favor and what super says he'll go with.

ShadowDeath6:
why not have it just be a prob (whatever) to display so it's not as spammy and can't be """"""""exploited"""""""" by examining something else for the message

Dagdammit:
Not a bad idea. High probability if first examine in a while, else
something like 20%?

ShadowDeath6:
why

Super3222:
Are you REALLY asking why

ShadowDeath6:
yes

Super3222:
@Dagdammit The thing about random chances is that it encourages you not to take it. I've got a better solution in mind.

Super3222:
So basically examining other people can only be seen by other people standing on an adjacent tile. Hopefully fixes the problem with looking suspicious in front of a couple hundred people.

If spams a problem then I can make a preference option to disable / reduce (cooldown) / enable examine messages.

ShadowDeath6:
make it start disabled

code/game/atoms.dm
```diff
@@ -168,6 +168,10 @@
 
 
 /atom/proc/examine(mob/user)
+	if(loc != user) // so inventories don't get spotted
+		for(var/mob/M in view(7, user))
```
ShadowDeath6: ?

Super3222: Those are objects, not people.

ShadowDeath6: Don't even have messages for that, this is the entire reason that examining is spammy. In my experience, I'm almost always examining things around the room, and multiply that by 3 or 4 considering in large areas there are always more than that.

Super3222: As I've told you this entire thing is experiment. It's an actual feature that players from another server have come to enjoy and deal with. People examine tons of things on those servers, and yet it's not a problem.  I just want to try it out to see how it goes. That's the entire point of an experimental PR, Shadow.
I'm also working on a preference for players that DON'T want to see what other people are examining.

The text is also **SUPER** small so I doubt it's going to be as much of a problem as you're making it.

ShadowDeath6: why would anyone care what doors I look at when they're 6 tiles away from me

ShadowDeath6: I hope experimental implies that this will be testmerged rather than fully merged as I can guarantee nobody is going to make a revert pr that will pass if they don't like it

Keekenox:
Fuck no, you have idea how much I examine spam things , especially when I am stalking a murder target, fuuuuck that.

Dagdammit:
All we can do is make uninformed guesses or actually test the darn thing. So let's test it already. A few days with it on won't kill nobody.

code/game/atoms.dm
Super3222: You can keep complaining about it, but if the community doesn't the entirety of the idea then it'll be reverted without a doubt.

Super3222:
PREFERENCE OPTIONS INCOMING;

- The first one lets you see other people examining stuff with range restrictions
- The second one will set each examine you see 6 seconds apart. So after noticing one examine your character won't notice any more until those 6 seconds are up.
- The third one makes you blind to other people examining stuff. That's it. Your examines will still be noticeable by other people however.

ShadowDeath6:
isobserver is a helper flag

Super3222:
there are other (abstract) mob types that i don't want picked up.

ShadowDeath6:
`#define isobserver(A) (istype(A, /mob/dead/observer))`

Super3222:
check `Update atoms.dm`

ShadowDeath6:
why would you make it like that though

Super3222:
> there are other (abstract) mob types that i don't want picked up.

also can you tell a council member to update or take down this PR from the test merge.

ShadowDeath6:
what (abstract) mobs are there tho

Jc1469:
Stuff like blob eyes and gods are mob/camera.   Ghosts are /dead.   So really you could probably get away with just doing a living check, both.. for mob/living and seeing if they're alive as a mob/living.

ShadowDeath6:
this pr is too buggy pls close and rethink your decisions >:^(

Super3222:
When I get back on my PC I'll make it isliving().

ShadowDeath6:
revenants are mob/living too

code/game/atoms.dm
```diff
@@ -172,9 +172,8 @@
 
 /atom/proc/examine(mob/user)
 	if(loc != user) // so inventories don't get spotted
-		if(ishuman(user) || ismonkey(user) || isrobot(user))
-			for(var/mob/M in view(7, user))
-				M << "<span class='small'>[user] looks at [src].</span>"
+		if(isliving(user))
```
ShadowDeath6: Revenants are living

Mrcreepssss:
maybe you should make it so it only messages you the first time you see someone examine a specific object but when they examine another it does notify? you could put this on a X second cooldown before it notifies you when they examine object 1 again.

PR #1781, opened 2017-02-08
Added antag_weight to player table in feedback database

tomamoto:
Note that this column was added as a float to stay on the safe side based on observing antag_weight being divided by other floating point numbers in the code.  Looking further through the code, I did notice this value is parsed using txt2num, so maybe I should be storing it as a string/varchar (as far as the code says, anyway). Or does the a database query in this case only return an array of strings? Please pardon my ignorance of DM. 

All I know is that adding this column to by database fixed a nasty bug on my server where antags were never being set while the database was enabled.  If you don't want to merge this as is, I would understand, but then please suggest and implement a solution using a more appropriate MySQL data type.

### DIFF
--- a/SQL/database_changelog.txt
+++ b/SQL/database_changelog.txt
@@ -1,3 +1,13 @@
+
+7 February 2017, by tomamoto
+
+Modified table 'player', adding column 'antag_weight', to bring it inline with the schema expected by the yogstation code.
+
+ALTER TABLE `feedback`.`player` ADD COLUMN `antag_weight` FLOAT DEFAULT NULL
+
+Remember to add a prefix to the table name if you use them
+
+----------------------------------------------------
 21 September 2015, by Jordie0608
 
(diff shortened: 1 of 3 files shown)


tomamoto:
Sorry for all the typographical errors.  Let me try this once more:

Note that this column was added as a float to stay on the safe side based on observing antag_weight being divided by other float point numbers in the code.  Looking further through the code, I did notice this value is parsed using txt2num, so maybe I should be storing it as a string/varchar (as far as the code says, anyway). Or does the a database query in this case only returns arrays of strings? Please pardon my ignorance of DM. 

All I know is that adding this column to by database fixed a nasty bug on my server where antags were never being set while the database was enabled.  If you don't want to merge this as is, I would understand, but then please suggest and implement a solution using a more appropriate MySQL data type.

tomamoto:
Fixed a bunch of typographical issues in my pull request.  Sorry about that!

ShadowDeath6:
@oisin100

oisin100:
Good catch.

Issue #1808, opened 2017-02-12
Make Fire extinguisher extinguish welder?


00aquir:
It would make sense, since it's fire, especially when somebody throws one into plasmafire, the extinguisher should take care of that little flame too. Or when someone is assaulting with welder, one would be able to defend themselves by extinguishing the welder in attacker's hand.

Super3222:
Sounds cool. But to get a decent weapon seems like I'll have to break in and hack the chefs kitchen vendor

Dagdammit:
Fire extinguisher working on dropped ignition sources like welders and flares could be cool. I don't care one way or another from a feasibility standpoint, considering both function in the vacuum of space. Putting it out when it's in the hands of someone attacking you feels like a real minor concern.

...though what WOULD be significant is if extinguishers shut off the pilot light on a held flamerthrower.

Drakkon:
*active flaming items burn the container they're in if it's flammable*
One part realism, one part hilarity.

Dagdammit:
I'll admit, I've gone "hmm" every time I stow a lit molotov in a backpack
and/or box.

PR #1811, opened 2017-02-12
Fire extinguishers can extinguish welders

Super3222:

it's not meant to "nerf" welders or anything. it's meant to be fun.
go to here -> https://github.com/yogstation13/yogstation/issues/1808
also works if the welders on the floor or a table.
i could make it an if prob chance if it's a problem.

##### Changelog

:cl:
rscadd: Fire extinguishers can extinguish welders.
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/effects_water.dm
+++ b/code/game/objects/effects/effect_system/effects_water.dm
@@ -23,6 +23,17 @@
 /obj/effect/particle_effect/water/Bump(atom/A)
 	if(reagents)
 		reagents.reaction(A)
+	if(istype(A, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = A
+		var/obj/item/weapon/weldingtool/W
+		if(istype(H.r_hand, W))
+			W = H.r_hand
+			W.toggle()
+		// not else if in case they are dual-wielding
+		if(istype(H.l_hand, W))
+			W = H.l_hand
+			W.toggle()
+
 	return ..()
 
 
--- a/code/game/objects/effects/effect_system/effects_water.dm
+++ b/code/game/objects/effects/effect_system/effects_water.dm
@@ -30,23 +30,14 @@
 			W = H.r_hand
 			if(W.welding)
 				W.toggle()
-				visible_message("[W] extinguishes.")
+				visible_message("[H]'s [W] extinguishes.")
 		// not "else if" in case they are dual-wielding
 		if(istype(H.l_hand, W))
 			W = H.l_hand
 			if(W.welding)
 				W.toggle()
-				visible_message("[W] extinguishes.")
+				visible_message("[H]'s [W] extinguishes.")
 	return ..()
-
-/obj/effect/particle_effect/water/Crossed(atom/A)
-	if(istype(A, /obj/item/weapon/weldingtool))
-		var/obj/item/weapon/weldingtool/welding = A
-		if(welding.welding)
-			welding.toggle()
-			visible_message("[welding] extinguishes.")
-	return ..()
-
 
 /////////////////////////////////////////////
 // GENERIC STEAM SPREAD SYSTEM
--- a/code/game/objects/effects/effect_system/effects_water.dm
+++ b/code/game/objects/effects/effect_system/effects_water.dm
@@ -28,15 +28,23 @@
 		var/obj/item/weapon/weldingtool/W
 		if(istype(H.r_hand, W))
 			W = H.r_hand
-			W.toggle()
-		// not else if in case they are dual-wielding
+			if(W.welding)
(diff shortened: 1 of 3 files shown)


code/game/objects/effects/effect_system/effects_water.dm
```diff
@@ -23,6 +23,20 @@
 /obj/effect/particle_effect/water/Bump(atom/A)
 	if(reagents)
 		reagents.reaction(A)
+	if(istype(A, /mob/living/carbon/human))
+		var/mob/living/carbon/human/H = A
+		var/obj/item/weapon/weldingtool/W
+		if(istype(H.r_hand, W))
+			W = H.r_hand
+			W.toggle()
```
Cruix: You should probably check to see if it is on first

Cruix:
I am pretty sure that Bump() only works on dense objects, so this will not affect welders on the floor.

ShadowDeath6:
wtf nerfs welder attacking

code/game/objects/effects/effect_system/effects_water.dm
```diff
@@ -28,15 +28,23 @@
...
+		// not "else if" in case they are dual-wielding
 		if(istype(H.l_hand, W))
 			W = H.l_hand
-			W.toggle()
+			if(W.welding)
+				W.toggle()
+				visible_message("[W] extinguishes.")
+	return ..()
+
+/obj/effect/particle_effect/water/Crossed(atom/A)
```
Cruix: Crossed() is called when the thing that is moving crosses you, this should be in water's reaction_turf() or the welding tool's Crossed()

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -149,7 +149,7 @@
 		var/obj/item/weapon/weldingtool/W = O
 		if(W.welding)
 			W.toggle()
-			visible_message("[W] extinguishes.")
+			O.visible_message("[src] extinguishes [W]'s flame!.")
```
ShadowDeath6: what is [src] actually shown as when it extinguishes

Super3222: Water...? I'm hoping.

Cruix: ...did you test this?

Super3222: Just tested it, and it works. I just need to add spans to every visible_message()

EDIT: Hold off on merging. I'm fixing some things.

Time-Green:
Finally, I found a way to avoid the phytosian-welder meta

Satile1:
This seems so minor that it isn't worth the extra bump on water. 

Tis cool, though. Would also be good if the extinguished-but-still-on welder sprayed its fuel on the floor, OR whoever it was used to attack..

code/modules/reagents/chemistry/reagents/other_reagents.dm
Super3222: Water.

Dagdammit:
Feature creep options:
-Have them put out lit flares, torches, maybe candles.
-Have them put out the pilot lights on flamethrowers.

Keekenox:
I like this, this adds consistency, which is SS13s very greatest asset.

Nickvr628:
Wow man GG now the water reagent blob will be stupidly OP.

How are we supposed to fight it now?

Dagdammit:
Nick: Oh hey that's an interesting point. It'll just mean toggling welder
back on or using anything else, but still. I like it.

ShadowDeath6:
only applies to the particle effect which the blob doesn't use

code/game/objects/effects/effect_system/effects_water.dm
```diff
@@ -23,9 +23,22 @@
...
+			W = H.r_hand
+			if(W.welding)
+				W.toggle()
+				visible_message("[H]'s [W] extinguishes.")
+		// not "else if" in case they are dual-wielding
+		if(istype(H.l_hand, W))
+			W = H.l_hand
+			if(W.welding)
+				W.toggle()
+				visible_message("[H]'s [W] extinguishes.")
```
Cruix, line 39: Bump does not actually work for this because the water effect itself goes through humans, you will have to put this somewhere else.

Nickvr628:
YOU FOOLS there is no water blob!

I think.

Maybe.

Either you guys are gullible or I am good at guessing.

ShadowDeath6:
there is indeed a pressurized slime blob that puts water on the ground when one of its tiles dies

Cruix:
Re-open this when it is finished.

Issue #1821, opened 2017-02-13
Recalling then instantly calling the shuttle on shadowling


puremanbird:
Red alert, shuttle gets called. Lings sacrifice a thrall to delay the shuttle by fifteen minutes. recalls then instantly calls the shuttle. this is broken. Make the shuttle unrecoverable once you sacrifice a thrall to delay it.

PR #1827, opened 2017-02-14
[s]Fixes an exploit with recalling and destroy engines

Time-Green:
There was an exploit wich keek would use and instantly recall and call again to nullify destroy engines.

This prevents recalling after destroy engines has been used. Suggested here: https://github.com/yogstation13/yogstation/issues/1821

:cl:
fix: Fixes recalling after destroy engines
/:cl:

### DIFF
--- a/code/controllers/subsystem/shuttles.dm
+++ b/code/controllers/subsystem/shuttles.dm
@@ -210,9 +210,6 @@ var/datum/subsystem/shuttle/SSshuttle
 			if(!ignoreCD)
 				return 3
 
-
-/datum/subsystem/shuttle/proc/moveShuttle(shuttleId, dockId, timed)
-	var/obj/docking_port/mobile/M = getShuttle(shuttleId)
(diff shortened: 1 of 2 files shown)


1fbff5f83b23d39d38b1dfcb4cac8d9b:
rename the variable or proc to something else, I think they are conflicting

Time-Green:
That aint it, it worked fine on my local, I honestly didnt even know it broke before you commented

Time-Green:
Works now, for some reason weird reason their was new code I never even touched

ShadowDeath6:
Is there a way to make it unable to recall until the added time is removed so shadowlings can cheese the shuttle

Lightning38088:
is it me or does keek always use exploits to his advantage when possible?

Keekenox:
I'm going to be completely honest here, I didn't even realise it was an exploit until everyone started yelling at me, hm you can do the same thing when cult lengthens the shuttle, it just seemed logical to recall and call, kinda like when you  recall, then call red alert, then call again.

Issue #1838, opened 2017-02-16
Allow library computer to print books much more efficiently


ArgyleSweatshop:
The most tedious part of playing librarian is stocking the shelves. It'd be nice to not have to spam click the computer. The librarian menu needs to have specific add to queue buttons that don't cause its menu to reset- it is currently extremely tiring simply to print out 30-40 good books.

Dagdammit:
Oh, you want a queue function like the exosuit fabricator. That would make a ton of sense.

PR #1851, opened 2017-02-17
Print All Library Computer Function

ArgyleSweatshop:
Refer to issue #1838 

##### Changelog

-Massively increases print queue size
-adds a simple print all function. Significantly reduces tedium of playing librarian.

### DIFF
--- a/code/modules/library/lib_machines.dm
+++ b/code/modules/library/lib_machines.dm
@@ -178,7 +178,7 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	var/page = 1	//current page of the external archives
 	var/print_busy = 0 // LOL NO SPAM (1 minute delay) -- Doohl
 	var/list/print_queue = list()
-	var/max_print_queue_len = 50;
+	var/max_print_queue_len = 1000;
 	var/clearprintqueue = 0;
 
 /obj/machinery/computer/libraryconsole/bookmanagement/proc/build_library_menu()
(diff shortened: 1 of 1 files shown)


code/modules/library/lib_machines.dm
```diff
@@ -178,7 +178,7 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	var/page = 1	//current page of the external archives
 	var/print_busy = 0 // LOL NO SPAM (1 minute delay) -- Doohl
 	var/list/print_queue = list()
-	var/max_print_queue_len = 50;
+	var/max_print_queue_len = 1000;
```
Time-Green, line 181: This is WAY to high. I've spammed the station with books and it was HARD to hit that maximum. 1000 is a little to high. Maybe 100 as max?

ArgyleSweatshop: If you don't set it that high, the queue wouldn't have room to print all the books.

Time-Green: I understand, but printing all the books is insane. I had no idea you actually meant it to print ALL the books. This will just get spammed, and one does not need over 1000 books

ShadowDeath6:
creating 1000 new objects at once will probably lag the shit out of the server

ArgyleSweatshop:
It doesn't create it all at once! It adds to the queue, which itself has a decent delay. There would be minimal lag!

ShadowDeath6:
1000 new objects is still no joke, I don't think there's anything else that comes close to it other than guthens mass reproducing in nullspace

Dagdammit:
Look performance concerns aside, is it gonna ruin anything if that particular variable is 100 instead?

Also I'm sorry, is this literally just printing all the books in the external archives, rather than ones the librarian has somehow selected? Because one of the current problems with the library is that the archive is full of garbage. Way too many books that are literally just one-line "hello world" equivalents or 1-2 paragraphs of minimal information. Tons of duplicates, too. And no way to filter good stuff from the dreck, and.....

It really seems like exo fabricator's buttons would be the right way to go here. You either choose "print" or "add to queue", and can remove individual queue items, remove all queue items, or print the queue. If you wanna print X copies of Baits for Clicks, hit "add to queue" X times.

ArgyleSweatshop:
The concerns are simply too valid to leave this open.

Someone else needs to tackle an overhaul of the library menu, then. It currently is a mess, and hitting print causes the menu to reset, making doing the librarian job have excessive tedium.

Dagdammit:
Agreed that fixing librarian job will require an overhaul of the library catalog interface. That's a core problem you can't paper over with other neat mechanics.

Issue #1846, opened 2017-02-17
Shouldn't Arrivals shuttle be protected from Rad storms?


00aquir:
Somebody joins, only to find out they are getting retarded, blind / deaf. This kind of a thing happened to  many people... this area should be an exception.

Dagdammit:
It should yeah.

doerakpoes:
Probably

Gintoki1:
The escape shuttle isn't protected from radstorms. I've just witnessed it today, and there was a good amount of puke

Dagdammit:
I'd vote protect escape shuttle too. There's no direct maint connection to it, so radstorms still inhibit crew reaching shuttle in interesting ways.

PR #1861, opened 2017-02-19
Protects arrivals and escape shuttle from radstorms

Time-Green:
Protects the arrivals and escape pods from radiation storms.
Also protects pods.
Fixes: https://github.com/yogstation13/yogstation/issues/1846
:cl:
tweak: Arrivals and escape shutte are protected from radiation storms.
/:cl:


Super3222:
You beat me to it!

PR #1862, opened 2017-02-19
Disables lavaland atmos

Time-Green:
Disables atmos from updating on lavaland.
Hopefully cuts down on the lag.

I failed to actually notice a difference gameplay wise, so that's good. 
:cl:
experiment: Disables lavaland atmos
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/lavaland_areas.dm
+++ b/code/modules/mining/lavaland/lavaland_areas.dm
@@ -27,6 +27,7 @@
 /area/lavaland/surface/outdoors
 	name = "Lavaland Wastes"
 	outdoors = 1
+	atmos = 0
 
 /area/lavaland/surface/outdoors/mapgen_protected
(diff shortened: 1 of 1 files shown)


oisin100:
Its mainly the mobs causing any sort of lag.

Time-Green:
Meh

Super3222:
Uh... wow.

Time-Green:
I tested it, and it doesn't change anything. It just makes the tiles not constantly check for updates, the gameplay is the same.

Jc1469:
balance change detected.  miners can no longer be pulled out of mining by wind of space.

>  it doesn't change anything.

Time-Green:
Spacewind is still there. For some reason. I don't wanna think about it

ShadowDeath6:
If pressure is still there, then there's still a difference between mining and lavaland, so it'll probably still pull you around.

PR #1863, opened 2017-02-19
Fixes binoculars

Super3222:

- Having to use the action button instead of activating it in hand (though both options are still viable)
- Decreasing scope range not working.
- Small grammar error

#### Changelog

:cl:
bugfix: Binoculars work properly now.
/:cl:


ShadowDeath6:
why would you ever want to use an action button on an item you can only use inhand

Super3222:
this pr FIXES that.

ShadowDeath6:
oh

PR #1879, opened 2017-02-21
horrorform removes mutations now

ShadowDeath6:
you can't stack chameleon and shit onto horrorform dummy

also fixes strained muscles being comboed for speedy abominations


Super3222:
Horror form + Chameleon is modern day combo cafe. We need a bar sign for it.

PR #1881, opened 2017-02-21
Gives ocular wardens a gaze beam

Time-Green:
Requested by: @AdamElTablawy 
Sprite by: @Kmc2000 
:cl:
rscadd: Ratvar has given his ocular wardens laser eyes, which scare the heartbeating crap out of their victims
/:cl:

schermopname 406


ghost:
that looks like crap

ghost:
also you spelled 'wich' wrong and changelogs like that are bad

Super3222:
Looks good. You should talk to KMC about making a "blur" effect around the line to create that intensity feeling.

But anyway - you don't really need the sound tag because that implies a new sound is being added to the repo.

Jc1469:
So you're not only giving a visual indication of a warden targeting you, but also an audio one..? 

Seems a bit un-necessary to have both to me.

Super3222:
I'm not sure what's unnecessary about this. In any case it looks good to go. But as required we need that other PR to go.

Time-Green:
I set alpafade to 1 in the other PR, so this can safely be merged.

PR #1896, opened 2017-02-25
You can now transfer AI's into potatos

Super3222:
when game references go 2 far

:cl: 
rscadd: You can now transfer AI's into potato's. It must be from an intellicard however.
/:cl:

### DIFF
--- a/code/game/gamemodes/objective_items.dm
+++ b/code/game/gamemodes/objective_items.dm
@@ -103,10 +103,19 @@
 	targetitem = /obj/item/device/aicard
 	difficulty = 20 //beyond the impossible
 
-/datum/objective_item/steal/functionalai/check_special_completion(obj/item/device/aicard/C)
-	for(var/mob/living/silicon/ai/A in C)
-		if(istype(A, /mob/living/silicon/ai) && A.stat != 2) //See if any AI's are alive inside that card.
-			return 1
+/datum/objective_item/steal/functionalai/check_special_completion(obj/item/I)
+	var/obj/item/device/aicard/C
+
+	if(istype(I, /obj/item/weapon/stock_parts/cell/potato))
+		var/obj/item/weapon/stock_parts/cell/potato/P = I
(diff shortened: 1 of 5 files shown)


ShadowDeath6:
can I tactically shunt into a potato to be invincible

Super3222:
Well.. You need to hit a potato cell with an intellicard to do this. You can only transfer it back to another intellicard thereafter.

Cruix:
Put an  AI_TRANS_TO_CARD interaction in the potatoe instead of adding a var to the AI mob's transfer_ai

Also update the steal an AI objective

Super3222:
> Put an AI_TRANS_TO_CARD interaction in the potatoe instead of adding a var to the AI mob's transfer_ai

That'd require a lot of copy and paste from the AI section. I thought adding a variable would be a better solution?

Cruix:
The whole reason the transfer_ai() proc exists is so you don't have to put a hidden intellicard in anything that can hold an AI, like mechs or the ai fixing computer.

Super3222:
Fair enough. I'll get around it eventually.

ShadowDeath6:
You spelled "too" wrong in the OP, close this

doerakpoes:
Portal 2 refrences

Cruix:
It now always deletes the inactive AI core when transferring to an intellicard, and there is no way to transfer the AI out of the potato.

There should also be a special text on examine if there is an AI in the potato.

Cruix:
Re-open this when it is fixed.

Mrcreepssss:
this, is one of those things you don't know you want untill you get them.

Super3222:
I tested it and it worked.

X-TheDark:
Sorry, but it seems like you still can't transfer the AI from potato to intellicard (or AI core).
Unless that's intended, you'll have to fix this.

ThatGuyPhil:
The intellitato already exists

PR #1901, opened 2017-02-26
makes dismembering legs not stun- also makes blood bubble when slaughter demons use it

### LINKED ISSUE #1898: Monthly Salt Issue : Remove/nerf slaughter demons
I was going to write an essay, but I decided against it

Points being: 
1. They do stupid amounts of damage, three hits to crit an  unarmoured person, then they're only one grab away from instant ggnore death
2. You need a high DPS weapon to even stand a chance of killing it, they'll just retreat otherwise, or even worse just tank whatever you're hitting them with, crit you and gib you, then recover whatever health they lost 
3.There is no telegraphing when they come out of blood pools, and they come out with a speed boost meaning you require both full health/ stims to even stand a chance of escaping
4.Cleaning blood only delays them, all they need to do is wait around in medbay, or for someone to inevitably drag a corpse, which gives them all the opportunity they need to continue their rampage
5.They're just not fun to play against, there's no outwitting or ambushing a slaughter demon, it's either kill or be kille - sorry did you say you aren't at full health nor conveniently have a combat with buckshot loaded? Ah well y'know gitgud

And saying that they're wizard only is not an argument, it doesn't remove the fact they're a complete pain in the ass to play against, nor does it matter when bubblegum loot can release a slaughter demon too.

ShadowDeath6:
wow pseudo more like boohoodo
fixes #1898 somewhat

if you wanna get mad about dismembering leg thing not working remember that it already slows you down

:cl: ShadowDeath6
rscdel: Removed the stun from leg dismemberment.
rscadd: Blood bubbles when slaughter demons are about to pop out of it.
tweak: Slaughter demons now lose their speed boost from exiting blood much faster.
/:cl:

### DIFF
--- a/code/modules/mob/living/bloodcrawl.dm
+++ b/code/modules/mob/living/bloodcrawl.dm
@@ -81,9 +81,9 @@
 	else if(victim.reagents && victim.reagents.has_reagent("demonsblood"))
 		visible_message("<span class='warning'>Something prevents [victim] from entering the pool!</span>", "<span class='warning'>A strange force is blocking [victim] from entering!</span>", "<span class='notice'>You hear a splash and a thud.</span>")
 	else
-		victim.forceMove(src)
 		victim.emote("scream")
 		src.visible_message("<span class='warning'><b>[src] drags [victim] into the pool of blood!</b></span>", null, "<span class='notice'>You hear a splash.</span>")
+		victim.forceMove(src)
 		kidnapped = TRUE
 
 	if(kidnapped)
@@ -161,16 +161,18 @@
 	if(src.notransform)
 		src << "<span class='warning'>Finish eating first!</span>"
 		return 0
-	src.loc = B.loc
-	src.client.eye = src
-	src.visible_message("<span class='warning'><B>[src] rises out of the pool of blood!</B>")
-	exit_blood_effect(B)
-	if(iscarbon(src))
-		var/mob/living/carbon/C = src
-		for(var/obj/item/weapon/bloodcrawl/BC in C)
-			BC.flags = null
-			C.unEquip(BC)
-			qdel(BC)
-	qdel(src.holder)
-	src.holder = null
+	B.visible_message("<span class='warning'>[B] begins to bubble...</B>")
+	if(do_after(src, 25, target = B))
+		src.loc = B.loc
+		src.client.eye = src
+		src.visible_message("<span class='warning'><B>[src] rises out of the pool of blood!</B>")
+		exit_blood_effect(B)
+		if(iscarbon(src))
+			var/mob/living/carbon/C = src
+			for(var/obj/item/weapon/bloodcrawl/BC in C)
+				BC.flags = null
+				C.unEquip(BC)
+				qdel(BC)
+		qdel(src.holder)
+		src.holder = null
 	return 1
--- a/code/modules/mob/living/bloodcrawl.dm
+++ b/code/modules/mob/living/bloodcrawl.dm
@@ -163,9 +163,9 @@
 		return 0
 	B.visible_message("<span class='warning'>[B] begins to bubble...</B>")
 	if(do_after(src, 25, target = B))
-		src.loc = B.loc
-		src.client.eye = src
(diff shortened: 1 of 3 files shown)


code/modules/surgery/bodyparts/dismemberment.dm
```diff
@@ -169,7 +169,6 @@
 
 /obj/item/bodypart/r_leg/drop_limb(special)
 	if(owner && !special)
-		owner.Weaken(2)
```
Super3222, line 172: Instead of removing this make a variable that ignores pain, and work around it with your dismember code. You'll probably need to create a new variable - or just make it so mobs won't cause you to fall on your face after being dismembered.

code/modules/surgery/bodyparts/dismemberment.dm
```diff
@@ -183,7 +182,6 @@
 
 /obj/item/bodypart/l_leg/drop_limb(special) //copypasta
 	if(owner && !special)
-		owner.Weaken(2)
```
Super3222, line 186: Instead of removing this make a variable that ignores pain.

code/modules/mob/living/bloodcrawl.dm
```diff
@@ -161,16 +161,18 @@
...
-		var/mob/living/carbon/C = src
-		for(var/obj/item/weapon/bloodcrawl/BC in C)
-			BC.flags = null
-			C.unEquip(BC)
-			qdel(BC)
-	qdel(src.holder)
-	src.holder = null
+	B.visible_message("<span class='warning'>[B] begins to bubble...</B>")
+	if(do_after(src, 25, target = B))
+		src.loc = B.loc
```
Super3222: forceMove(B)

code/modules/mob/living/bloodcrawl.dm
```diff
@@ -161,16 +161,18 @@
...
+		src.client.eye = src
+		src.visible_message("<span class='warning'><B>[src] rises out of the pool of blood!</B>")
```
Super3222: visible_message

Super3222: delete src

code/modules/mob/living/bloodcrawl.dm
```diff
@@ -161,16 +161,18 @@
...
+		exit_blood_effect(B)
+		if(iscarbon(src))
+			var/mob/living/carbon/C = src
+			for(var/obj/item/weapon/bloodcrawl/BC in C)
+				BC.flags = null
+				C.unEquip(BC)
+				qdel(BC)
+		qdel(src.holder)
```
Super3222: holder

code/modules/mob/living/bloodcrawl.dm
```diff
@@ -161,16 +161,18 @@
...
+		src.holder = null
```
Super3222: holder

Super3222 (changes requested):
src pruning.
adjust code to forcemove.
don't be lazy and remove stunning from dismember. just don't.

code/modules/surgery/bodyparts/dismemberment.dm
ShadowDeath6: why would I add a pain variable
would you prefer it to do stamina or something

Super3222: `if(pain) owner.Weaken(2)`
`/obj/item/bodypart/r_leg/drop_limb(special,var/pain = TRUE)`
make it set pain as false.

Super3222:
After talking with Shadow yesterday he makes a pretty good point that it's more a universal problem then just slaughter demons.

Super3222 (approved):

Time-Green:
You mean getting hit once and get your leg dismembered, and then getting instantly killed?
Yeah that's a problem.

ghost:
Legstun dismember was even nerfed on /tg/ it's such an issue.

Dual eblade is always a 100% stun on melee range, as it will always dismember the leg on the first hit. Or, almost always. 

This in addition to it having a 75% blockchance means that the only reliable way to down them is to be in melee range, where they are immune to disarms and you need a melee weapon like a stunbaton to block it.

PR #1912, opened 2017-02-28
Combat shotgun crate costs more points for less + WT autorifle

ghost:
The combat shotgun is the best weapon in the game, even counting some nukeop weapons aside from the outright LMG.

An autofire shotgun that fires 90 damage buckshot or 60 damage slugs for cheaper than an energy gun and an ammo belt of it, with more, and three of those on top of that is a bit much.

However I don't want to nerf the combat shotgun too hard because 1. muh blob Xray and 2. It deserves to be powerful, just not as hilariously available as it is now.

So, the nerf x3's it's cost, meaning it's about the same as 3 laser crates. For one shotgun. Yes this seems excessive until you've seen how good the combat shotgun really is. It also means that one shotgun will show up in the crate rather than 3.

The WT autorifle also got a little nerf and costs 4500 points up from 3500 because it was also silly available and as usual bullet weapons 2strong. 

Remember that cargo usually ends up sitting on metric fucktons of points within 15 minutes because of exports and stocks, so the point thing isn't much of a downside except for bad cargo-it's mostly the number of shotguns.

#### Changelog

:cl:
rscadd: Combat shotgun crates now only contain one combat shotgun, down from three. They also cost more. The WT-autorifle crate has had it's price slightly raised.
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -294,12 +294,8 @@
 
 /datum/supply_pack/security/armory/ballistic
 	name = "Combat Shotguns Crate"
-	cost = 2000
+	cost = 6000
 	contains = list(/obj/item/weapon/gun/projectile/shotgun/automatic/combat,
-					/obj/item/weapon/gun/projectile/shotgun/automatic/combat,
-					/obj/item/weapon/gun/projectile/shotgun/automatic/combat,
-					/obj/item/weapon/storage/belt/bandolier,
-					/obj/item/weapon/storage/belt/bandolier,
 					/obj/item/weapon/storage/belt/bandolier)
 	crate_name = "combat shotguns crate"
 
@@ -328,7 +324,7 @@
 
 /datum/supply_pack/security/armory/wt550
 	name = "WT-550 Auto Rifle Crate"
-	cost = 3500
+	cost = 4500
 	contains = list(/obj/item/weapon/gun/projectile/automatic/wt550,
 					/obj/item/weapon/gun/projectile/automatic/wt550)
 	crate_name = "auto rifle crate"


ShadowDeath6:
Do you plan on actually nerfing combat shotguns or is that for later

drderp3635:
This seems a bit excessive

ghost:
Not excessive, I covered why. Tell me why it would be excessive.

Shadow, no I don't plan on nerfing the combat shotgun too much. If anyone has ideas that would make it not just a shitty riot shotgun i'm all ears.

ShadowDeath6:
Probably need to lower the net damage or change the spread

ghost:
that's an issue with all buckshot

Super3222:
How is this an improvement?

ShadowDeath6:
Then nerf it ,:^)

ghost:
This is an improvement because why would the strongest weapon in the game be available for less than an energy gun, which is much shittier. And why do you get more of them than an energy gun.

I would nerf buckshot, but i'll have to review projectile code to figure out how spread is handled.

drderp3635:
I just think 9 times the regular price is a bit much. Maybe keep the only 1 gun but set the price to 3000 instead of 6000

ghost:
It shouldn't be less expensive than the SWAT crate, which is much less powerful than this, or the combat knife crate, also less powerful.

ShadowDeath6:
There is a real problem with buckshot though, I'm pretty sure you can kill a horrorling from max health in about 3 or 4 shots

ghost:
Yeah buckshot is bullshit-All of the pellets hit for scratch damage so armor dindu shit, they penetrate armor, cause bleeding and shitdick juggs to death and outright dunk anything in their way.

I don't want to make them WEAK, but

Frostben:
Wait what? Buckshot penetrates armor??? Muh immursion senses call bullshit on that! Why would buckshot be armor penetrating? That doesn't make a lot of sense because your average buckshot pellet is smaller than a 9mm round and travels at a lower velocity as well. Which would make it bounce off something like sec armor no problem.

ShadowDeath6:
I don't think it actually has armor pen. 

https://github.com/tgstation/tgstation/pull/24633
^very relevant, you should definitely port this

Super3222:
That's definitely going to help in some places, but not all of them.

One of the most OP things about the combat shotgun is it's power in close range. As the HoS I've managed to instantly kill wizards and revs (who had weapons trying to kill me and was valid) just by spam clicking them and mashing the Z button to pump my shotgun.

You won't believe how fast both combat shotguns and riot guns will killl or crit someone if you unload the entire barrel into them. And by the time you're already shooting you've managed to crit them with third shot with buckshot or managed to completely drain their stamina for the riot rounds.

So that'd be a good thing to port, but also finding a way to nerf by the pumping action by maybe 1-2 seconds from where it is now.

ShadowDeath6:
Each pellet does 12.5 damage and I think it fires 6 or 7 each shot

Cruix:
Why can't you only do one of these things and not change their cost by a factor of nine in one go

ghost:
Because @Cruix for reasons explained above.

You can't make 'atomic' PR's for big changes, because 'atomizing' them doesn't do shit except make their effects long and drawn out and not immediately noticable, leading to further and further changes or giving up on it as saying 'well that changed nothing why not revert.'

It is more logical to go with the stronger foot out, and pull it back some if it doesn't work out rather than to baby-step your way around the issue.

Cruix:
When has anyone ever said 'well that changed nothing why not revert.' I would much rather have two changes that might barely change the game than one that might change it far too much. A factor of nine is too much for one go.

Do one or the other, but not both.

ShadowDeath6:
it's actually super hard to revert small shit because of "it changes nothing why revert it"

PR #1918, opened 2017-03-01
Killer (Emagged) Photocopiers

ArgyleSweatshop:
Photocopiers should be emaggable, so why not let them do something horrific to people shoved in them? That's the question I think I solved with this pull request- finally, you can murder people with paperwork.

Photocopiers weren't able to be emagged, and I thought that was sad. Now, photocopiers can be used by traitors to brutally crush people and get a neat souvenir.

#### Changelog

:cl:
rscadd: -Photocopiers can now be emagged
rscadd: -Emagged photocopiers now gib people dragged into them
rscadd: -After gibbing, produces either a blood-red cutout of the person or a less-than-useless talisman
/:cl:

### DIFF
--- a/code/modules/paperwork/photocopier.dm
+++ b/code/modules/paperwork/photocopier.dm
@@ -328,6 +328,10 @@
 			new /obj/effect/decal/cleanable/oil(get_turf(src))
 			toner = 0
 
+/obj/machinery/photocopier/emag_act(mob/user)
+	if(!emagged)
+		emagged = 1
+
 /obj/machinery/photocopier/MouseDrop_T(mob/target, mob/user)
 	check_ass() //Just to make sure that you can re-drag somebody onto it after they moved off.
 	if (!istype(target) || target.anchored || target.buckled || !Adjacent(user) || !Adjacent(target) || !user.canUseTopic(src, 1) || target == ass || copier_blocked())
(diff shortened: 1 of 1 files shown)


ghost:
I'm not a fan of body gibbing, plus more utility to the emag, nah.

Would prefer it if they were just hackable, sort of like suit storage units, and instead of gibbing they just did a load of damage. As in, death.

ArgyleSweatshop:
It's the same as the gibber, albeit very slightly more accessible. In fact, its worse than the gibber, because all photocopiers are in fairly visible locations.

ShadowDeath6:
have it grind em up and spit out gibs with the bloody body like the emagged crusher does

ShadowDeath6:
also the gibber is literally called the gibber and is in either a job specific area or has to be made with R&D

ArgyleSweatshop:
I'll look into making that work that way, and make sure the PR works

ArgyleSweatshop:
Can someone help me figure out why this won't compile?

code/modules/paperwork/photocopier.dm
```diff
@@ -349,7 +356,21 @@
...
-
+		if(emagged && (target != user))
+			target.gib()
+			visible_message("<span class='warning'>[target] is crushed in a tragic photocopying accident!</span>")
+			if(toner <= 0 || !isCarbon(ass))
+				return
+			
+			else
+				if(prob(90)) //90%
+					/obj/item/target/syndicate = new(src.loc)  //photocopier spews out a blood red paper person
```
ShadowDeath6, line 367: this is probably where it's messing up since it's not an item
Also you don't need to use src.loc, loc on its own works the same way

PR #1934, opened 2017-03-04
makes changelings able to remove flesh clothes

ShadowDeath6:
ports https://github.com/tgstation/tgstation/pull/24462 by @XDTM

:cl: ShadowDeath6
rscadd: Changelings can now remove the flesh clothing created by transform, by XDTM.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/transform.dm
+++ b/code/game/gamemodes/changeling/powers/transform.dm
@@ -17,6 +17,7 @@
 		user << "<span class='notice'>You absorb [src] into your body.</span>"
 		user.put_in_active_hand(src)
 		user.drop_item()
+		user.visible_message("<span class='warning'>[src] crumbles into flakes of... something.</span>", , 3)
 
 /obj/item/clothing/under/changeling
 	name = "flesh"
@@ -28,6 +29,7 @@
 		user << "<span class='notice'>You absorb [src] into your body.</span>"
 		user.put_in_active_hand(src)
 		user.drop_item()
+		user.visible_message("<span class='warning'>[src] crumbles into flakes of... something.</span>", , 3)
 
 /obj/item/clothing/suit/changeling
 	name = "flesh"
@@ -40,6 +42,7 @@
 		user << "<span class='notice'>You absorb [src] into your body.</span>"
 		user.put_in_active_hand(src)
 		user.drop_item()
+		user.visible_message("<span class='warning'>[src] crumbles into flakes of... something.</span>", , 3)
 
 /obj/item/clothing/head/changeling
 	name = "flesh"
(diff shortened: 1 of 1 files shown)


Groudonmaster:
Isn't the intention to make the flesh clothes a giveaway?

Super3222:
Only changelings can remove flesh clothing. it will still operate as NODROP to everyone else. So if you have a changeling in cuffs they're fucked.

code/game/gamemodes/changeling/powers/transform.dm
```diff
@@ -9,37 +9,94 @@
 
 /obj/item/clothing/glasses/changeling
 	name = "flesh"
-	flags = NODROP
+	flags = NODROP | DROPDEL
+
+/obj/item/clothing/glasses/changeling/attack_hand(mob/user)
+	if(loc == user && user.mind && user.mind.changeling)
```
Super3222, line 15: Speaking of cuffs, could you make these public with a visible message.

ShadowDeath6: why

Super3222: It promotes using this seriously / in stealth instead of walking down the hallway popping clothes on and off. Powers that don't immediately reveal lings are taken for granted, and generally abused because they can get away with it.

ShadowDeath6: how would making it reveal them improve the game at all other than making it autofuck them from the game if they're in dorms or something and some random assistant walks by and yells "x is ling" because they remove their fake ID to replace it

Time-Green: Have it only do so when the person is directly/almost directly adjacent

ShadowDeath6: that sounds better

XDTM:
Do changelings send a message when transforming here? If not, there's not much reason to make a lesser aspect of it more noticeable.

ShadowDeath6:
They don't, no.

Cruix:
Why are you putting these in peoples hands and depending on dropdel to delete them instantly instead of just deleting them instantly?

ShadowDeath6:
>instead of doing this via qdel, set the NODROP flag to off and set DROPDEL and then just let the parent proc run and it should have the same effect without requiring a forced qdel and doing snowflake things to sidestep NODROP
 
If it's a problem, it's easy enough to make it qdel.

Cruix:
But this is sidestepping NODROP 

You are literally removing NODROP in order to do this

ShadowDeath6:
_shrug_

PR #1886, opened 2017-02-23
Buffs transform sting, makes mutesting use reagents, changes hallucinogen pathogen sting

ghost:
First off before you read this remember my admin/mod status has no bearing on what I try to PR on github and thus has no impact.

Anyway I was meaning to do this for a while. 

1. Mutesting-A sting not using chemicals that can be countered, especially one as prolific and useful as mutesting, left a bad taste in my mouth. It now uses mute toxin, which means that if someone cannot speak, checking their chemicals in the bloodstream might clue someone in a ling was involved...or an enterprising chemist. Additionally, mutesting silence time remains the same. Yes, this is effectively a slight nerf. I don't worry about chemical purgatives however, as anyone running around constantly hopped up on purging chems 'just in case' or 'becuz lings' is probably gonna get yelled at.

2. Transform sting-/tg/ nerfed this to hell and i'm not sure why. No, you still can't transform sting people into plasmamen for instant kills, but you can transform sting without it damaging your genomes. Before, transform sting was an effectively useless and overpriced sting after /tg/'s nerfs since you couldn't use it to cover your tracks, couldn't use it to disguise in a crowd or fool people, or any of that. Now, after the buff, it now doesn't cost as much dNA as it used to for it's relative utility and doesn't damage genomes. It still has a hefty chem cost however. If 'disability sting' is still an issue, I can make the sting transfer mutadone, or alternatively re-add the hefty DNA cost. Let's just see how it is though.

3. Hallucinogen pathogen sting-I hated this and I think a lot of people did too. It's an era from the days of ephermalis. You couldn't, after eph's changes, use it as it was intended-instead, as a change nobody looked forward to, you infected someone with a plague for a very hefty DNA cost (that hardly matters since you can readapt anyway due to the plague), which makes everyone hallucinate like hell. Not to mention the plague itself was somewhat buggy and poorly designed, making recurrances and causing hallucinations even after treatment, and sending false positive 'it's cured' messages (which wasn't the intent.) Now it does what old hallucination sting does-Doses a single target with hallucinations, via injection of mindbreaker (so that you can actually abate it's symptoms with synaptizine unlike old hallucisting)

#### Changelog

:cl:
rscadd: Changeling chemical glands have increased their viability, vastly improving the utility of changeling stings. In particular, Nanotrasen crews should be aware that 'mute sting' now injects mute toxin, and the 'transform sting' no longer damages changeling genomes.
[...]

PR #1936, opened 2017-03-05
Revert "Buffs transform sting, makes mutesting use reagents, changes hallucinogen pathogen sting" Or How I Learned To Actually Read Adam's PR's Before APPROVING THEM

Ephemeralis:
Reverts yogstation13/yogstation#1886

Appalled this passed, and not for the reasons you'd might think.

Genome damaging is basically IC/changeling flavour for adding a cooldown to abilities to prevent them from being spammed. I just finished a round where a Changeling spam transformed four people in about a second and a half because Adam's a fucking halfwit and removed the cooldown from the sting because "muh genomes".

The Mutetoxin change is good, we should keep that one.

We can look at **REDUCING** the transform sting cooldown and cost, but this was just absolute fucking insanity. We can also maybe rework the virus to be NOSPREAD and simulate the transformation over time, giving people a means to counter a FULL IDENTITY change outside of getting their gene encoding stored by a Geneticist at round start, because who the fuck does that?

Stop approving Adam's PR's without thinking about them first. All of you should know better.

### DIFF
--- a/code/game/gamemodes/changeling/powers/tiny_prick.dm
+++ b/code/game/gamemodes/changeling/powers/tiny_prick.dm
@@ -63,10 +63,11 @@
 /obj/effect/proc_holder/changeling/sting/transformation
 	name = "Transformation Sting"
 	desc = "We silently sting a human, injecting a retrovirus that forces them to transform."
-	helptext = "The victim will transform much like a changeling would. The effects will be obvious to the victim."
+	helptext = "The victim will transform much like a changeling would. The effects will be obvious to the victim, and the process will damage our genomes."
 	sting_icon = "sting_transform"
-	chemical_cost = 30
-	dna_cost = 1
+	chemical_cost = 40
+	dna_cost = 3
+	genetic_damage = 100
 	var/datum/changelingprofile/selected_dna = null
 
 /obj/effect/proc_holder/changeling/sting/transformation/Click()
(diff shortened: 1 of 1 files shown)


Ephemeralis:
What I have in mind for TF sting:

Gives virus, virus has 30s minimum incubation time. Variable chance to progress a stage every game tick, causes small amounts of brute (and I mean small, 1-3 at a very low chance per cycle) and gives some fucknasty "oh god my entire dna is changing help" lines so people know what the fuck is going on. Once virus hits severity 5 and self-cures without a trace, then it changes identity as per old TF sting. Will not self-cure before severity 5, will be affected by Preternis disease immunity.

Average time to stage 5 would be tuned to be 1-3 minutes.

Virologists can cure it before severity 5 and stop the virus. Geneticists can take SE copies during the virus to restore the person to normal after it self-cures.

I'm okay with the hallucination virus getting axed because it did ultimately prove out to be pretty unfun in practice given that people are shit at virology.

Hallucinations are still a really annoying mechanic in general though. We should probably tidy them up a little bit and make some of them far more compelling than they currently are.

Cruix:
If you want to remove something but not everything else, you of all people are more than capable of making your own commit to change those specific things instead of hitting that big lazy revert button.

If you want to you can also take ten seconds to look at Adam's PR history and see that nearly half of his PRs do not get merged, and those that do often wait weeks to do so and are heavily modified in that time, either because they are broken or because he knows that they will never get merged as they are.

ShadowDeath6:
the virus thing is dumb though, can't it be like a reagent that's stored in someone that after x cycles it turns them into the transformed thing?

PR #1939, opened 2017-03-06
message shows when shadowlings are getting burnt in the light

ShadowDeath6:
he does it for free


Cruix:
This breaks tactical shadowling cloning / SE injection stealthlings

ShadowDeath6:
they already show up as big black men when you take em out of cloning, the least you could do is let them pop out in the dark

doerakpoes:
Good job jelly

Jc1469:
Kinda hurts sling stealth as a whole really.  Slings can someones sneaky beaky like past people with lights in maint, while taking a tiny bit of damage, un-noticed because what is an eyes.   Now they can't really do that.

ShadowDeath6:
They only take damage when they're entirely visible, so there's not really a stealth aspect at that point. The main reason this was made is because drones have been accidentally burning shadowlings by repairing broken lights and can't tell if they're shadowling since they can't examine people anymore

Issue #1943, opened 2017-03-07
Buff blobernaut healing 


drderp3635:
It takes absolute ages for a blobernaut to regenerate after getting hit at all, to a point where you spend three times as much time licking your wounds next to a node then you actually do fighting things.

ShadowDeath6:
Good idea.

PR #1944, opened 2017-03-07
increases blobbernaut healing

ShadowDeath6:
#1943 

:cl: ShadowDeath6
tweak: Blobbernauts now heal 5% health while on blob tiles, up from 1.25%.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -37,7 +37,7 @@
 				H.color = overmind.blob_reagent_datum.complementary_color
 			else
 				H.color = "#000000"
-		adjustHealth(-maxHealth*0.0125)
+		adjustHealth(-maxHealth*0.050)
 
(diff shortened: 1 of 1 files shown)


ghost:
Blobbernauts are extremely expendable and extremely cheap. No reason to buff.

ShadowDeath6:
>implying 40 bp is cheap

Notamaniac:
Can we have this scale from 1.25% to 5% over time (i.e. 0.25% every two ticks?) while blobbers aren't taking damage? 

From 1% health 1.25% regen takes about 80 ticks (2 mins in byond time irrc) to heal to full, while for 5% regen it only takes 20 ticks (irrc 30 seconds in byond time). On paper this allows blobbernauts to attack 4 times as frequently than before, which could be pretty OP.

ShadowDeath6:
I could probably do that but it's more work smh

Jc1469:
I concur with their regen being pathetic.  That and they die pretty fucking quickly by not being on blob tiles too.  

>extremely expandable and extremely cheap.  

were I a person to meme and care, I think my response to this would be "lmao"

PR #1949, opened 2017-03-10
Fixed being unable to attack blobs

Cruix:
:cl:
fix: Blobs are no longer impossible to attack with melee weapons.
/:cl:


ShadowDeath6:
why was this added in the first place

PR #1950, opened 2017-03-10
holla holla get dolla

ShadowDeath6:
:cl: ShadowDeath6
rscadd: Money bags can now hold infinite money. That's still all they can hold, though.
/:cl:


Super3222:
Why? I'll never know.

ShadowDeath6:
tl;dr I wanted to get jackpot as assistant to I went to mining but I couldn't hold enough coins in my bag so I had to go back with only like 40 coins so I went back which sucked and then won jackpot but I couldn't hold any of the damn coins in my bag

PR #1960, opened 2017-03-11
Slightly buffs the invisible touch ability.

Super3222:

Come on now, it's a simple trick that can barely cause any harm. Only confusion.

- It now has five uses instead of 3.
- You can use it on closets.
- When it's reverting it will vanish and reappear ten times before finally recovering.

Things that stay the same
- You can't use it on the same thing twice.
- Everything else not mentioned.

#### Changelog

:cl:
rscadd: The Mime's invisible touch can be used on more objects (3 -> 5), can be used on closets, and has a cool reversion effect on objects.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/godhand.dm
+++ b/code/modules/spells/spell_types/godhand.dm
@@ -164,9 +164,9 @@
 					if(!user)
 						break
 					passes--
-					user.alpha = 0
+					animate(user, alpha = 0, time = 1)
 					sleep(3)
-					user.alpha = initial(user.alpha)
+					animate(user, alpha = initial(user.alpha), time = 1)
 					sleep(2)
 			else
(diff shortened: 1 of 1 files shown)


Cruix:
It would look better of you used Animate() instead of having it flash

code/modules/spells/spell_types/godhand.dm
```diff
@@ -195,7 +196,20 @@
 
 /obj/item/weapon/melee/touch_attack/proc/reverttarget(atom/A)
 	if(A)
+		var/initA = initial(A.alpha)
+		var/countleft = 10
+		while(countleft > 0)
+			if(!user)
```
Time-Green: Probable cause of it not compiling. User is not define anywgere

Cruix:
Animate makes it fade smoothly, just don't make it flash

PR #1967, opened 2017-03-12
LOOC - Take Two

Super3222:
Ported https://github.com/NTStation/NTstation13/pull/796 with modifications.

What is LOOC?
Local OOC.

What does it do?
It lets you message people about stuff you'd wrap around in (()).  It's a neat way to filter the chat, and it's fair to say that in this method this is more noticeable.

What did you change?
- LOOC only works for mobs that can speak your language. No exceptions. If (())'s weren't going to work, then you can bet your ass that this isn't going to either.
- LOOC does not work if you're a ghost. The only exception is if you have an admin rank.
- I added in a stylesheet (font color / style) for LOOC. Taken from what was given to me.
- LOOC does not work if you are unconscious or in crit (health below 0).

#### Changelog

:cl:
experimental: TAKE TWO BOYS. IF YOU'RE AN OLDIE YOU'LL REMEMBER THIS ONE: LOOC IS BACK.
/:cl:

Things that need to be fleshed out:
- [ ] Make LOOC disabled if the mob is muted / unable to speak. Possible exception if you're the mime. Possible.
- [ ] Global mute LOOC option?
- [ ] Separate listen_looc() into two verbs. One for players to mute/unmute LOOC at will, and the other for admins to control whether they want to listen to EVERY LOOC message.
- [ ] Make LOOC output your key + IC name. However, with aghosts it'll be only their ckey. Not their name.


ShadowDeath6:
no thanks, I've played servers with LOOC and it doesn't add very much to the game :-1:

Super3222:
LOOC is optional to every player. It's just another one of those things where if you want to use it, go ahead and use it. If you don't than that's fine too, nobody's going to fault you for it. It's all about game design at this point, and abridging a way to filter OOC, but still possible relevant IC chatter, is a smooth decision to take.

ShadowDeath6:
not a big fan of the "it's optional so it should be added" meme tbh

Super3222:
That's just the thing Shadow - LOOC isn't that big of a deal. Especially with the way that I'm doing it.

Maybe it's just the specific type of server, and how people were using it, that made it bad. We have a community of different amounts of people to experiment with here.

If this sooner down the line gets merged, and it turns bad, and I mean BAD BAD, then ping me so I can revert it or just do it yourself.

code/modules/client/verbs/ooc.dm
```diff
@@ -265,4 +265,88 @@ var/global/normal_ooc_colour = OOC_COLOR
...
+	set name = "LOOC" //Gave this shit a shorter name so you only have to time out "ooc" rather than "ooc message" to use it --NeoFite
+	set desc = "Local OOC, seen only by those in view."
+	set category = "OOC"
+
+	if(say_disabled)	//This is here to try to identify lag problems
+		usr << "<span class='warning'>Speech is currently admin-disabled.</span>"
+		return
+
+	if(!mob)	return
+	if(IsGuestKey(key))
```
ShadowDeath6, line 281: is this necessary

code/modules/client/verbs/ooc.dm
```diff
@@ -265,4 +265,88 @@ var/global/normal_ooc_colour = OOC_COLOR
...
+
+	if(istype(mob, /mob/dead/observer))
+		if(!holder)
+			src << "<span class='warning'>Ghosts cannot use LOOC.</span>"
+			return
+
+	if(isliving(mob))
+		var/mob/living/L = mob
+		if(L.health < 0)
+			src << "<span class='warning'>LOOC doesn't work while you're in crit.</span>"
```
ShadowDeath6, line 293: say critical condition, crit sounds like jargon

code/modules/client/verbs/ooc.dm
Super3222: Ported from NT

code/modules/client/verbs/ooc.dm
Super3222: Sure.

matskuman5:
I'm... sorta up for this.

Not sure if it would work with this community (memers will just use it to spam "U'RE GETTING REPORTED U FUCKG GRIFR").

Then again it's just used with ((these)) anyway so I don't see a reason why it shouldn't be implemented

code/modules/client/verbs/ooc.dm
oisin100: I would leave it just incase, Guests can uses OOC, should be the same for LOOC.
Might be the case that a downstream allows guests to join, highly unlikely though.

Kmc2000:
I'm for this, but against how ghosts can't feed you sweet memes thru LOOC like when it was originally added.

Mrcreepssss:
a small issue i have with the whole you can't use it when mute thing is that it disallows people muted for something like genetics from educating a perhaps not very experienced geneticist how to cure said mute.

ShadowDeath6:
Looc: "HELP I GOT MUTE STAnged"

Kmc2000:
@Mrcreepssss


Yeah or they can abuse /me and NOT have every admin ever see their metagamery

Mrcreepssss:
@ShadowDeath6 im pretty sure admins have already thought of said possibilities, also similair counts with ghosts helping people who could use it though, the mentor system also helps with that.
@Kmc2000 ehm, i'm not sure how helping someone understand their job is  metagamey?

Nickvr628:
I support this. It prevents ((USE THE DAMN SURGICAL CONSOLE)) accidentally broadcasted over comms for all to hear, and it lets me shout **LOOC "YOU IDIOT THE SURGICAL CONSOLE TELLS YOU THE NEXT STEP"**.

+1

PR #1969, opened 2017-03-13
borer bugfixes and qol

ShadowDeath6:
fixes bug where borers would get bugged and couldn't use powers inside a host
fixes the vomit message not playing for a little while after the borer was actually vomited
also makes the health scanner show up the borer message at the end rather than at the beginning

remember to review my code because I can't test a lot of it as DD is really awful and doesn't work on my computer but I'm pretty sure there's nothing wrong with it

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer.dm
@@ -35,6 +35,7 @@ var/banned_borer_emotes = list("*collapse", "*collapses", "*surrender", "*surren
 	var/dominate_cooldown = 150
 	var/control_cooldown = 3000
 	var/leaving = 0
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/borer/borer.dm
```diff
@@ -213,6 +215,12 @@ var/banned_borer_emotes = list("*collapse", "*collapses", "*surrender", "*surren
 			if(hud.icon_state == "borer")
 				victim.client.images -= hud
 
+	leaving = 0
+	var/obj/item/organ/borer_home/oldhome = vessel
+	oldhome.borer = null
```
ShadowDeath6, line 220: this runtimes because `cannot set null.borer` but it doesn't cause any problems and I don't know why it's happening

PR #1971, opened 2017-03-13
Nerfs Dual-Eswords and Raises the price of eswords

ShadowDeath6:
:cl: ShadowDeath6
tweak: Dual Esword has a much lower chance to block things it doesn't already reflect, like stunbatons or floor tiles.
tweak: Esword cost increased from 6 to 7 telecrystals.
/:cl:


ShadowDeath6:
if this doesn't work I think it's possible(tbh I don't know how to do it since desword hit reaction code is hard to read) to either make it have the block chance for only melee or only ranged, so it's possible to make it so it more effective at range / less effective against stunbaton, both which counter it

Super3222:
:-1: just for the title.

ShadowDeath6:
sorry no jokes allowed here folks

ghost:
Good pr.

PR #2003, opened 2017-03-17
Removes a test object from the uplink

Time-Green:
I used it to test the ported code, I forgot to remove it, honk.

(I commented it out for the sake of an example, keep it?)
:cl:
fix: Removes a test obj from the uplink
/:cl:


Slinky5890:
Okay but what I wanna know is, why was it 'testobjyay' and not just 'testobj' ?

Cruix:
Because testing is exciting, so naturally you would exclaim 'yay' while doing it.

Super3222:
Just delete it. There's no point in having it there anymore.

PR #2018, opened 2017-03-19
Messed with changeling DNA handling

Cruix:
This is to fix the issue addressed in https://github.com/yogstation13/yogstation/pull/1936

Stole a proc from https://github.com/tgstation/tgstation/pull/22608

Incidentally this fixes the bug where you could not transform into a person if they were named "Drop Flesh Disguise"
:cl:
tweak: Using transformation sting or hive channel DNA as a changeling will remove the DNA from the changeling who used it.
tweak: Multiple changelings cannot extract DNA from the same person using extract DNA sting, and no changeling can use extract DNA sting on someone affected by transformation sting. Absorbing the person will always work.
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/tiny_prick.dm
+++ b/code/game/gamemodes/changeling/powers/tiny_prick.dm
@@ -75,11 +75,14 @@
 	if(changeling.chosen_sting)
 		unset_sting(user)
 		return
-	selected_dna = changeling.select_dna("Select the target DNA: ", "Target DNA")
+	if(user.mind.changeling.stored_profiles.len < 2)
+		user << "<span class = 'warning'>We do not have enough stored DNA for this.</span>"
+		return
+	selected_dna = changeling.select_dna("Select the target DNA: ", "Target DNA", FALSE, FALSE)
 	if(!selected_dna)
 		return
 	if(NOTRANSSTING in selected_dna.dna.species.specflags)
-		user << "<span class = 'notice'>That DNA is not compatible with changeling retrovirus!"
+		user << "<span class = 'notice'>That DNA is not compatible with changeling retrovirus!</span>"
 		return
 	..()
 
@@ -92,6 +95,8 @@
 	return 1
 
 /obj/effect/proc_holder/changeling/sting/transformation/sting_action(mob/user, mob/target)
+	if(!user || !user.mind | !user.mind.changeling | !(selected_dna in user.mind.changeling.stored_profiles))
+		return
 	add_logs(user, target, "stung", "transformation sting", " new identity is [selected_dna.dna.real_name]")
 	var/datum/dna/NewDNA = selected_dna.dna
 	if(ismonkey(target))
@@ -105,8 +110,13 @@
 		spawn(10)
 			C.real_name = NewDNA.real_name
 			NewDNA.transfer_identity(C, transfer_SE=1)
+			C.dna.sting_extractable = FALSE
 			C.updateappearance(mutcolor_update=1)
 			C.domutcheck()
+	user.mind.changeling.stored_profiles -= selected_dna
+	user.mind.changeling.profilecount--
+	qdel(selected_dna)
+	selected_dna = null
 	feedback_add_details("changeling_powers","TS")
 	return 1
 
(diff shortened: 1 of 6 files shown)


ShadowDeath6:
>different changelings can't extract dna from the same person

???? Why?

code/game/gamemodes/changeling/powers/tiny_prick.dm
```diff
@@ -175,16 +185,21 @@
...
 	var/protect = 0 //Should the system be prevented from automatically replacing this DNA?
 	for(var/datum/objective/escape/escape_with_identity/ewi in user.mind.objectives)
 		if(ewi.target == target.mind)
 			protect = 1
 			break
+	if((user.mind.changeling.has_dna(target.dna)))
+		user.mind.changeling.remove_profile(target)
+		user.mind.changeling.add_new_profile(target, user, protect)
+		user << "<span class='notice'>We refresh our DNA information on [target]!</span>"
+		return 1
```
ShadowDeath6, line 197: Is it physically possible to do this anymore since you made it so you can't extract DNA from someone who already had their DNA extracted?

Cruix:
If it were possible to use extract sting twice, you could just keep going back to the same person and farming them for DNA to transform people with.

code/game/gamemodes/changeling/powers/tiny_prick.dm
Cruix: The check is below this, so yes.

ShadowDeath6:
that takes ages to do and it costs a lot of chemicals to do, I think it's really shitty to deny other changelings the DNA of someone when one dumbass takes it then uses transform sting on someone else

Cruix:
You can just absorb the person instead.

ShadowDeath6:
>forcing other changelings to do more work in order for to get the same DNA that one asshole extracted

:cry: 

can't you just make it so the same changeling can't spam get the same DNA over and over again without making it so every other changeling can't get it

doerakpoes:
Well I mean we still work with the logic of hey i just put a bit of DNA in to your blood stream now your entire body has changed so....

Time-Green:
I think the lore was that it uses a genetic virus to quickly apply the changes, same with genetics.

doerakpoes:
That would atleast make some sense

Mrcreepssss:
i think it shouldn't be like this either.

ShadowDeath6:
@cruix can you please make it so that another ling using extract on somebody doesn't cockblock other lings it's a really bad idea

Cruix:
They can just absorb them

ShadowDeath6:
That's not always easy to do considering it forces you to kill the person for their dna, which makes people whine about murderbone / puts the ling that has to absorb them in a compromising situation

Why would you make it like that in the first place

Cruix:
Because the whole point is that you cannot spam the changeling retrovirus

ShadowDeath6:
How does more than one ling having access to the same dna changing that

Cruix:
Because you can extract sting the same person over and over again to make a station full of john sweeps

ShadowDeath6:
ok cruix
ok 
ok 
ok
ok
I get how having the same ling spamming the same guy is a problem that should be remedied

I'm asking
why
are
you
stopping
other players
from dna extracting the same dude
I can't see any situation where you can "duplicate" dna without super good teamwork and that shouldn't really be prohibited
I'm sure you can make a way that stops the same guy from spamming extract that doesn't force other players to absorb them rather than using extract

LarissaMayrink:
Stations full of john sweeps are cool, its one of the few Interesting ling tactics to force a shuttle call that don't actually make the station kill itself, nor comes with murderboning. Plus people are already used to it and most manage to deal with it without much confusion as soon as they realize that now there are 5 cali wallick roaming the station, and the ling still needs to expose himself to inject people.

Its a mostly innofensive way of generating confusion without the need to force people out of the game(which is really good). If this gets cut out we will see a hell of a lot more agressive moves from the lings, bombing regions, attacking heads, or using spread infestation/horror form, the last two which need a bunch of absorptions, which will force even more people out of the round. 

After everything said; this is a really bad design decision as it will encourage destructive behavior that forces people out of the round.

ShadowDeath6:
The problem with transform sting being that spammable is that it lets you extract dna from a genetics human that has full disabilities then go to town on everyone you see

Time-Green:
Well then, we just dont make it carry over mutations and we'll still be able to make John Sweeps station a reality

PR #2053, opened 2017-03-27
Fixes a bug with the global print option

Super3222:

Fixes the global print option on security computers.
I think we all learned an important lesson today. Never locate() numerical values.

#### Changelog

:cl:
fix: The global print option on security computers has been restored.
/:cl:


ShadowDeath6:
noice

Issue #1839, opened 2017-02-16
Make library books provide minor bonus (Warning: long)


ArgyleSweatshop:
Intended to add to a major librarian overhaul, that would start with #1838 . This centers on the addition of a new stat for creatures identified as "Book Pleasure".

Reading a leisure book provides “Book Pleasure”(which would be a hidden stat) after a delay of 30 seconds, during which other books can’t be read. This bonus can only be gained once per book for each person that reads it. Higher pleasure levels decay much faster- reading one book would last 10 minutes.
-Pleasure decay is reduced by hot cocoa, tea, and coffee.
-The librarian has a slower pleasure decay, and a reduced delay on book reading.
-Low levels of Book Pleasure provide a miniscule movement speed bonus and a slight reduction in stun times.
-Medium levels of Book Pleasure very slowly heals brain damage. Allows you to walk at the speeds other people run, but not move faster when you are running.
-High levels of Book Pleasure very slowly heals eye damage, and heals brain damage at a slightly faster level than before. In addition to the medium bonuses, you regain the bonuses of low level book pleasure.
-Extreme levels of book pleasure make these effects permanent.

AdamElTablawy:
By reading books, you get stun resistance and perfect-walk.

Yes, this makes sense.

no.

ArgyleSweatshop:
The important thing with my idea is providing reasonable minor bonuses, not the specific bonuses.

The idea behind the bonuses I listed was that after you read a book, you become more focused. You have more energy to put into the things you are doing. The stun resistance would be at nicotine levels, if not lower. Perfect walk would be nice, but also wouldn't be very strong.

Lightning38088:
being faster is just ridiculous... have you seen people on drugs from chem ? (cough bob before the nerf cough ) Their stun times are cancerous and unless you lethal them they'll always get up before cuffing is complete.

ArgyleSweatshop:
Nicotine levels are extremely low stun resistance. And when I say a minimal speed increase, I mean minimal.

Cruix:
>Gang is setting up the dom in viro
>Boss: Alright gang time to dom, but first
>*dumps 20 copies of the lusty xenomorph maid on the ground*
>Boss: Everyone read up we need the stun resist
>*The reading rainbow gang is attempting a network breach in virology*

It's almost worth it just for this

Dagdammit:
[...]

Issue #1840, opened 2017-02-16
Further librarian overhaul: Addition of "Technical Tablets"


ArgyleSweatshop:
Continued overhaul that goes with #1838 and #1839 . This centers around the addition of books that would provide wide ranging bonuses for specific purposes.

**Technical Tablets(E-Books that provide job bonuses)**
-These E-books are not readable like leisure books
-Technical tablets are assigned a type. When tablets are read, they provide a temporary bonus. This bonus is permanent for librarians.
-Technical tablets are created and “upgraded” by a machine in the librarian’s study. Feeding the machine certain objects will upgrade all current and future tablets of related type(s), allowing to them to provide better bonuses to their readers.

There are several types of technical tablets:
**Medical Instruction Tablets-**
-A base tablet makes you slowly heal yourself when holding a first aid kit.
-Feeding the machine a toolbox will increase success for ghetto surgery.
-Feeding the machine a stethoscope allows you to see lung damage and check the pulse of humans just by examining them.
-Feeding the machine a cyborg scanner allows you diagnose cyborgs and identify damaged components
-Many other things are potentially feedable, lots of experimenting to be had!

**Engineering Instruction Tablets-**
-A base tablet allows you to stop yourself before touching live wires without insulated gloves.
-Feeding the machine a toolbox slightly speeds up all sorts of actions involving tools, including most deconstructions and most constructions.
-Feeding the machine an industrial welder makes you temporarily experienced enough to weld without looking. You no longer take eye damage when welding.
-Feeding the machine a stun baton allows you to craft stun gloves
-Many, many other things are potentially feedable

**Security Instruction Tablets-**
-Feeding the machine a security weapon allows you to be more effective with it. This ranges from increasing stun times to dealing crippling eye damage with pepper spray.
-Feeding the machine cable cuffs allows for creating cable cuffs that take longer to break than normal
-Feeding the machine armor increases armor effectiveness and your speed in armor.
-If the machine is blessed by the chaplain, feeding the machine a cult book allows you to identify cultists by sight.
-Many, many other things are potentially feedable, potentially including traitor items.

**Traitor**
-Only available with syndicate uplink. Looks like a normal tablet.
-Grants all of the learned effects of other tablets to the reader
-Works as a strong weapon
-Feeding the machine an intellicard will allow the tablet to carry an AI on the tablet, feeding the machine an ID card gives you its access, feeding the machine a remote signaling devices allows you to remotely detonate C4, etc.
[...]

PR #2065, opened 2017-03-30
Books heal brain&stamina damage while being read

ArgyleSweatshop:
-Every 60 seconds that a person doesn't move while reading heals 5 brain damage and a decent amount of stamina damage, as well as healing a miniscule amount of external organ damage.
-Books finally have a non-rp purpose
-surprisingly much easier to code than thought
-MERGE READY
Refer to issue #1839

#### Changelog

:cl:
rscadd: when a human opens a book, initiates a loop that checks to makes sure they reading the book. then they get healed once every 60 seconds.
tweak: Multiple books can no longer be "read" at once.
/:cl:

### DIFF
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -174,6 +174,8 @@
 	var/unique = 0		//0 - Normal book, 1 - Should not be treated as normal book, unable to be copied, unable to be modified
 	var/title			//The real name of the book.
 	var/window_size = null // Specific window size for the book, i.e: "1920x1080", Size x Width
+	var/reading_time = 600
+	var/being_read = 0
 
 /obj/item/weapon/book/attack_self(mob/living/user)
 	if(is_blind(user))
@@ -181,31 +183,19 @@
 	if(ismonkey(user))
 		user << "<span class='notice'>You skim through the book but can't comprehend any of it.</span>"
 		return
+	if((winexists(user, "book")) || (being_read == 1))
+		user << "<span class='notice'>You are already reading a book!</span>"
+		return
 	if(dat)
+		being_read = 1
 		user << browse("<TT><I>Penned by [author].</I></TT> <BR>" + "[dat]", "window=book[window_size != null ? ";size=[window_size]" : ""]")
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
-		onclose(user, "book")
-		var/readingtime = 600
-		var/stimulants = 0
-		var/class = 0
-		for(var/s in list("hot_coco","coffee","tea","soy_latte","cafe_latte","pumpkin_latte","hot_ramen"))
-			if(user.reagents.has_reagent(s))
-				readingtime = 300
-				stimulants = 1
-				break
-			
-		for(var/d in list("whiskey","gintonic","whiskey_cola","irish_cream","driestmartini"))
-			if(user.reagents.has_reagent(d))
-				readingtime = max(600,readingtime + 50)	//classy reading alcohols slows down reading
-				class = 1
-				break
-		while(do_after(user,readingtime,progress = 0))	//60 seconds of reading by default
+		while(do_after(user,reading_time,progress = 1) && winexists(user, "book"))	//60 seconds of reading by default
 			user.adjustBrainLoss(-5)
 			user.adjustStaminaLoss(-20)
-			if(stimulants)
-				user.heal_organ_damage(1,1)
-			if(class && prob(20))
-				user.maxHealth += 2	//knowledge is power 20% of the time
+			user.heal_organ_damage(1,1)
+		onclose(user, "book")
+		being_read = 0
 	if(!dat)
 		user << "<span class='notice'>This book is completely blank!</span>"
 
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -191,10 +191,8 @@
 				user.adjustBrainLoss(-1)
 				user.adjustStaminaLoss(-4)
 				spawn(79)
-	else
-		user << "<span class='notice'>This book is completely blank!</span>"
-
-
+	
+		
 /obj/item/weapon/book/attackby(obj/item/I, mob/user, params)
 	if(istype(I, /obj/item/weapon/pen))
 		if(is_blind(user))
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -196,10 +196,10 @@
 			
 		for(var/d in list("whiskey","gintonic","whiskey_cola","irish_cream","driestmartini"))
 			if(user.reagents.has_reagent(d))
-				readingtime += 50	//classy reading alcohols slows down reading
+				readingtime = max(600,readingtime + 50)	//classy reading alcohols slows down reading
 				class = 1
 				break
-		while(do_after(user,600,progress = 0))	//60 seconds of reading by default
+		while(do_after(user,readingtime,progress = 0))	//60 seconds of reading by default
 			user.adjustBrainLoss(-5)
 			user.adjustStaminaLoss(-20)
 			if(stimulants)
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -185,13 +185,12 @@
 		user << browse("<TT><I>Penned by [author].</I></TT> <BR>" + "[dat]", "window=book[window_size != null ? ";size=[window_size]" : ""]")
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
 		onclose(user, "book")
-		if(istype(user,/mob/living/carbon/human))
-			spawn(200)
-			for(var/i=1;i<=10 && winexists(user,"book");i++)
-				user.adjustBrainLoss(-1)
-				user.adjustStaminaLoss(-4)
-				spawn(79)
-	
+		while(do_after(user,600,progress = 0))	//60 seconds of reading
+			user.adjustBrainLoss(-5)
+			user.adjustStaminaLoss(-20)
+			if(user.reagents.has_reagent(hot_coco) user.heal_organ_damage(1,1)
+	else
+		user << "<span class='notice'>This book is completely blank!</span>"
 		
 /obj/item/weapon/book/attackby(obj/item/I, mob/user, params)
 	if(istype(I, /obj/item/weapon/pen))
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -186,15 +186,18 @@
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
 		onclose(user, "book")
 		var/readingtime = 600
+		var/stimulants = 0
+		var/class = 0
 		for(var/s in list("hot_coco","coffee","tea","soy_latte","cafe_latte","pumpkin_latte","hot_ramen"))
 			if(user.reagents.has_reagent(s))
 				readingtime = 300
-				var/stimulants = 1
+				stimulants = 1
 				break
+			
 		for(var/d in list("whiskey","gintonic","whiskey_cola","irish_cream","driestmartini"))
 			if(user.reagents.has_reagent(d))
 				readingtime += 50	//classy reading alcohols slows down reading
-				var/class = 1
+				class = 1
 				break
 		while(do_after(user,600,progress = 0))	//60 seconds of reading by default
 			user.adjustBrainLoss(-5)
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -185,11 +185,24 @@
 		user << browse("<TT><I>Penned by [author].</I></TT> <BR>" + "[dat]", "window=book[window_size != null ? ";size=[window_size]" : ""]")
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
 		onclose(user, "book")
-		while(do_after(user,600,progress = 0))	//60 seconds of reading
+		var/readingtime = 600
+		for(var/s in list("hot_coco","coffee","tea","soy_latte","cafe_latte","pumpkin_latte","hot_ramen"))
+			if(user.reagents.has_reagent(s))
+				readingtime = 300
+				var/stimulants = 1
+				break
+		for(var/d in list("whiskey","gintonic","whiskey_cola","irish_cream","driestmartini"))
+			if(user.reagents.has_reagent(d))
+				readingtime += 50	//classy reading alcohols slows down reading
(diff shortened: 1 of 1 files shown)


Jc1469:
I can see so many ways to abuse this just with the few window exploits I know, and I don't even look for exploits.

ArgyleSweatshop:
Except you can't abuse this. You have to have the window open for a decent amount of time to receive heals and the healing is both slow and minimal.

Jc1469:
I can still abuse it, no matter how slow it is :thinking:

Unless you have a way to fix a bug that's been around forever; windows not closing.

ArgyleSweatshop:
There's a limit to the max amount of healing a book can do, and it's pretty low.


NOW I NEED SOME CODER TO HELP ME WITH THIS: 
First of all, how can I get the else that follows the whole if statement to not be totally broken

second of all, i need to use the brain damage/stamina procs, but its giving me undefined because they do not exist. Do I just need to suck it up and add another commit giving a generic proc of adjustBrainLoss to mobs that does nothing? I need advice here.

Jc1469:
I'm not hugely into coding but I imagine the issue is that the procs for brain/staminaloss don't exist for /mob, maybe try swapping it to /mob/living


Also which line is it spitting the error out for with the else.

code/modules/library/lib_items.dm
```diff
@@ -185,6 +185,13 @@
 		user << browse("<TT><I>Penned by [author].</I></TT> <BR>" + "[dat]", "window=book[window_size != null ? ";size=[window_size]" : ""]")
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
 		onclose(user, "book")
+		if(istype(user,/mob/living/carbon/human))
+			spawn(200)
```
Time-Green: This is not good. 
Try a if(do_after()), it makes sure the item and player dont move, if they do it stops it.
You can hide the progress bar by simply making the third parameter without the target = 
Otherwise make it process as long as they stand still and hold the book.

code/modules/library/lib_items.dm
```diff
@@ -185,6 +185,12 @@
 		user << browse("<TT><I>Penned by [author].</I></TT> <BR>" + "[dat]", "window=book[window_size != null ? ";size=[window_size]" : ""]")
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
 		onclose(user, "book")
+		if(istype(user,/mob/living/carbon/human))
+			spawn(200)
+			for(var/i=1;i<=10 && winexists(user,"book");i++)
```
Time-Green: This is not how for loop works in byond. Probably causing the compile error.
The braindamage procs are fine as they are inhereted by /mob/living

ArgyleSweatshop: It's actually causing no problems with compilation. By the way, editing github on mobile is nigh impossible. I'be given the ability for maintainers to edit, so you can edit this.

ArgyleSweatshop:
Yeah I quite literally cannot fix shit right now. Mobile github is cancer.

matskuman5:
What if I click the book and open the window, then drag the window to another monitor and throw the book away? I basically now get free, permanent brain damage healing.

ArgyleSweatshop:
So, you didn't read what I wrote then. The healing is capped at a max of 10 brain damage.

Of course that won't matter, since this will be replaced by doAfter once i'm on not on mobile or if time green does it for me.

matskuman5:
What about storing a book from your backpack, opening a window, waiting 20 secs, then closing the book and opening the window again?

ArgyleSweatshop:
Again, this is a pointless exercise. The code will be changed to a doAfter() loop that will prevent any of this, probably.

Alek2ander:
I suggest making every unique book only provide the benefit once, so the librarian has a reason to stock the shelves with different books and not just print 50 copies of ~~WGW~~ the first book in the list.

code/modules/library/lib_items.dm
```diff
@@ -185,7 +185,12 @@
 		user << browse("<TT><I>Penned by [author].</I></TT> <BR>" + "[dat]", "window=book[window_size != null ? ";size=[window_size]" : ""]")
 		user.visible_message("[user] opens a book titled \"[title]\" and begins reading intently.")
 		onclose(user, "book")
-	else
+		while(do_after(user,600,progress = 0))	//60 seconds of reading
+			user.adjustBrainLoss(-5)
+			user.adjustStaminaLoss(-20)
+			if(user.reagents.has_reagent("hot_coco"))
+				user.heal_organ_damage(1,1)
```
Cruix: You would have to drink a tremendous amount of hot coco for it to still be in your system after 60 seconds, you should store whether they had it in a variable before the loop.

ghost:
-20 stam damage is fairly robust 

also i was going to do something with brain damage with xthedark so shoo

ArgyleSweatshop:
Alek2sander, there wouldn't be a good way of doing that, since this process can also end by getting pushed. Therefore, any setting of a list like "books read" could easily be ruined. 

I should hope librarians would still attempt to place different books on the shelves for rp reasons. 
Cruix I'm going to make it so stimulating drinks speed up the process and I'll put it outside the loop.

ArgyleSweatshop:
Adam, consider you have to be STANDING STILL, standing up, and reading a book for 60 seconds. During that time, you would naturally heal a lot of stamina damage already. The only thing this would be good for would be after a holodeck fight or a boxing match.

ArgyleSweatshop:
Alright, guys, this is completely finished. None of the possible bonuses are strong, but it does make reading not a waste of time.

code/modules/library/lib_items.dm
```diff
@@ -185,7 +185,28 @@
...
+				readingtime = 300
+				stimulants = 1
+				break
+			
+		for(var/d in list("whiskey","gintonic","whiskey_cola","irish_cream","driestmartini"))
+			if(user.reagents.has_reagent(d))
+				readingtime += 50	//classy reading alcohols slows down reading
+				class = 1
+				break
+		while(do_after(user,600,progress = 0))	//60 seconds of reading by default
```
Cruix: You forgot to use the readingtime var

code/modules/library/lib_items.dm
```diff
@@ -185,7 +185,28 @@
...
+			user.adjustBrainLoss(-5)
+			user.adjustStaminaLoss(-20)
+			if(stimulants)
+				user.heal_organ_damage(1,1)
+			if(class && prob(20))
+				user.maxHealth += 2	//knowledge is power 20% of the time
```
Cruix: Absolutely not

ArgyleSweatshop:
Absolutely not... what? You'd have to read for 15 minutes nonstop if you wanted a bonus of 6! health. That's pretty low. I wanted reading to provide some sort of tiny bonus.  If it needs a nerf, we'll find out later.

code/modules/library/lib_items.dm
ArgyleSweatshop: yeah i dont see a problem there. that bonus is miniscule

ghost:
I don't see how you don't understand that this is disgustingly cringy.

code/modules/library/lib_items.dm
ghost: This adds health when read.

It stacks.

DO THE MATH.

ArgyleSweatshop: After an entire minute of doing nothing, with a probability of just 20%.

If you did nothing but read for an hour you'd have 24 more health. If did nothing but read for 5 minutes you'd have 2 more health. That's nothing!

ArgyleSweatshop: There, I did the math and it's still miniscule. I feel like you'd still hate this irrationally if it was set to +1 max health. That's not even factoring in that reading is slower while you are drunk.

Cruix:
Don't listen to Adam he's being an idiot.

The books should not give permanent bonuses to something as important as health, particularly not based on RNG.

There is also a problem where you can put a book in your hand and spam it to read multiple times at the same time. You need to add a variable to books that records if they are being read or not, and not let someone read it for health if it is already being read. The window should still re-open if they click it again.

It should also have a progress bar since there is no other way for the player to know when they have finished reading. You can put the progress bar over the person instead of the book by making them the target.

ghost:
@Cruix 
'don't listen to adam he's being an idiot'

*proceeds to list all of my concerns*

idiot

Cruix:
Calling something 'disgustingly cringy' is not constructive criticism and no one has any reason to pay attention to you when you make such comments, Adam.

ghost:
To be quite blunt this is the only way I can actually characterize it.

It's a ridiculous idea proposed in absolute seriousness, which therefore makes it cringe worthy. 

The concept of it is to draw attention the librarian, a traditionally useless role, which is a good idea. Unfortunately, the actual implementation of the idea plays to in-jokes and memes (the hot cocoa thing) which both has no effect and falls flat as a joke. The additional idea that reading a book will confer permanent boosts is a bad idea. Most games that have this have books 1. Be rare items that cannot be infinitely printed from the library, and 2. Are not multiplayer. What this is is essentially giving people bonuses to standing in the library and clicking a book repeatedly so the effects stack. There's no benefit to this and doesn't give the librarian anything-People will just break into the library and print the books anyway. As a result, the librarian remains as useless as a role as ever, and the 'benefits' of this are so miniscule that they won't be noted by anyone but the most extreme powergamers, who will proceed to abuse it however they can. 

Essentially, it provides powerful benefits for no downside, but it does it in such a fashion that it is boring to the casual player, who will ignore it. The people who it WILL attract are the type of people you'd want away from the game entirely, the hardcore powergamer who will vault the library counter and print the books, and if the librarian tries to kick them out, whoops, he's critted and sent to medbay. The librarian gains nothing from this, library computers can be printed from research and development anyway, it's an item that is essentially infinitely available and the benefits it provides are both irritating to obtain for anyone who would actually bother to have fun and talk with the librarian and powerful enough to be abused by the type of person who would go to the extent of doing a repetitive, tedious task for the sake of 'winning' at a game where there is no win condition aside from ruining the fun of other people.

It invites powergame, serves no benefit other than having an enormous logic gap in the context of everything else in the game, which functions on the concept of 'somewhat logical except where it is either funny, would impede with gameplay, or a better idea can be proposed.' That's not claiming SS13 aims to be realistic-it's not, there are space wizards-but it works to avoid traditional game tropes like pick up X phlebotinum for misc stat boosts. There aren't stat-boosting items in the game aside from this-the closest you could come is armor, which just logically follows in the vein of being able to take an extra laser hit because of stronger defenses. This seems to come out of the blue with 'read a book to become more durable.'

No, there's not going to be an epidemic of immortal people with books, but for a relatively cheap time investment you can still be able to take one or two extra laser hits, which is a MASSIVE change to how things work. Being able to take one extra hit can mean everything in a fight.

The tl;dr of the above is that it A. Doesn't fit with the theme of the game at all, B. Is overwhelmingly poorly balanced and in it's current form can be abused to no end, and even if modified to remove the 'click repeatedly to stack reading' results in significant balance issues while doing nothing to resolve the issue the PR was created to fix, and C. Comes out of nowhere as a shitty joke that is, yes, cringey.

But of course, i'm not going to really bother writing all that out, because the creator of a pull request won't actually -care- in most cases because they have decided what is best for the game. I could write a thesis on it and they'd still blow it off as much. There's very little point with constructive criticism to people you don't explicitly trust to take it into account, and judging by this guy's behavior and the very content of the PR itself, he's not going to take it into account. So no, there's no point to me writing a well-worded essay on why this is a bad idea when writing 'U SUCK THE COCK' is going to get the exact same amount of attention and regard.

In short, U SUCK THE COCK

Jc1469:
missed opportunity to spell out U SUCK THE COCK with capital letters in saltpost.
:-1:

ArgyleSweatshop:
First of all, salt (jk)

When I get my next chance to code, here are what I'll change. I think it'll be merge ready.

1. Addition of beingRead variable. No more clickspam!

2. Re-Addition of winExists() check. This should prevent dual book reading in most cases, since switching hands kills the doAfter () proc/check. 

3. Complete removal of the chemical aspects of code. Didn't need to exist, needless complication.

4. Movement of healOrganDamage to a base effect. If your unfamiliar with the proc, it randomly selects one external organ, without checking if it is hurt, and heals x burn/brute damage. This is pretty much the most minor way of healing, found on many juices and food items.

5. (Ninja edit) enable progress bar

As I've said before,  this will be pretty much unabuseable. There are plenty of easily found alternatives. The effects are very minor. 

What this pull request does, is give a small reward for seeking out the library. If you need to heal, a role play option other than medway is opened to you. The effects may be minor, but they are necessary. The effect is similar to how the beneficial effects of eating and drinking encourages role play with the bartender and chef. 
Of course, the librarian will get further usage when I code technical tablets (see issue #1840)

ghost:
You imply this will be merged.

Remove the permanent stat boost, as well.

ArgyleSweatshop:
First of all, cruix seems supportive of the addition; 
Second, that falls under removing the chemical parts. I guess it's sorta implied.

Dagdammit:
My thoughts from last time are unchanged:
-Opening a library-printed book give the same recuperative benefits as being in the Rest state plus slow healing of brain damage. **Benefits last until you move.**
-Also an alternate cure for Brain Rot.

Stretch goal: books whose title or contents include "woody" inflict brain & eye damage instead.

P.S: Argyle we still need to talk librarians, get on the yogs discord already. If that's not feasibly my gmail is brooks dot dagdamor.

ArgyleSweatshop:
Dagdammit, it has pretty much the exact effect as sleep. The benefits also last until you move. It'd be easy to code snowflake code like brainrot cure and "woody"  but I don't intend to. I just want this to be tested in game, first. I don't think I'll put in later, either. This is fine.

Cruix:
Upon testing, you can only start reading a particular book once, and then every time you try it says that you are already reading it. This appears to be due to a bug with the winexists() proc, and can be fixed by relogging. Try just using the being_read var and test it until it works.

Super3222:
> Try just using the being_read var and test it until it works.

This can really help your case. The codex gigas in https://github.com/yogstation13/yogstation/pull/2250 does the same thing.

X-TheDark:
I am not quite sure why you'd give books some arbitrary benefit (healing stam/brain damage). While this **might** be the way to go, I'm just not feeling it. This feels unnecessary and hamfisted to make librarians useful.

Dagdammit:
It'd be hamfisted if you tried to make it the entire solution, but small
perks (on par with eating the cook's fresh-baked food, or wearing
prescription glasses for eye damage) seem totally fine and I THINK that's
the angle this has wound up taking. Sounds like the main obstacle at this
point is finding a decent non-exploitable implementation.

Cruix:
I am not against the idea of this, but it is still bugged.

Issue #2067, opened 2017-03-31
Bath salts cannot currently be made


MacHac:
Bath salts as a chemical have been around for a while, and I always got the impression from the seven ingredients that they weren't something you were supposed to make, IE they were designed to be too much trouble.  It seems that whoever first added them all that time ago didn't bother to actually mix some, or they would have noticed that enzyme and nutriment in the same bottle produces Moonshine, rendering the recipe unmixable.

Proposal:  Replace Enzyme in the recipe with Atropine, which is similarly inconvenient to procure but won't react with anything else.  This also lowers the barrier to making it from 'sedating the chef and robbing him/her blind' to 'using the microwave for a minute and having lots of patience for mixing things', since you're no longer robbing the chef of their livelihood to go make exotic drugs.

ShadowDeath6:
why was moonshine even added in the first place

MacHac:
Because without its own illegal drugs chef is just a downgraded chemist with a microwave

ArgyleSweatshop:
close this

PR #2070, opened 2017-03-31
Bath Salts can now be made(Again?)

ArgyleSweatshop:
enyzyme was reacting with other things in the recipe, producing moonshine. I've switched out enzyme for soy sauce, because that's similarly difficult to acquire, but also acidic. Makes enough sense, right?
Refer to issue #2067

#### Changelog

:cl:
rscadd: soysauce added to bathsalts recipe, enzyme removed
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/drugs.dm
+++ b/code/modules/reagents/chemistry/recipes/drugs.dm
@@ -36,7 +36,7 @@
 	name = "bath_salts"
 	id = "bath_salts"
 	result = "bath_salts"
-	required_reagents = list("badfood" = 1, "saltpetre" = 1, "nutriment" = 1, "cleaner" = 1, "enzyme" = 1, "tea" = 1, "mercury" = 1)
+	required_reagents = list("badfood" = 1, "saltpetre" = 1, "nutriment" = 1, "cleaner" = 1, "soysauce" = 1, "tea" = 1, "mercury" = 1)
 	result_amount = 7
 	required_temp = 374
 


ShadowDeath6:
Capsaicin seems better for this, how do you even get soy sauce? Is there a soybean plant?

ArgyleSweatshop:
Yes. A bottle also spawns in the chefs office, just like enzyme.

MacHac:
As I mentioned in the original issue, I'm slightly hesitant about requiring people to steal the chef's irreplaceable possessions to make a drug, particularly a drug with significantly more drawbacks than the easier to make meth.  This goes a step further by requiring the chemist to not only climb into the kitchen, but force entry to the rear office (or just ask to have it but haha yeah right).

The fact is any recipe that requires Botany's help, ESPECIALLY if it doesn't require something botany was going to make anyway (Ambrosia Deus), is never going to be made.  Ever.  Has anyone ever actually seen someone make Nitroglycerin?  When I originally made that issue my intent was partially to make it a drug that could and would actually be made, instead of one that could stick around in the codebase for months without anyone so much as trying to mix it once.

ShadowDeath6:
Mac just wants more powergame stun immunity chems for chem, make it capsaicin

ArgyleSweatshop:
Mac, maybe the actual problem is that botany should be making different kinds of plants. Botany should be making the plants people want.

MacHac:
Bathsalts are hardly better than meth.  They're good for stun avoidance, but you can't actually do anything with them except eat stuns and wander around doing your best hobo impression.  There's few if any situations in which Meth isn't a better choice, since you don't get the wandering + hallucinations AND you get a speed boost to boot.  Thus, from a pure balance standpoint, either Meth and Ephedrine need a nerf or Bath Salts should be no more difficult than Meth to produce.

@ArgyleSweatshop At the end of the day, you run into the same issue that results in rounds with 5 roundstart miners that never bring any minerals back, or with two chemists high off their asses on Krokodil and no medicine in the smartvend.  That issue is, most people prefer to play in the way that they consider most fun, and not in the way that benefits the crew the most.  How would you propose to force botany to grow plants on request?  Make a rule that if they loaf off all round, they get a jobban?  Nerf everything's growth rate into the ground unless it's useful for something else?  There's no easy way to fix that sort of issue from a development standpoint, and that will remain the case until there's an enforced rule on the books stating 'DO YOUR JOB'.

PR #2118, opened 2017-04-09
Fixed the recycler breaking the laws of physics by pulling qdeleted items from the event horizon of nullspace

### LINKED ISSUE #2117: Slipping on empty tile
"You slipped on the mimana peel!"
"You slipped on the banana peel!"

Bunch of bananas go through the crusher fast, and **after that you slip on empty tiles** where the conveyor belt leads.

Yes, you slip on that tile (plus the tile on the right side of it, at the least). Before and after you crowbar it, and after you replace it. For a while (the effect apparently disappears after some time).
Right click shows nothing.

ss13-banana-crusher-slip

Cruix:
Fixes #2117

The list returned by GetAllContents() contains the item it was called on, so the item would be in the recycle list twice. The first time the item was eaten, it would be qdeleted, and the second time it would be moved to the crusher's loc but not qdeleted because it had already been qdeleted. No one usually noticed because qdeleting an item makes it invisible, but you can still slip on invisible banana peels.

### DIFF
--- a/code/game/machinery/recycler.dm
+++ b/code/game/machinery/recycler.dm
@@ -109,9 +109,12 @@ var/const/SAFETY_COOLDOWN = 100
 		eat(AM)
 
 /obj/machinery/recycler/proc/eat(atom/AM0, sound=TRUE)
-	var/list/to_eat = list(AM0)
+	var/list/to_eat
 	if(istype(AM0, /obj/item))
-		to_eat += AM0.GetAllContents()
+		to_eat = AM0.GetAllContents()
+	else
+		to_eat = list(AM0)
+
 	var/items_recycled = 0
 
 	for(var/i in to_eat)


ghost:
So the recycler was pulling nonexistent items from nullspace to delete them again, only to find that it couldn't delete them because they already didn't exist?

Tbh sounds like an elder god summoning ritual gone wrong

Dagdammit:
Not only that, he's saying a product of said ritual going wrong was a goddamn **antibanana**. Devoid of all physical presence, completely undetectable, it lurks in wait- still able to bring every ounce of its terrible slipping power to bear against those who walk the material plane.

Mrcreepssss:
*GAPS* THANK THE MIGHTY ADMINS FOR SLAYING THESE MONSTEROUS BANANAS!

PR #2122, opened 2017-04-11
Random maintenance room framework

Cruix:
To add a random maintenance room:
* Remove all objects in the room on the yogstation map, moving any vital objects outside into maintenance if necessary.
* Make sure there are no pipes or wires in the area that will be edited. If there are, they will be initialized twice, and they do not like that. We can fix that issue with the maploader later.
* Surround the room with walls, removing all windows and doors. This is because you can use a template to change a turf that is already there, but not edit an object that is already there. Ideally, a template will always be loaded, but if something breaks it should just be a dark, exitless empty room <strike>with a shadowling ghostrole spawner</strike>.
* Place an /obj/effect/landmark/maintroom in the lower left corner of the room, and varedit its template_names variable to a list of names of the templates it should choose between.
* Make your room template and put it in _maps/RandomRuins/MaintRuins. Use the area/template_noop area for every tile and the /turf/template_noop turf for all the walls that will not have a window or a door on them.
* Add an appropriate /datum/map_template/ruin/maint subtype to code/datums/ruins/maint.dm 

Here is the example room I made for what used to be the firefighting storage room outside primary tool storage:
img

#### Changelog

:cl:
rscadd: Added a framework for random rooms in maintenance.
tweak: The room in maintenance to the left of primary tool storage will now randomly select from a pool of one rooms. The one room it will always choose is a ghetto hydroponics room.
/:cl:


Daman453:
Make shadowling spawn points, i dare ye

Dagdammit:
Woohoo! For what it's worth, I've been drafting a design doc cataloging various maint rooms and proposing potential interiors: http://bit.ly/2oWLw5w

It's already set to allow comments; if people want, I could also give group edit permission so others could add their own suggestions.

Dagdammit:
On subject of design practices for the maint rooms, do these seem worthwhile?
-Fire safety and Biohazard closets should get moved to hall outside the room being replaced, so that players can rely on them being present every round (at least at the start).
-All potential interiors for room might want to have the same door (including door name) at the same location? It'd spoil things a little if a room's door position/lack thereof told you what was inside.

Time-Green:
Dammit, I tried to code this exact thing, but it just completely broke and exploded.
Nice job tho, this open ups alot of possibilities.

Cruix:
I would rather have the ability to change the door's variables and have to go through the trouble of changing it to the same thing in every room than being unable to change it at all.

Dagdammit:
Oh yeah, zero disagreement here- I'm not saying we should somehow handicap
what the foundation of actual game code is capable of. Just wondering if
it'd be worth trying to keep the door's appearance, location and mouseover
name consistent with other possible interiors for the same room in maint.

PR #2123, opened 2017-04-11
Reduce The Speed of Swarmers

ArgyleSweatshop:

Swarmers are both small and fast, which makes even one of them able to stand up ridiculously well against the crew, even when the crew has a large numbers advantage. Making them a little slower makes it more realistic to hit them.

#### Changelog

:cl:
tweak: Swarmers now move slightly slower.
/:cl:

### DIFF
--- a/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
+++ b/code/game/gamemodes/miniantags/bot_swarm/swarmer.dm
@@ -89,7 +89,7 @@
 	attacktext = "shocks"
 	attack_sound = 'sound/effects/EMPulse.ogg'
 	friendly = "pinches"
-	speed = 0
+	speed = 1
 	faction = list("swarmer")
 	AIStatus = AI_OFF
 	pass_flags = PASSTABLE


ShadowDeath6:
no thanks I like swarmers

ArgyleSweatshop:
swarmers are great, but it shouldn't be that a heavily armed security force struggles to deal with just one when they know where it is. This is because swarmers move so fast they are very difficult to fight with anything other than op shotguns

ShadowDeath6:
the ion rifle / emps will kill them in one shot, the point is to use weaponry that they can't get away from, this change makes it so they can be caught up to and just melee'd to death which completely ruins their usability bar extremely good jukes

Dagdammit:
I like swarmers and wholeheartedly support this change. Enemies with Speed 0 are a bad design practice; hunting them is terrible gameplay, frustrating and often downright pointless. When they die it's because they were bad or you just got lucky, not thanks to your wits or robust aim.

If speed 1 ends up nerfing swarmers too much, buff their HP by 1 laser rifle shot and cut replication costs by 25%.

Argyle, you should know: From what I'm told, speed 1 is healthy running human speed and Byond has no viable way to do anything in between. Sucks, huh?

Dagdammit:
Alternately, have them turn invisible for 2 seconds upon being hit. Or give them a reactive teleport ability that drains 20 matter per use (and won't trigger if they don't have that much)

Super3222:
Good change. Swarmer speed is bull.

Time-Green:
Just make swarmers 10 times as big, and allow them to eat humans.
Problem solved.

matskuman5:
WOW!!!!!!!!!

Dagdammit:
For the record, this could definitely be a design change that clearly goes too far without follow-up adjustments. I suggest we do a test merge, have some admin-instigated swarmers every couple rounds, just quickly gauge if said adjustments are needed.

Mrcreepssss:
YES. they are way too fast as it is RN.

"the ion rifle / emps will kill them in one shot, the point is to use weaponry that they can't get away from, this change makes it so they can be caught up to and just melee'd to death which completely ruins their usability bar extremely good jukes"
i thought their whole gig was overwhelming the crew with numbers?

ShadowDeath6:
:angry:

PR #2130, opened 2017-04-12
Fixed a bug with beams disappearing

Cruix:


Time-Green:
Are you sure this doesn't make the beams ugly?
I had to change it because it did

PR #2142, opened 2017-04-18
Visible bot paths

Cruix:
Playing as an AI is now one step closer to playing an RTS

untitled

If there is a Pai in the bot, they will see the astar calculated shortest path to the location for 30 seconds, after which they loose the all-access that they get from an AI summon.

:cl:
rscadd: AIs can now activate their diagnostic hud to see the paths that bots will take when patrolling or when summoned.
fix: Player controlled bots now lose their all access from an AI summon after 30 seconds instead of keeping it forever.
/:cl:


ShadowDeath6:
while you're at it can you remove pull verb from pAIbots if that wasn't already done
that might be out of the scope of the pr tho

Kmc2000:
hooooooooo


That looks cool

code/modules/mob/living/silicon/silicon.dm
```diff
@@ -33,17 +33,17 @@
 	var/multihud = FALSE //If you can use more than one hud at the same time
 	var/med_hud = DATA_HUD_MEDICAL_ADVANCED //Determines the med hud to use
 	var/sec_hud = DATA_HUD_SECURITY_ADVANCED //Determines the sec hud to use
-	var/d_hud = DATA_HUD_DIAGNOSTIC //There is only one kind of diag hud
+	var/d_hud = DATA_HUD_DIAGNOSTIC
```
ShadowDeath6, line 36: this kind of change always makes me laugh for some reason

Cruix: Gotta make the comments accurate or it won't compile

Cruix:
Not pulling thing can be a separate PR because my computer is already in sleep mode at this point

matskuman5:
cool!

PR #2159, opened 2017-05-01
Heavily rebalanced the most important station role

Cruix:
Updated the janitor
damp rags are now a subtype of mops because that makes more sense than being a subtype of beakers
janicarts are now a subtype of mop buckets because that makes sense
ported the advanced mop from /tg/
fixed some obscure bugs with mops handling reagents
I was going to make mops wet floors for longer, but I realized that you can just mop the floor multiple times to keep it wet longer, and having the multiplier be more than what it is could make certain reagents (more) abuseable when used in a mop.
:cl:
tweak: mops can now clean objects and people when on help intent and deal more damage, but are bulky items.
rscadd: the advanced mop now fits in backpacks and refills itself with water.
rscadd: janiborgs now have an advanced mop instead of a regular one.
fix: the janitor's custodial locator no longer tracks the location of janiborg mops.
tweak: janitors now start with all-access so they can clean more.
tweak: due to rampant abuse, janitors no longer start with all-access.
/:cl:


ghost:
>pretends to do what oak said

>does not do what oak said at all

evil

ArgyleSweatshop:
>when your compile errors fix doesn't compile
>karma for not increasing slip times

Increase the multiplier, please! If someone wanted to abuse it they could just wet the floor multiple times

matskuman5:
This has nothing to do with the PR but that "fixed compile errors" commit that caused merge errors is going on my signature

Cruix:
Compile errors are for maps I did not update

Lightning38088:
For the all access thing why not just make it to where janitors can only enter the beginning area of places except the brig,science,and the bridge. I feel like that would work since the rest of the places have nothing that could be abused (much)

Time-Green:
The janitor can just be let in. Quite often janitors ask to clean something up, and I let them in to do so

But giving civillian jobs all acces is good 👍

ShadowDeath6:
Nice CL

Penguaro:
_Don't mind me Cap, just in here to clean a bit. Oh the gun display case looks like it needs a cleaning. Nonsense don't worry about calling security but oh hey this trigger definitely needs a wipe. Oh my god! Captain, are you okay? You've turned into a pile of ash..._

PR #2162, opened 2017-05-04
More dynamic combat with hotkeys

LarissaMayrink:

Adds hotkeys for each body part so that combat(and surgery) may be done easier and faster, it will also open up the possibility of actually changing aim during combat for the extra effects(disarms, cutting arms off, extra knockdown from the legs) instead of the current head aim spam.

I hope i did this right.

Edit: I didin't

#### Changelog

:cl:
Update: Still trying to get the last errors fixed at mob_movement.
/:cl:

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -358,55 +358,63 @@
 	if(usr.client && usr.client.prefs.afreeze)
 		usr.client << "<span class='userdanger'>You are frozen by an administrator.</span>"
 		return 1
+
 	var/list/PL = params2list(params)
 	var/icon_x = text2num(PL["icon-x"])
 	var/icon_y = text2num(PL["icon-y"])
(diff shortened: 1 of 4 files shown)


ShadowDeath6:
that cl either won't work or will be super terrible in-game, just make it like

add: numpad1 = right leg
add: numpad2 = groin
and so on

LarissaMayrink:
@ShadowDeath6 It works really smoothly on tg, and i expect to work just like that on yog as it is based on tgcode.

LarissaMayrink:
Oh well, now i got why it won't, i will fix when i'm back from work, in retrospective that was a lazy way of doing it.

LarissaMayrink:
I fucked up, starting over. I will come back when my compiled version is actually working.

ShadowDeath6:
when I did that I meant the changelog, not the code itself
I can't tell if it'd work or not

Issue #2025, opened 2017-03-21
Add option/toggle to disable new sound effects


MasterPenguin5:
Or maybe just tone them down a bit. I get that it's cool and all, but some things just don't fit with it. For instance, the AI voice and announcement noises have either too much echo or way too little decay on said echo and seem to linger forever. And in some cases it's just ridiculous. I think I'm going to have nightmares from hearing the clown shoe and airhorn sounds now.

Super3222:
This can be done.

AsV9:
yes please.

Paragonid:
Is it about echo? Can we also make "Toggle Echo Effect" setting persistent across sessions?

PR #2170, opened 2017-05-06
Preserve echo effect settings across sessions

Paragonid:

Preserve echo effect settings across sessions

Increment SAVEFILE_VERSION_MAX

Connected #2025

#### Changelog

:cl:
rscadd: Preserve echo effect settings across sessions.
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -100,7 +100,7 @@ var/list/preferences_datums = list()
 	var/quiet_round = 0
 	var/purrbation = null
 
-	var/soundenv = TRUE
+	var/soundenv = 1
 
 /datum/preferences/New(client/C)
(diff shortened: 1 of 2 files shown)


code/modules/client/preferences.dm
```diff
@@ -100,7 +100,7 @@ var/list/preferences_datums = list()
 	var/quiet_round = 0
 	var/purrbation = null
 
-	var/soundenv = TRUE
```
Super3222, line 103: it's a bool var though

Paragonid: @Super3222 Bear with me, isn't it defined at this place? Didn't I change the definition from boolean to integer?

Super3222: Yes, you changed it into an integer. My question is why when because it's used as a bool var in the code.

PR #2201, opened 2017-05-17
Ports /tg/'s SMES fix, allowing them to charge if power is available even below 'input'

ghost:
https://github.com/tgstation/tgstation/pull/27237

"It's rather silly that when you set the solars SMES units to 90k, it won't charge when the solars are half on and such. What's the point in those SMES units if it's just simpler to wire the solars right into the grid."

#### Changelog

:cl:
tweak:  SMES units will now accept charge from the power network even if the available load is less than the input rate. Port from /tg/.
/:cl:

### DIFF
--- a/code/modules/power/smes.dm
+++ b/code/modules/power/smes.dm
@@ -265,9 +265,9 @@
 		input_available = terminal.surplus()
 
 		if(inputting)
-			if(input_available > 0 && input_available >= input_level)		// if there's power available, try to charge
+			if(input_available > 0		// if there's power available, try to charge
 
-				var/load = min((capacity-charge)/SMESRATE, input_level)		// charge at set rate, limited to spare capacity
+				var/load = min(min((capacity-charge)/SMESRATE, input_level), input_available)		// charge at set rate, limited to spare capacity
 
 				charge += load * SMESRATE	// increase the charge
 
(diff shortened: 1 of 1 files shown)


Daman453:
I love you

ghost:
whoops

ShadowDeath6:
radical

Cruix:
This is not a fix, it is a straight buff

ShadowDeath6:
Uhhh no dudd it's qol

Cruix:
This is not just qol, it significantly buffs every form of power generation other than the singularity and tesla.

ghost:
/tg/ called it QoL

also it's not really anything tbh

wiring to the grid is easier and probably safer, so there's no real point to having the SMES up in general in that case until now

and  you could always just toggle power input to 1 in the first place, meaning that it would charge anyway, so this really is just QoL

Cruix:
If you set the input to 1 it would barely charge, and wiring to the grid has disadvantages that storing power in the SMESs does not. This change is significant and more than just qol.

ghost:
this still charges at an absurdly slow rate, though.

ShadowDeath6:
At base upgrades a tesla and singlo both charge 800k with the engi smes, more if you wire other smes into the grid

All solars at max power (which fluctuates) can charge 360k

Jc1469:
Balance police agree this is buff.  Balance police also agree that SMES are picky as fuck so this is a QOL buff which helps lazy people who don't know how to properly set SMES.

Balance police go back to headquarters to discuss further.

Issue #2197, opened 2017-05-15
Cloning not successful if players remain in body?


00aquir:
There were some reports about this and I experienced problems with cloning too, it is suspected that the clone of a character gets "catatonic" as soon as it is ejected when the player is not a ghost but remains in the dead body, waiting.

00aquir:
Reference https://github.com/yogstation13/yogstation/issues/2045

Issue #2045, opened 2017-03-25
Cloning issues


Slinky5890:
A few times recently, ghosts have not been able to go back into their body after being cloned- this has also happened to me once. Unfortunately I can't really think of anything that tied these problems together.

The cloned body will be catatonic while the ghost needs admin intervention to get into their body.

doerakpoes:
Also if the admin helps it apears to not count you as alive as a traitor

Slinky5890:
Actually, some maybe helpful info for my situation atleast is that I was prescanned before I died

Time-Green:
Prescanning adds an implant to your body, you die, it triggers. If it gets qdelled, it also triggers but for some reason it doesn't recognize your ckey. I've tried to fix this before but failed, i'll try again some time

Slinky5890:
Turns out doesnt have to be prescanned earlier, a traitor couldn't get back in their body from a normal cloning.

PR #2205, opened 2017-05-17
Fixes the dreaded cloning bug

Time-Green:
This took me way to long.
Fixes a bug that would not clone gibbed people, and in other edge cases.
Closes: https://github.com/yogstation13/yogstation/issues/2197
Closes: https://github.com/yogstation13/yogstation/issues/2045

:cl:
fix: Fixes not getting placed in your new cloned body in certain events
/:cl:

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -342,9 +342,10 @@
 		//Don't dump us into clone if we somehow got a body during cloning finished
 		if(clonemind.current && clonemind.current.stat == DEAD)
 			clonemind.transfer_to(occupant)
-		occupant.grab_ghost()
-		occupant << "<span class='notice'><b>The world is suddenly bright \
-			and sudden and loud!</b><br>\
+		else
+			occupant.ckey = clonemind.key
+		occupant << "<span class='notice'><b>The world is sudden, bright \
+			and loud!</b><br>\
 			<i>You feel your body weight suddenly, as your mind suddenly \
 			comprehends where you are and what is going on.</i></span>"
 		occupant.flash_eyes()
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -343,9 +343,9 @@
 		if(!clonemind.current || clonemind.current.stat == DEAD)
 			clonemind.transfer_to(occupant)
 			occupant.ckey = clonemind.ckey //For some reason, this doesn't initialize directly
-		occupant << "<span class='notice'><b>The world is sudden, bright \
+		occupant << "<span class='notice'><b>The world is bright \
 			and loud!</b><br>\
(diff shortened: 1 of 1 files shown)


Time-Green:
There appears to be a problem with it forgetting antag status...

Cruix:
See if the mind gets deleted when the mob does. If it does not, you can keep a reference to it in the cloner to keep it from being garbage collected.

Time-Green:
The problem was it actually deleting the mind with it, so the reference was useless since it made a new one. The antag stuff and everything goes away with it. I am currently trying to prevent it from actually deleting the mind the way it does

Time-Green:
Alternatively...
We could make a copy of the mind datum and keep that stored. And edit some things to make it work right.
This has the fun side effect of getting completely reverted to the point you were scanned, so if you got antag after getting scanned, it would be gone. Or the other way around.

But there is the side effect of people abusing it to use one-time powers like destroy engines infinitely

Time-Green:
Or now I think of it, just use the old mind and put the player inside that, fixing all problems..

matskuman5:
ah, beautiful work time, i get so many bwoinks about bugged cloning

Time-Green:
I'll fix it tomorrow. I already have it, but it hasn't been tested

ShadowDeath6:
I think destroy engines works on a global var so other shadowlings can't use it

Time-Green:
Nah, every single one can us em
Anyway, I think I fixed it without doing the old mind thing

code/game/machinery/cloning.dm
```diff
@@ -342,9 +342,10 @@
 		//Don't dump us into clone if we somehow got a body during cloning finished
 		if(clonemind.current && clonemind.current.stat == DEAD)
 			clonemind.transfer_to(occupant)
-		occupant.grab_ghost()
-		occupant << "<span class='notice'><b>The world is suddenly bright \
-			and sudden and loud!</b><br>\
+		else
+			occupant.ckey = clonemind.key
+		occupant << "<span class='notice'><b>The world is sudden, bright \
```
ShadowDeath6: please make this less terrible, anything else that isn't this sentence will do

Time-Green: I already changed it, but I need to test it

Time-Green:
Done. It works exactly as intended now. It clones the way it's supposed to and also keeps the mind

ShadowDeath6:
Can you fix / remove the sudden and loud message you get on cloning it's so awful

"You feel dull pain as your newly formed body is dropped out of [pod], exposing it to the outside elements." Or anything that doesn't involve the word sudden

ShadowDeath6:
thanks

Cruix:
Did you fix the mind datum being deleted when the body is destroyed?

Time-Green:
No, it instead recovers the one from the initial scan

PR #2210, opened 2017-05-21
no more tesla stacking

ShadowDeath6:
tesla is way too low effort to set up considering it has easier set up than the singlo with the same or more power generated

you can still put tesla balls in separate pens wide enough for the PA

:cl: ShadowDeath6
rscadd: Tesla energy balls now destabilize and disappear when directly adjacent to one another.
/:cl:

### DIFF
--- a/code/modules/power/tesla/energy_ball.dm
+++ b/code/modules/power/tesla/energy_ball.dm
@@ -64,7 +64,8 @@ var/list/blacklisted_tesla_types = typecacheof(list(/obj/machinery/atmospherics,
 					if(ishuman(L))
 						L.emote("scream")
 					L.dust()
-				empulse(loc, 25, 40)
+				empulse(loc, 45, 70)
+				explosion(loc, 10, 20, 30)
 				qdel(EB)
 				qdel(src)
 				return
--- a/code/modules/power/tesla/energy_ball.dm
+++ b/code/modules/power/tesla/energy_ball.dm
@@ -51,6 +51,22 @@ var/list/blacklisted_tesla_types = typecacheof(list(/obj/machinery/atmospherics,
 	return ..()
 
 /obj/singularity/energy_ball/process()
+	for(var/obj/singularity/energy_ball/EB in range(3, src))
+		if(EB != src)
+			visible_message("<span class='warning'><b>[src]'s energy field contacts [EB] and destabilizes!</b></span>")
+			for(var/mob/living/L in range(5, src))
+				L << "<span class='userdanger'>Your body twitches horribly before disintegrating from the intense electric energy!</span>"
+				if(ishuman(L))
+					L.emote("scream")
+				L.dust()
+			for(var/mob/living/L in range(40, src))
+				L << 'sound/effects/supermatter.ogg'
+				L << 'sound/magic/lightningbolt.ogg'
+				L << "<i>Your skin tingles as a wave of energy passes through the air.</i>"
+			qdel(EB)
+			qdel(src)
+			return
+
 	if(!orbiting)
 		handle_energy()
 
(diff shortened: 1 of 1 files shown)


X-TheDark:
Did you test this?
I'm about 90% sure that this will cause any Tesla to destroy itself, seeing as `/obj/singularity/energy_ball` is both the type of the center tesla and the type of orbiting teslas...

According to the code, if the Tesla has any orbiting balls, it will just break itself, seeing as the orbiting balls are most definitely not the center tesla.

ShadowDeath6:
yeah I know I'm working on it, I just pr'ed it now

ShadowDeath6:
whoops I didn't read mcdonald's comment, that would have saved me like 20 minutes doing trial and error with checks

Time-Green:
Can you have the deleted tesla turn into the orbiting balls of the other tesla? Fun little feature.
Also, are you sure using a for loop with range 40 wont kill the entire server?

Alek2ander:
So, what happens when 2 super-murder-balls collide?
It kills people who are already inside containment or right next to it, so basically nobody who wouldn't have run into the tesla and died anyway. And it makes some spooky sounds.

:-1: unless this becomes something suitably violent.

Time-Green:
Alek, someone in engi hardsuit with insulated is normally tesla proof. So they can still get ashed

ShadowDeath6:
The destabilization will dust any living mob near it, unlike the bolts which won't hurt anybody that's near a grounding rod

@Alek2ander how about making a big emp when they collide

also @Time-Green the SM shard runs a for loop for every living mob

ShadowDeath6:
new commit makes an emp that's really big
it hits psych office apc and puts it at about half energy

Iamgoofball:
lame, at least make it explode

ShadowDeath6:
EMPs are more fun though

Iamgoofball:
do both

ShadowDeath6:
k

Cruix:
Make one tesla chaos dunk the other

ShadowDeath6:
This actually does work so it's r4m unless somebody wants another thing to happen

code/modules/power/tesla/energy_ball.dm
```diff
@@ -51,6 +51,25 @@ var/list/blacklisted_tesla_types = typecacheof(list(/obj/machinery/atmospherics,
 	return ..()
 
 /obj/singularity/energy_ball/process()
+	for(var/obj/singularity/energy_ball/EB in range(1, src))
+		if(EB != src && !orbiting)
+			if(!(EB in src.orbiting_balls))
+				visible_message("<span class='warning'><b>[src]'s electrical field contacts [EB] and destabilizes!</b></span>")
+				for(var/mob/living/L in range(100, src))
```
Cruix: This range is way too big, loop through the living_mob_list and check if each mob is on the same z-level and check if its distance to the tesla is less than 100 with get_dist(). 100 also seems to be a big large.

code/modules/power/tesla/energy_ball.dm
```diff
@@ -51,6 +51,25 @@ var/list/blacklisted_tesla_types = typecacheof(list(/obj/machinery/atmospherics,
...
+					L << 'sound/effects/supermatter.ogg'
+					L << 'sound/magic/lightningbolt.ogg'
+					L << "<i>Your skin tingles as a wave of energy passes through the air.</i>"
+				for(var/mob/living/L in range(5, src))
```
Cruix: The explosion is enough, don't dust people too.

ShadowDeath6: K

PR #2216, opened 2017-05-22
The Face of the Revolution

Cruix:
Made revolutionaries able to discern a head of staff from their face, but also unable to recognize each other if their faces are hidden. This is to promote some paranoia and restraint among revs.

Currently, revolutionaries can go all-in on a nonrev with no hesitation, because they know 100% that the person is not a revolutionary, and they probably will not get bwoinked if they can justify the killing. However, non-revs (particularly the heads of staff) cannot just kill people on sight, even if they have a sechud, because even an unimplanted person might not be a revolutionary and, depending on the admin, they could get in trouble for killing someone for no reason. As a result, the only thing they can really do is hide in a locker/fly off into space and wait out the revolution, if they don't have access to stun weapons and a large, competent security force.

With revolutionaries facing a similar risk, the loyal crew might stand a chance if they can hide their face and pretend to be a rev, unless someone stuns them and takes off their mask, or a headrev just tries to flash them to be sure. This also makes it a bit easier for security, since there will be fewer unknowns running around and they can actually use their sechud arrest status to keep track of things.

Left to right, the huds are:
headrev, rev, hidden, non-head-non-rev, head of staff
untitled

The hud icons seen by admins using their antag hud remains the same as it used to be.

:cl:
experiment: Revolutionaries can now recognize the faces of heads of staff, but cannot recognize other revolutionaries who are hiding their face.
/:cl:


Alek2ander:
Can we not mark targets please? Just hide the revhud if face is hidden, no need to mark everyone.
Or even remove rev sight from revs, so they only see revheads. Revheads see all revs and can coordinate or something.
Or maybe split the revolution into cells, each led by its own revhead (who know each other), so only people converted by the same revhead can identify each other, and the revheads have to pass info so their converts don't get killed by the other cells.

ShadowDeath6:
marking targets is fine since you can still use the same tactics to block it
fake moustache + an ID will work

Cruix:
IDs do not affect it at all.

You see Ivan, you find dirty capitalist not by name on ID, but by greedy capitalist gleam in eye.

ShadowDeath6:
Well you cover your face

ghost:
As far as I have always enforced it once it's rev heads can blast anyone they see if they are not restrained and don't want to implant.

Super3222:
@Cruix http://wiki2.yogstation.net/index.php?title=Revolution#The_Face_of_the_Revolution Mediocre work at the bottom.

LarissaMayrink:
@Super3222 wait until it gets merged, i'm pretty sure he is using the new icons i sent him, then you print ingame for the wiki.

monster860:
The indicator will remind them to check.

AsV9:
image
Ha ha ha

PR #2219, opened 2017-05-23
Gives phytosians lips

Time-Green:
Now you can properly use lipstick on plants
:cl: 
rscadd: gives phytosians lips by adding a define. This was worth the changelog.
/:cl:


matskuman5:
You sure this is worth rewriting dozens of pages of lore over? There's a reason that lore for this game exists and I don't think it's worth throwing that out the window for an insignificant change like this.

Close this now.

Super3222:
who coded the fly versions of every race we have?

Time-Green:
That was me, I think it was my first real project.

PR #2220, opened 2017-05-23
blobbernauts have less max health and lose health faster when off blob

ShadowDeath6:
makes blobbernauts slightly more defensive
also this fixes an "oversight" I forgot to fix when I made blobbernauts heal twice as fast since they can actually outheal the damage from losing their factory which is really dumb
also makes sorium not stun you by slamming you into walls fuck you

:cl: ShadowDeath6
tweak: Blobbernauts now die more quickly upon losing their factory.
tweak: Blobbernauts die more quickly off of the blow.
tweak: Sorium blob doesn't stun you anymore.
tweak: Lowers blobbernaut max health from 200 to 160.
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blobs/blob_mobs.dm
+++ b/code/game/gamemodes/blob/blobs/blob_mobs.dm
@@ -196,8 +196,8 @@
 	icon_state = "blobbernaut"
 	icon_living = "blobbernaut"
 	icon_dead = "blobbernaut_dead"
-	health = 140
-	maxHealth = 140
+	health = 160
+	maxHealth = 160
 	damage_coeff = list(BRUTE = 0.5, BURN = 1, TOX = 1, CLONE = 1, STAMINA = 0, OXY = 1)
 	next_move_modifier = 1.5 //slow-ass attack speed, 3 times higher than how fast the blob can attack
 	melee_damage_lower = 20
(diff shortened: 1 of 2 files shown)


LarissaMayrink:
But blobs are already so weak that i haven't seen a single blob win through this entire month

ghost:
if you have half a brain cell you can win as a blob

proof: won blob by spawning in incinerator

blob still has like a 60% winrate

ShadowDeath6:
it was an oversight that they could outheal having their factory destroyed, that should be a guaranteed death

Super3222:
i realllyy wish we had statistics to prove reasons for or against balance changes. yogstats where are you?

Super3222:
Holy shit, my mistake. I accidentally closed this

ShadowDeath6:
Accidentally

Jc1469:
Reducing max health and also increasing degen rate :thinking: 
Seems a bit overkill to m- 
**THE BALANCE POLICE DO NOT APPROVE OF THIS CHANGE.**
**LESS HEALTH AND DIE FASTER? WHAT.   JUST DO ONE OR THE OTHER.** 

>makes blobbernauts slightly more defensive

**MAKES BLOBBERNAUT FUCKING DIE IF IT LEAVES BLOB TILES**

ShadowDeath6:
@Jc1469 they can just run at people and facetank 5 lasers off blob tiles then run back and heal 10 health per blob movement, which is actually more than per blob pulse since they get healed by the overmind creating tiles near them

PR #2225, opened 2017-05-26
Adds asay and dsay hotkeys

Time-Green:
**Requested by Slinky**
Hotkey , = asay
Hotkey . = dsay
asay and dsay hotkeys
:cl:
add: Adds asay and dsay hoteys, "," and "."respectively.
/:cl:

### DIFF
--- a/code/modules/admin/verbs/adminsay.dm
+++ b/code/modules/admin/verbs/adminsay.dm
@@ -21,3 +21,9 @@
 
 	feedback_add_details("admin_verb","M") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
+/client/verb/asay_popup(message as text)
+	set name = "asay"
+	set category = "Admin"
+	set hidden = 1
+	src.cmd_admin_say(message)
+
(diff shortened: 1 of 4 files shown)


Groudonmaster:
:+1:

Alek2ander:
Please change from "." to something else, "." is radio channel select in normal speech and will be constantly pressed by mistake

Time-Green:
Tell me wich keys to change it to, and I'll do it

Cruix:
Why do we need these hotkeys for admin-only commands? I have never had a problem typing asay or dsay into the command line even when sending multiple messages.

Time-Green:
I don't know either, but it was like no 1. requested by staff members

Groudonmaster:
QoL cruix. Qol.

Jc1469:
staff are just bad at typing and only know how to type ooc and accidentally type in ooc, so they want a hotkey to fail to press and hit the ooc hotkey instead.

qol and all that crap sure, a bit pointless though in my eyes, and will probably end up being more trouble than you'd think

PR #2230, opened 2017-05-27
fixes bug where borers didn't lose control from taking sugar

ShadowDeath6:
found this bug like 3 weeks ago and forgot to fix it lul


code/modules/mob/living/simple_animal/borer/borer.dm
```diff
@@ -148,7 +148,6 @@ var/banned_borer_emotes = list("*collapse", "*collapses", "*surrender", "*surren
 		for(var/M in banned_borer_emotes)
 			if(M == message)
 				src << "<span class='warning'>You don't have the strength to do this...</span>"
-			//	src << "Nice try, asshole."
```
Super3222, line 151: lul

PR #2234, opened 2017-05-28
Unpowered mechs can't bumpsmash

Super3222:
#### Changelog

:cl:
fix: Unpowered mechs can no longer bumpsmash.
/:cl:

### DIFF
--- a/code/game/mecha/working/ripley.dm
+++ b/code/game/mecha/working/ripley.dm
@@ -33,14 +33,6 @@
 	cargo.Cut()
 	return ..()
 
-/obj/mecha/working/ripley/HasProximity(atom/movable/AM as mob|obj)
-	if (istype(AM, /obj/item/weapon/ore))
-		if((locate(/obj/item/mecha_parts/mecha_equipment/hydraulic_clamp) in equipment))
-			var/obj/structure/ore_box/ore_box = locate(/obj/structure/ore_box) in cargo
-			var/obj/item/weapon/ore/O = AM
-			if(ore_box)
-				O.Move(ore_box)
-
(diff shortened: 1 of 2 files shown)


code/game/mecha/working/ripley.dm
```diff
@@ -33,6 +33,14 @@
 	cargo.Cut()
 	return ..()
 
+/obj/mecha/working/ripley/HasProximity(atom/movable/AM as mob|obj)
+	if (istype(AM, /obj/item/weapon/ore))
+		if(. && (locate(/obj/item/mecha_parts/mecha_equipment/hydraulic_clamp) in equipment))
```
Cruix: . will always be null

drderp3635:
but they already did

code/game/mecha/working/ripley.dm
```diff
@@ -33,6 +33,14 @@
 	cargo.Cut()
 	return ..()
 
+/obj/mecha/working/ripley/HasProximity(atom/movable/AM as mob|obj)
```
X-TheDark: as mob|obj is not necessary
Although I'm not sure if this is something important or not

X-TheDark:
If this is supposed to fix **"Very Fast Gygax Running at Incredible Hihg Speed Permanently Bump-Smashing Stuff"**, then this won't fix anything, because it's the `get_charge()` proc that causes this. 

SPECIFICALLY (as in, intended behavior...blame /tg/), it will return a positive number if you have a tesla relay installed, even if you have no battery charge, which, incidentally, is also the cause of Gygax being able to Overload permanently.

Super3222:
@X-TheDark Wow!

PR #2241, opened 2017-05-30
medical healing stuff balancing

ShadowDeath6:
apparently patches were supposed to have a delay but there's some weird thing that I can't figure out where both pills AND patches will either always have the same delay or never have a delay and if someone could tell me how to fix that I would appreciate it because it'd make the separation of the two much more fine

also the stacking stuff like bruise packs and ointment don't have a self-app delay because nobody uses those unless they don't want to waste patches, but since it's isolated to one limb and it's very finite and burst-healing, I doubt that it could be effectively abused as much as patches are

:cl: ShadowDeath6
tweak: Bruise packs and ointment no longer have a delay.
tweak: Pills and patches now have a delay- 3 seconds to apply.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -106,7 +106,7 @@
 	icon_state = "brutepack"
 	heal_brute = 40
 	origin_tech = "biotech=2"
-	self_delay = 20
+	self_delay = 0
 
 /obj/item/stack/medical/gauze
 	name = "medical gauze"
@@ -136,4 +136,4 @@
 	icon_state = "ointment"
 	heal_burn = 40
 	origin_tech = "biotech=2"
-	self_delay = 20
+	self_delay = 0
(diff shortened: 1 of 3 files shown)


LarissaMayrink:
this is actually pretty good, i don't know why this never got ported to yog

ShadowDeath6:
it was ported a long time ago, it's just bugged
probably bugged on tg as well

YogsMcree:
We are getting body part hotkeys. Ointments and bruise packs are going to be abused

Super3222:
Not saying I don't agree with this PR, but these changes will make MD"s even more useless. Part of the reason Eph made a delay on ointment and bruise packs was so that having someone else heal you (most likely an MD) would be the faster option, thus given them some sort of use. Especially when sleepers are down / they are outside of medbay. Just something to point out.

ShadowDeath6:
Bruise packs and ointment are still finite, and only heal one part at a time, doctors using sleepers was and still is the superior option

PR #2257, opened 2017-06-03
REDUCE SHADOWLING POWERGAMING

ghost:
:cl:
tweak: If an unhatched shadowling is given a mindshield implant, it will no longer show the resisting message
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/implants/implant_loyality.dm
+++ b/code/game/objects/items/weapons/implants/implant_loyality.dm
@@ -22,9 +22,13 @@
 		var/obj/item/weapon/implant/mindslave/imp = locate(/obj/item/weapon/implant/mindslave) in target
 		if(imp)
 			imp.removed(target)
-		if((target.mind in (ticker.mode.head_revolutionaries | ticker.mode.get_gang_bosses())) || is_shadow_or_thrall(target))
+		if((target.mind in (ticker.mode.head_revolutionaries | ticker.mode.get_gang_bosses())) || is_thrall(target))
 			target.visible_message("<span class='warning'>[target] seems to resist the implant!</span>", "<span class='warning'>You feel something interfering with your mental conditioning, but you resist it!</span>")
 			removed(target, 1)
+			qdel(src)
+			return -1
+		if((target.mind in (is_shadow(target)))
+			removed(target, 1) //be silent about it and reduce implant powergaming with slings
 			qdel(src)
 			return -1
 		if(target.mind in ticker.mode.revolutionaries | ticker.mode.get_gangsters())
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
1. don't get caught before you hatch as shadowling
2. this doesn't change anything because it still won't show up as them being implanted
3. git gud

ghost:
1. well if you get caught removing power cells... game over.  
2. might fool some.  
3. reee powergamers

code/game/objects/items/weapons/implants/implant_loyality.dm
```diff
@@ -22,11 +22,15 @@
 		var/obj/item/weapon/implant/mindslave/imp = locate(/obj/item/weapon/implant/mindslave) in target
 		if(imp)
 			imp.removed(target)
-		if((target.mind in (ticker.mode.head_revolutionaries | ticker.mode.get_gang_bosses())) || is_shadow_or_thrall(target))
+		if((target.mind in (ticker.mode.head_revolutionaries | ticker.mode.get_gang_bosses())) || is_thrall(target))
 			target.visible_message("<span class='warning'>[target] seems to resist the implant!</span>", "<span class='warning'>You feel something interfering with your mental conditioning, but you resist it!</span>")
 			removed(target, 1)
 			qdel(src)
 			return -1
+		if((target.mind in (is_shadow(target)))
```
ShadowDeath6: `in` is only used for checking if they're in `ticker`, you just want to use `(is_shadow(target))` by itself

Super3222:
I don't like this.

I think shadowlings are in a good spot right now- as far as implanting goes. I feel like this PR is a knee-jerk reaction to a bad round.

The basis of the mode is that the shadowlings quickly gather whatever they need and find somewhere in maintenance to hatch. This PR gives leeway to that and removes the drawback of running around and acting like a traitor.

And I know - no sane HoS or Warden is going to be implanting people that come into the brig under normal circumstances. But when hell breaks loose, and people start declaring slings they _will_ be implanting people that break the law. I don't think it's fair to remove this for the purpose of thwarting powergamers when it can be useful to normal players. If someone is breaking the rules, and using mindshield implants incorrectly you can always ahelp.

Slinky5890:
Yeah i'm against this, agree entirely with super

ShadowDeath6:
also this change won't stop powergamers because they're going to say "where's the HUD" and know confirmed shadowlings because it's snowflake

Jc1469:
balance police believe shadowling to be overpowered if played correctly, and shit if played badly.
balance police do not believe in hand-holding bad players.

balance police believe OP should "git good"

PR #2263, opened 2017-06-05
Adds jaunter beacons to survival capsule fridge

Super3222:
#### Changelog

:cl:
rscadd: Undeployed jaunter beacons and wrenches spawn in the survival capsule vendor.
/:cl:


X-TheDark:
Eh, why not?

PR #2265, opened 2017-06-06
makes zombies decidedly not so incredibly powerful

ghost:
as to encourage admins to spawn zombies in the future, one of the main issues with zombies is that one hit from a claw = death in a way that the previous one, biting (which was curable) does not. the 25 dmg melee hit from a zombie, on one hit, introduces an organ that poisons you over time. your death is assured.

now that might seem thematically appropriate, but hold on here. one, it's unfun-zombies steamroll because of it. one hit is death, and that's the end of that. plus, the organ heals them over time very rapidly, making them difficult to fight at all.

what this does is make it not poison humans and not heal zombies. you still turn if you die when you have the organ in you-but you don't die from having the organ.

#### Changelog

:cl:
rscadd: Zombie hits are no longer absolutely guaranteed death.
/:cl:

### DIFF
--- a/code/modules/zombie/organs.dm
+++ b/code/modules/zombie/organs.dm
@@ -18,18 +18,6 @@
 	zombie_infection_list -= src
 	. = ..()
 
-/obj/item/organ/body_egg/zombie_infection/on_life() //heals the zombie a la legion soul, poisons non zombies
-	..()
-	if(owner)
-		if(iszombie(owner))
-			owner.adjustBruteLoss(-2)
-			owner.adjustFireLoss(-2)
-		else
-			if(prob(45))
-				if(prob(20))
-					owner <<"<span class='danger'>You feel sick.</span>"
-				owner.adjustToxLoss(3)
-
 /obj/item/organ/body_egg/zombie_infection/on_find(mob/living/finder)
 	finder << "<span class='warning'>Inside the head is a disgusting black \
 		web of pus and vicera, bound tightly around the brain like some \


ShadowDeath6:
bad pr, zombies are still completely vulnerable to stuns and have no range. No healing just encourages them to just attack themselves to death which is stupid and shitty, take a look at tg's zombie buff pr, it heals them for over twice what our zombies heal.

 I'd probably increase the message rate to show that something is wrong with you when you have the organ, so people feel a pressing need to remove it but don't die in 3 minutes. There are a lot of ways to tweak it without making zombies complete shit

I'm against the "lol improve don't remove" meme but you're starting to make me change my mind

LarissaMayrink:
Cut the poison to humans and keep the heal to zombies and you got a way to go, they won't steamroll for getting one hit but they won't get mauled by stuns that can be massproduced, which will become the meta for zombiefighting as soon as it starts to be spawned more often.

ghost:
@ShadowDeath6 theyr'e supposed to be  vulnerable to stuns and have no range, they're zombies.

they can come back to life anyway. making them facerolling tanks is a bit silly.

if I could, I would make them immune to literally everything but being hit in the head, but I don't know how to do that, and it would probably contribute to, well facerolling tanks.

ShadowDeath6:
facerolling tanks is a bit overkill considering they have effective 20% damage resist and the only way for them to get damage resist is through having armor when they die, which unless the people who kill them are done, will result in them being cuffed or decapitated so they can't do anything

Super3222:
> I would make them immune to literally everything but being hit in the head

That actually sounds pretty interesting. How about doubling damage to the head in addition?

Lightning38088:
I don't know bout about code much, but couldn't you make them grow sorta "armor" on their body except their head and make these armor's immune to damage and couldn't be taken off?  

Sorta like how the sling has it's armor for effect but for the zombie it wouldn't really change the skin it'd just be invisible armor.

ghost:
i could do that but it'd be gimmicky

ShadowDeath6:
afaik armor can't be stronger than 80% melee resist

ghost:
it can

plus worst case scenario i could just add a really gimmicky check that nulls the damage

oisin100:
Make them turn to ashes if burnt to death(Although not lasered).

Super3222:
I'd prefer it if zombies were stun/stamina immune, didn't have the passive healing, BUT could heal off of clawing dead bodies (with a limit attached).

Cruix:
If zombies lose both their healing and their one-shot-I-won attacks, they will be utterly laughable threats and the first zombie will probably just get toolboxed to death by a bored assistant.

PR #2272, opened 2017-06-08
Boomerang & Frisbee

Time-Green:
# Adds a boomerang and a frisbee to the gift vendor.
### **Boomerang**
You throw it and it comes back, most code was stolen from cruix's bananium shield
### **Frisbee**
It travels twice the distance you throw it at.

:cl:
rscadd: Adds a frisbee to the gift vendor
rscadd: Adds a boomerang to the gift vendor
/:cl:
I was told by a spriter to sprite it myself.
naamloos
:sprites needed:
I also stopped using verbs in my PR titles


code/game/objects/items/toys.dm
```diff
@@ -1415,4 +1415,41 @@
...
+	desc = "Comes further in life then you."
+	icon_state = "frisbee"
+	icon = 'icons/obj/toy.dmi'
+	force = 0
+	throw_speed = 0.5
+	throw_range = 14
+
+/obj/item/toy/frisbee/throw_at(atom/target, range, speed, mob/thrower, spin=0)
+	var/new_x = target.x*2 - thrower.x  //I created this formula myself, it's beautiful
+	var/new_y = target.y*2 - thrower.y
```
Cruix: Do you know what this is going to do?

Time-Green: Create a new variable with the new coordinate?

Jc1469: are target coordinates relative or actual station coordinates.

Cruix: It is going to make the Frisbee fly out of bounds if you throw it at the edge of the map

Time-Green: For jc:
They are the actual coordinates, not relative.

Way it works is, it takes the difference between the two, and then add that to the target loc
target.x - thrower.x would be relative.
To make it accurate, you'd need to add the targets coordinates. And instead of using target + target - thrower, I just made it target*2 - thrower since that looks cooler

To cruix: shit, didnt think of that, I assume thats a problem.
I'll go and test that tomorrow

Jc1469: friend is drunk and bad at math and was yelling at me about something dw bout it

code/game/objects/items/toys.dm
```diff
@@ -1415,4 +1415,41 @@
...
+	icon = 'icons/obj/toy.dmi'
+	force = 0
+	throw_speed = 0.5
+	throw_range = 10
+
+/obj/item/toy/boomerang/throw_impact(atom/hit_atom)
+	if(iscarbon(src.loc)) //Did someone catch it?
+		..()
+	throw_at(thrownby, throw_range+3, throw_speed, null)
+	..()
```
Alek2ander: Is it supposed to call parent proc twice?

code/game/objects/items/toys.dm
Alek2ander: You should use get_turf just in case, target atom might have weird coords because that always happens for some reason. And yeah, needs to be clamped to valid x/y values, or target will be null.

Also, totally spin=1 for both of them.

Alek2ander (changes requested):

code/game/objects/items/toys.dm
Time-Green: Oh, haha, I forgot the make it return ..()
Nah, not intended

Time-Green: I cant actually comment on the other comment, so ill do it here.

Turf one is a good idea, and the clamp thing was told by jc of all people
>Also, totally spin=1 for both of them.

Thats not how frisbee works

Frostben:
totally make a syndie version of these that can chop off a limb when you throw it at someone.

Time-Green:
I used clamping and it works.
I tested it, and throwing at the edge just makes it land at the edge, instead of falling right under you

Alek2ander (approved):

drderp3635:
Add a syndicate boomerang that comes with a free safari hat, that does a knockdown/short stun when thrown at legs or head respectively, and 3 times more damage to carps

Time-Green:
Drderp, I thought about the concept of adding a traitor weapon, then I realized how disgustingly OP that would be. I really like the idea, but it would need a serious downside for it to not be death on a boomerang

Nickvr628:
It has a downside. You have to pick it up every time or else you lose it.

5 TC would be a fair price.

Time-Green:
Not really, the point of a boomerang is that it comes back, and you automagicly catch it

Frostben:
How about the boomerang has a certain range when you throw it before it flies back to you. If it hits someone however then it won't come back to you because hey it just hit a solid object.

PR #2190, opened 2017-05-14
removes multiplying dragons

ghost:
Removes ash drake multiplication

#### Changelog
Ash drake attack chance of creating more dragons is now based off how pissed they are. The lower health, the more often they spawn dragons.

#### Changelog

:cl:
rscadd: Ash drakes no longer undergo mitosis
/:cl:

PR #2276, opened 2017-06-09
Revert "Merge pull request #2190 from AdamElTablawy/patch-90" aka the one that touched my baby

Jc1469:
This reverts commit a986ca52fb4cebe1934b223d23825da1db8dbedd, reversing
changes made to 888a8808d3599b224f0f1de9bad3fb6c59fbd6f5.

Let's just say Adam didn't fix the problem, he ignored it.

#### Changelog

:cl:
tweak: Ash drake duplication is back, but lessened...  because lesser drakes.. you get it? No?  Aight.  Anyways, incase you didn't know, lesser drakes are essentially 1/5th of an actual ash drake.  And now the lesser drakes themselves can't duplicate anymore.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/dragon.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/dragon.dm
@@ -58,11 +58,11 @@
 	if(rawScaling >= 100 && rawScaling <= 115)
 		var/prefix = "fiery"
 		name = "[prefix] [name]"
-		loot += /obj/item/weapon/ore/diamond
 		return
 	if(rawScaling > 115 && rawScaling < 130)
 		var/prefix = "blazing"
 		name = "[prefix] [name]"
+		loot += /obj/item/weapon/ore/diamond
 		return
 	if(rawScaling >= 130)
 		var/prefix = "infernal"
@@ -158,6 +158,8 @@
 			swoop_attack()
 			swoop_attack()
 			swoop_attack()
+	else if(prob(5) && !istype(src,/mob/living/simple_animal/hostile/megafauna/dragon/lesser))
+		xoxo_and_friends()
 	else
 		fire_walls()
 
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
meh, never really liked the idea of having to fight two megafauna, the lesser ones just serve as an obstacle rather than an extra fight to the real drake

Jc1469:
perhaps, but it also drops a fifth of what a normal drake drops.  It's not something for nothing, and it also isn't aggroed to you when it spawns.  That and they're now also going to spawn a lot less.

Cruix:
But can a normal dragon split more than once?

Also don't use istype() on src, override the lesser drake's xoxo_and_friends() proc to do nothing.

Jc1469:
Drakes don't split, they call in reinforcements, and if I do that, admins can't even proc call a lesser drake to use the ability.

ShadowDeath6:
If an admin wants to spawn another lesser drake they can just spawn one in, they don't need to call a proc on an existing lesser drake, that would be really inefficient and annoying to do.

Cruix:
A subtype overriding its own procs instead of the base type checking its own type when it tries to do something is part of object-oriented design, which is the codebase standard.

Jc1469:
Lesser drakes attack slower.
Lesser drakes will still have that attack in their possible moves, doing nothing.
That's a chunk of time where a drake will just not be doing anything.

Cruix:
Then make the proc return it's success or failure and have it do something else on failure, make the moves it could take and their probabilities into a list and set the lesser drake's probability of calling reinforcements zero, or override the OpenFire() proc on lesser drakes instead so that the reinforcement proc is never called.

X-TheDark:
Why would you need to be able to call the xoxo proc manually via admin commands on a lesser drake, when as Shadow pointed out, you can just spawn the damn things?

Cruix:
Do it with proc overrides

PR #2283, opened 2017-06-11
Added shuttle navigation computers for the syndicate shuttle and assault pod

Cruix:
The syndicate shuttle now has a navigation computer right next to the shuttle computer that lets you target any area on the station z-level that does not have any part of the shuttle intersect with a non-space turf or a shuttle docking turf. The person targeting the location gets a meson view of the station.

shuttle

The assault pod gets a similar interface, but it can target non-space turfs. It might need a slight nerf if people start using it in creative ways.

:cl:
rscadd: The syndicate shuttle and assault pod can now have their exact landing locations set by camera.
/:cl:

### DIFF
--- a/_maps/map_files/Yogstation/yogstation.2.1.3.dmm
+++ b/_maps/map_files/Yogstation/yogstation.2.1.3.dmm
@@ -68488,6 +68488,12 @@
 	},
 /turf/open/floor/plasteel,
 /area/tcommsat/computer)
+"DYZ" = (
+/obj/machinery/computer/camera_advanced/shuttle_docker/syndicate,
+/turf/open/floor/plasteel/shuttle{
+	icon_state = "shuttlefloor4"
+	},
+/area/shuttle/syndicate)
 "FCq" = (
 /obj/structure/table,
 /obj/item/weapon/stock_parts/subspace/ansible,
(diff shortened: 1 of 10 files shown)


ShadowDeath6:
crashes into the AI core

ghost:
the pod can already do that and the ship, detailed above, cannot

LarissaMayrink:
Crashes into the captain

Slinky5890:
@LarissaMayrink It's a meson view, so they won't see people

LarissaMayrink:
@Slinky5890 you still need to send someone on the shuttle earlier, so, whoever goes on it can keep an eye on the captain, or you can just watch the radio for when the cap gets called somewhere. Yes, it does need at least a little coordenation, but still, this can end up being just too strong, yet, its the best idea i've seen to deal with the previsible shuttle problem.

I guess it needs to be run before screaming about how strong it is.

ShadowDeath6:
so you need to spend 30 tc on the pod, then send someone in for recon before having the captain stand still long enough for someone to be able to slam the pod into him

LarissaMayrink:
You don't really need to hit the pod into him to get him killed as the pod got a few good turrets, plus a mob of angry syndies inside and the surprise factor, but yes, unless it becomes the rule for every other nuke round end up like that it won't be a problem. Its good, really good.

Mrcreepssss:
oooooohhhh i like this.

PR #2288, opened 2017-06-12
Fixes admin messages/logs for singularities

### LINKED ISSUE #2266: Remove the mini singulos in the wizard academy
Not only do these give false flags on the admin log of a loose singulo, but it also makes finding the real loose singularity harder because i have to scroll through 8 different singularities to locate it

X-TheDark:
fixes #2266

Also renames academy singularities, so it's easier to jump to the actual engineering one.

:cl:
tweak: Wizard Academy singulos will no longer be logged or bother admins.
tweak: Wizard Academy singulos renamed to 'magical gravitational singularity' for ease of finding the relevant ones.
/:cl:

### DIFF
--- a/code/modules/awaymissions/mission_code/Academy.dm
+++ b/code/modules/awaymissions/mission_code/Academy.dm
@@ -31,12 +31,11 @@
 //Academy Items
 
 /obj/singularity/academy
+	name = "magical gravitational singularity"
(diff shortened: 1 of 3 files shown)


Time-Green:
Do a changelog anyway, they are always good

PR #2297, opened 2017-06-16
Fixed being able to reset an emagged arcade game by using a screwdriver on it

Cruix:


ghost:
That was intentional.

ShadowDeath6:
there's no way to prevent it, they can just crowbar

Edit- nevermind I didn't read the code

PR #2303, opened 2017-06-18
Removes the permanent confuse from holy water

ghost:
For some reason, this literally set the mob to confused. Permanently.

I've removed that, since dizzy seemed to do it well enough.
#### Changelog

:cl:
rscadd: Holy water will now no longer permanently confuse crew in quantities over 30u.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -189,9 +189,6 @@
 				if("emote")
 					M.visible_message("<span class='warning'>[M] [pick("whimpers quietly", "shivers as though cold", "glances around in paranoia")].</span>")
 	if(data >= 75)	// 30 units, 135 seconds
-		if (!M.confused)
-			M.confused = 1
-		M.confused += 3
 		if(iscultist(M) || is_handofgod_cultist(M) || is_handofgod_prophet(M) || is_servant_of_ratvar(M))
 			if(iscultist(M))
 				ticker.mode.remove_cultist(M.mind, 1, 1)


ShadowDeath6:
it's because there was no upper limit to how confused you could be so you would have like confused = 120 upon deconvert

Cruix:
That is not how confused works. Like most status effects, it goes down by 1 every tick. Setting it to one makes them confused for 1 second, not permanently.

ghost:
But that's not what happened in effect.

They were stuck confused.

Cruix:
Did you check their confused var?

ghost:
Yes.

Confused was at 1. 

forever.

Adminheal didn't fix.

Cruix:
I cannot reproduce this on a test server or on the live server. Even if I could, this would not fix the bug, as there are many different things that affect the confused var the same way holy water does.

Time-Green:
It might not be infinite, but the confusion you get from it COMPLETELY randomizes all movement and it's absolutely horrible to live with it

LarissaMayrink:
@Cruix it seens to only happen some times, but every now and then during cult i find deconverted people who are stuck on confused after being feed 60+ units of holy water, which sticks for the entire round long after the holy water has been purged from their systems. So, i guess you should test it during deconversion, not just regular mobs as it might be the deconversion interation that loopholes it into eternity.

code/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -189,9 +189,6 @@
 				if("emote")
 					M.visible_message("<span class='warning'>[M] [pick("whimpers quietly", "shivers as though cold", "glances around in paranoia")].</span>")
 	if(data >= 75)	// 30 units, 135 seconds
-		if (!M.confused)
-			M.confused = 1
-		M.confused += 3
```
ShadowDeath6, line 194: remove this line only

PR #2325, opened 2017-06-24
[s] Webhooks

oisin100:
Removes ByondPOST dll and instead uses the inbuilt byond http GET ability to send messages to a webserver which can then do what it likes with the data such as sending OOC and asay messages.

Fixes the shitton of runtime errors the byondpostdll is creating and hopefully also fixes the discord bot.

Ill run this for a test merge in a few days as last time i attempted to integrate a system like this it wouldn't work, I think that's because I didn't realise byond doesn't support HTTPS.

The webserver which byond will communicate with can be found here: [https://github.com/yogstation13/yogbot](https://github.com/yogstation13/yogbot)

Hopefully it also fixes the blackbox data not being saved to the database on some rounds which has basically fucked up yogstats.


Amelia0010:
image
👌

PR #2329, opened 2017-06-26
Gives the warden his shotgun back and contraband locker modifications

LarissaMayrink:
Gives the warden his shotgun, he is very sad without it.
:cl:
rscadd:
Adds a cycler shotgun to the warden locker
Adds a contraband spawner on the contraband locker: That means that now the contraband locker starts with a random item from a list.
/:cl:


Kmc2000:
Ehhh i don't think this is needed tbh warden has the entire aresnal of the united states military already and a tazer. As for contraband starting in the contraband locker. Lol

Partheo:
Seems unecessary to me. As KMC said, the warden already has the whole armory, plus he already gets a unique weapon- the gloves. (and EMP grenades)

In my opinion, a role that is supposed to spend its entire time in the brig, processing prisoners, and assisting security should not get a special shotgun. The warden doesn't need a shotgun for his job, so giving it back to him is overkill. Plus, the gloves are already more than adequate for dealing with rowdy prisoners in a non-lethal manner.

As for the contraband, sure.

LarissaMayrink:
The armoury is, as the name implies; The armoury, which is exposed, very exposed, and anyone with a c4 can open a whole on it(which will suck out most of the guns) in less time than it takes for a AI realize it during highpop, let alone the time it takes to get at least the magboots on, which is why it ends up being so easy to raid the armoury as you don't have enough time to arm up before there is a hole sucking everything out.

The cycler gives the warden a way of actually having a chance at defending the armoury after the c4/x4 has been placed(and, if you are fast enough, after it has blown up)

Plus, it is a combat shotgun level gun for someone who is supposed to defend the big unsafe storage of guns with a giant target on it, and the addition won't affect the rest of the station at all during most rounds as it won't leave the brig, let alone be carried out blazing shells.

Slinky5890:
i'm totally against this for the same reason as the rest. He already has a pretty strong custom weapon, don't see why we need to add something like this apart from to make admin's lives harder. "Why did you shoot him with that shotgun?" "I spawned with one so it's fine right?"

LarissaMayrink:
@Slinky5890 eh, the gloves might be strong if you ever get in a place where its better to use them than to shoot, which is pretty much never as he is the dude who guards the brig, and, guess what, when someone attacks the brig, or manages to robust a officer inside perma said person won't be unarmed.

About the shooting thing; Isn't that the same that happens with any other gun? and that the warden already pulls something out of the armoury for himself anyway because of his lack of a weapon?  I don't see how this is different from someone being shot with any other gun on lethal, and, if shot on beanbag its kind of irrelevant as it does roughly the same damage as the detective pistol from a distance

LarissaMayrink:
This probably won't get anywhere, for now.


I will be back.

PR #2331, opened 2017-06-26
Reduces brain damage caused by adminhelp.

Jc1469:
Reduces brain damage caused by using ahelp.  

This is a balance change and nothing more.  Ahelp should not be causing brain damage.

### DIFF
--- a/code/modules/admin/tickets/admin_ticket_verbs.dm
+++ b/code/modules/admin/tickets/admin_ticket_verbs.dm
@@ -2,13 +2,24 @@
 /client/verb/admin_ticket(ticket_title as text)
 	set category = "Admin"
 	set name = "Adminhelp"
+	
+	var/obscurevariablebeingusedforthis = alert("Which do you wish to use?", "Admin-help Redirect", "Adminhelp", "Wiki", "Bug Report")
+	if(!obscurevariablebeingusedforthis)
+		usr << "<span class='ticket-status'>Please select one of the options to receive help.</span>"
+		return
 
+	if(obscurevariablebeingusedforthis == "Wiki")
+		usr << "<span class='ticket-status'>Here is a link to the TGStation wiki: https://tgstation13.org/wiki/Main_Page</span>"
+		return
+	if(obscurevariablebeingusedforthis == "Bug Report")
+		usr << "<span class='ticket-status'>Here is a link to our github issues page where you can report bugs.  Please check if a report exists about what you are reporting first. https://github.com/yogstation13/yogstation/issues</span>"
+		return
 	if(prefs.muted & MUTE_ADMINHELP)
 		src << "<font color='red'>Error: Admin-PM: You are unable to use admin PM-s (muted).</font>"
 		return
 
 	if(!ticket_title)
-		usr << "<span class='ticket-status'>You did not supply a message for your ticket. Ignoring your request.</span>"
+		usr << "<span class='ticket-status'>You did not supply a message for your ticket. A ticket has not been made.</span>"
 		return
 
 	ticket_title = replacetext(ticket_title, "'", "\'")


MayhemSailor:
All of this.

code/modules/admin/tickets/admin_ticket_verbs.dm
```diff
@@ -2,13 +2,24 @@
 /client/verb/admin_ticket(ticket_title as text)
 	set category = "Admin"
 	set name = "Adminhelp"
+	
+	var/obscurevariablebeingusedforthis = alert("Which do you wish to use?", "Admin-help Redirect", "Adminhelp", "Wiki", "Bug Report")
```
AshCorr, line 6: I hate you.

MayhemSailor: It helps with player induced brain damage...

code/modules/admin/tickets/admin_ticket_verbs.dm
```diff
@@ -2,13 +2,24 @@
...
+	if(!obscurevariablebeingusedforthis)
+		usr << "<span class='ticket-status'>Please select one of the options to receive help.</span>"
+		return
 
+	if(obscurevariablebeingusedforthis == "Wiki")
+		usr << "<span class='ticket-status'>Here is a link to the TGStation wiki: https://tgstation13.org/wiki/Main_Page</span>"
```
AshCorr, line 12: I think you made a typo here, wiki.yogstation.com is spelt W-I-K-I-.-Y-O-G-S-T-A-T-I-O-N-.-N-E-T

AshCorr (changes requested):

Time-Green:
Did you even test this? What the code is gonna do:
1. Type a message
2. Ask the wiki, bug report something
This is gonna make worse then it already is.
Also, if I were already talking with an admin, what would prevent me from getting THAT prompt everytime?

Kmc2000:
You da man

code/modules/admin/tickets/admin_ticket_verbs.dm
Jc1469: We agreed to not use the yogstation wiki, and to use the TG one instead.  In a few places ours looks like spaghetti was thrown at the screen.

Jc1469:
> Also, if I were already talking with an admin, what would prevent me from getting THAT prompt everytime?

Maybe the fact that every time you ahelp it might not be for the same thing.  ?

Time-Green:
If I am talking to an admin I really dont want to type out my whole message, and then have a stupid pop-up questioning what for.

At least make this not happen in an active ticket and make it give the pop-up before prompting to enter a text. Right now, it makes you type a message, and then it asks you what for.

You really didn't test this, did you?

Jc1469:
It doesn't happen if you use the ticket button. It happens if you use the adminhelp verb which isn't actually an adminhelp verb. 

I did test it, thank you very much. nitpick nitpick

Time-Green:
eh that's fine

code/modules/admin/tickets/admin_ticket_verbs.dm
Super3222: TGstations wiki covers content that Yogstation doesn't have, but at the same time I can see why John wants to link tgs instead of yogs. Our wiki mods need a direction because there's been a big stagnation in wiki development.

The wikia needs
- to promote me to a system administrator with the same powers as Oisin
- to promote all of the wiki mods on the discord channel to a special rank, and make two new channels. one public wiki channel, and another private.
- coders need to have a rank on the wiki, so after getting their PR merged they can jump on it to edit in their changes.
- say we can add an [Update Wiki] tag for wiki editors to monitor.
- the wiki application is outdated and needs to be revamped. for example, it still requires you to state your irl name... which is ridicolous
- a completely different theme, so we can kick off wiki development with a fresh looking wiki instead of what we have now
- possibly MOTIVATION. i don't know what started the drive to sign up to be a wiki mod in the first place, but the most successful wikias out there did it by putting a bargain for antag tokens w/ limits. which is exactly what we need to do for our own. possibly.

Super3222: somebody pitch that to groudon

code/modules/admin/tickets/admin_ticket_verbs.dm
```diff
@@ -2,13 +2,24 @@
 /client/verb/admin_ticket(ticket_title as text)
```
Time-Green, line 2: This gives the person a text pop-up

code/modules/admin/tickets/admin_ticket_verbs.dm
Time-Green: And then it asks for what it is.

reverse this

Jc1469:
il make a new verb entirely so that this doesn't even have to touch the ticket system later
busy for now though

Issue #2109, opened 2017-04-04
Wet floors dry too fast 2: Electric boogaloo


Oakboscage:
Wet floors still dry almost instantly after being mopped. Which makes being a janitor NO FUN ALLOWED and make the wet floor signs useless junk.

It's time to make janitor great again by setting the drying times to something more reasonable like they used to be. I believe this should be a rather simple fix and I'd do it myself if I had any idea where this shit is located and could trust myself to not just break everything forever.

Jc1469:
https://github.com/yogstation13/yogstation/blob/f589a9ca72e7273ba8d85d7d115da0ad4de6ce75/code/game/turfs/open.dm#L111 :ok_hand:

Oakboscage:
I looked at that but I couldn't figure where wet_time is actually calculated based on amount of water.

matskuman5:
Sure I'd support this, but I think that the last time this was discussed Cruix said that changing this would require reworking the entire new slip code, so idk

Oakboscage:
Why do we even have a new slip code if everything is broken now?

Cruix:
I never said that.

Cruix:
The reason they dry so fast with a mop is because it scales with how much reagent was splashed on the floor, and mops only apply 1 unit of water.

ShadowDeath6:
If you want to slip people just fill a cup with 50u water from a sink and throw it at the ground

Oakboscage:
Sure you could fill a bucket or cup to slip people but then it's clear it's intentional. Takes a lot away from being antag janitor where you could slip + robust targets and the slip itself wasn't clearly intentional so you had the benefit of the doubt if someone sees it.

I also think that the dynamic between a janitor doing his job and the crew who always gotta go fast was an important part of 2D spacemen.

Cruix:
It used to be when mopping was the only way to clean floors quickly at round start, but then the janitors had to go fast too and they got SPACE CLEANER SPRAYBOTTLES and SPACE CLEANER BACKPACK TANKS and SOAP and CLEANER GRENADES and they got them all at ROUND START because GOTTA GO FAST CANT WAIT FOR CHEMISTRY OR RESEARCH.

That dynamic is probably gone along with RnD needing acid from chemistry, chemistry needing corn oil from botany, botany needing exotic seed crates from cargo, cargo needing the bartender's shotgun to open crates, xeno needing anything from anyone, and anyone needing to communicate with anyone outside their department in general.

[...]

PR #2343, opened 2017-06-29
Makes Janitor fun again

Oakboscage:
Makes floors stay wet for a reasonable amount of time.

Refer to issue #2109 .

Fixes issue #2109 

#### Changelog

:cl:
rscadd: Buffed mop reagent amount.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/mop.dm
+++ b/code/game/objects/items/weapons/mop.dm
@@ -77,7 +77,7 @@
 				if(C.thermite)
 					C.thermite = 0
 					C.overlays.Cut()
-		reagents.reaction(A, TOUCH, 5)	//Needed for proper floor wetting
+		reagents.reaction(A, TOUCH, 200)	//Needed for proper floor wetting
 		reagents.remove_all(1)			//reaction() doesn't use up the reagents
 	else if(user.a_intent != "harm")
 		A.visible_message("[user] starts to wipe down [A] with [src]!", "<span class='notice'>You start to wipe down [A] with [src]...</span>")


LarissaMayrink:
200 fucking ticks, whats your damm problem

LarissaMayrink:
200, if we assume that it goes at 1u per tick thats 20s with 0% distortion, and, by being able to mop the floor multiple times thats enough to keep a tile wet for as long as it takes for nich to get woo wooed by void. Your janitor fantasies are worse than my fanfics

Oakboscage:
Actually wet_time maxes out at 300 no matter how much you mop it. Also wet_time depends on how much water the mop contains when you mop it so assuming regular janitor work (you don't wet the mop for each tile) only one in 5 tiles would have the full 200, the rest having fractions of that as it degrades. So you would have single tile wetspots that stay for a bit longer. 

This is actually pretty close to how janitor used to be before they ruined it. Wet floor signs will actually have a purpose again and the crew will have a reason to walk.

Jc1469:
balance police do not view this as a balance change, and instead view it as a "why the was it changed to begin with?" change

LarissaMayrink:
>you don't wet the mop for each tile
Unless you want to fill a corridor with wet floors, which you are not able to right now but your change will make it possible again. Its nor like it is hard to click once on the cart after each tile, and its not like you can't do it even while you are already moping.

Oakboscage:
Well if a janitor is just wetting floors for no reason and not putting up signs, security can scuttle over and arrest him for creating a workplace hazard, just like in the good old days.

I realize this is going to step on some toes due to getting in the way of some dank powergaming and validhunting.

matskuman5:
:+1:

LarissaMayrink:
How is giving the janitor back the possibility to wet entire hallways going to get in the way of validhunting?

ShadowDeath6:
as long as this doesn't make water from spraybottles make the floor wet

Oakboscage:
Unless I've really fucked something up this only affects mopping so liquids applied by any other methods are unchanged.

Slinky5890:
You've been campaigning for this for so long. I'm glad you've become the change you wish to see in the world.

Cruix:
This has horrendus implications for thermite, plasma, and every other reagent that reacts with turfs, and in return only improves the griefing ability of a legal grief tool for the janitor that also happens to be the worst tool he has for its actual intended purpose.

Oakboscage:
Well this sure had some unintended consequences, back to the drawing board.

PR #2350, opened 2017-07-01
Never option for ghost role pop-up's

Time-Green:
Highly requested by literally everyone including me.
Now that when someone spams for their fucking mining drone to become sentient, you can hit never and you will wont be prompted about that shit again. And you can still get your xenoheadslug! 

:cl:
rscadd: When a ghost role gets prompted, also gives another option: "Never", clicking this will never send you a notification for that kind of mob ever again. Unless you re-join the living in wich case it resets.
/:cl:

### DIFF
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -403,7 +403,7 @@
 		if(be_special_flag)
 			if(!(G.client.prefs) || !(be_special_flag in G.client.prefs.be_special))
 				continue
-		if(M && G.client && M.type in G.client.rejectedRoles)
+		if(M && G.client && (M.type in G.client.rejectedRoles))
 			continue
 		if (gametypeCheck)
 			if(!gametypeCheck.age_check(G.client))
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -391,7 +391,7 @@
 
 	return new /datum/projectile_data(src_x, src_y, time, distance, power_x, power_y, dest_x, dest_y)
 
-/proc/pollCandidates(var/Question, var/jobbanType, var/datum/game_mode/gametypeCheck, var/be_special_flag = 0, var/poll_time = 300)
+/proc/pollCandidates(var/Question, var/jobbanType, var/datum/game_mode/gametypeCheck, var/be_special_flag = 0, var/poll_time = 300, mob/M)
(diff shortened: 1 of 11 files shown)


Cruix:
Did you test this?

code/__HELPERS/game.dm
```diff
@@ -403,6 +403,8 @@
 		if(be_special_flag)
 			if(!(G.client.prefs) || !(be_special_flag in G.client.prefs.be_special))
 				continue
+		if(M.type in G.rejectedRoles)
+			continue
```
Cruix: This will runtime if M is null, which it is for the vast majority of pollCandidates() calls

code/modules/client/preferences_savefile.dm
```diff
@@ -50,8 +50,9 @@ SAVEFILE UPDATING/VERSIONING - 'Simplified', or rather, more coder-friendly ~Car
 		var/B_shadowling = 8192
 		var/B_abductor = 16384
 		var/B_revenant = 32768
+		var/B_special = 65536
```
Cruix: This number is too large for a BYOND bitflag

code/modules/mob/dead/observer/observer.dm
```diff
@@ -57,6 +57,8 @@ var/list/image/ghost_images_simple = list() //this is a list of all ghost images
 	var/body_attack_log
 	var/body_say_log
 
+	var/list/rejectedRoles = list()
```
Cruix: This should be in the Client, or a subsystem, not in a ghost mob.

Cruix (changes requested):

code/modules/client/preferences_savefile.dm
Time-Green: What should I do then? Revert it to use the xenomorph preference?
It worked fine on local, but I assume it creates performance issues.

Cruix: It worked on local because you never had to load it from the preferences file. Every time someone connects it will be set to zero.

You can add a new bitflag variable like we did for PDA bitflags or you can refactor the current system to use a list instead of a bitflag like we did for status_flags.

Time-Green:
The variable has been moved and the runtime patched.
About the byte storage, I reverted it to alien. It's complicated and something for another PR

Cruix (approved):

PR #2357, opened 2017-07-05
Fixes photocopiers

matskuman5:
Photocopiers on the station should work properly now, they used to be stuck unpowered because of their power_channel being EQUIP instead of 1

:cl:
bugfix: Photocopiers should now work properly.
/:cl:


Cruix:
EQUIP is a define for 1.

Cruix:
Actually no, super overwrote it with his own define for mining vendors.

PR #2359, opened 2017-07-05
Remodels the law office

matskuman5:
Makes it a bit bigger, it now looks like a proper OFFICE instead of some shitty closet

http://i.imgur.com/66ilKA9.png

Those bookcases are filled with Space Law btw
:cl:
tweak: NanoTrasen designers have remodeled the Law Office.
/:cl:


Cruix:
Lawyers have read-only access to security records through their PDA, they do not have access to security records consoles because the ability to set people to arrest is purposely limited to roles with mindshield implants (ignoring the fact that the HoP can give himself the access).

matskuman5:
Hm, how about we make lawyers mindshielded?

Cruix:
Should be a separate PR.

The console is fine, the lawyer can ask an officer to log in for him and if they trust him. I would love to see security and medical records used more than the current "antags delete their own record so they cannot be set to arrest" meta that exists now.

Partheo:
Very nice. It did look like a closet before. Are those wooden walls?

matskuman5:
Yup @partheo

Ratvar:
Devil metaclub revealed, now you can fireball crew contracts at godspeed outta wood, stealthy buff alert

Zostroll:
The rack is a bit to industrial for the theme of the room and the lawyer has nowhere for their clients to sit. Other than that I think it looks neat!

Zostroll:
I just saw the seat behind the red cross...

Super3222:
I see where Zostroll is going with the racks. You can change it if you want. Other than that, this looks good to go.

matskuman5:
I'll fix this later

PR #2383, opened 2017-07-14
Adds admin logging to signalers

matskuman5:
Whenever a signaler is activated, it is logged in the game logs.

See https://forums.yogstation.net/index.php?threads/logicalhazard-adameltablawy-bombing.14334/ as to why this is necessary.

### DIFF
--- a/code/modules/assembly/signaler.dm
+++ b/code/modules/assembly/signaler.dm
@@ -114,6 +114,7 @@ Code:
 	var/turf/T = get_turf(src)
 	if(usr)
 		lastsignalers.Add("[time] <B>:</B> [usr.key] used [src] @ location ([T.x],[T.y],[T.z]) <B>:</B> [format_frequency(frequency)]/[code]")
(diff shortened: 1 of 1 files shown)


Partheo:
I don't know why this isn't a thing yet. Getting the "last fingerprint" instead of the actual user of the signaller seems archaic.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Last fingerprint is actually the last user complete with ckey and name

Partheo:
Right, but I mean it's more useful to get the ckey of the person who pressed the send signal button, rather then the ckey of the person who's fingerprint is on the signaller.

matskuman5:
The ckey Partheo is talking about is actually of the person who held the BOMB last.

Who touched the signaler was never logged, which this PR fixes

Super3222:
We could add these to an investigation list if you want.

matskuman5:
@Super3222 that would be useful

matskuman5:
@Super3222 wait no, it already exists. It's just in secrets, not investigate.

Super3222:
Ah, I see then.

PR #2386, opened 2017-07-15
Fixes random golem

Time-Green:
I'm still bashing my head on the wall trying to comprehend what the fuck the guy that made that was thinking.
I can't really switch branches now, so this wasn't tested. But this couldn't possibly go wrong.
:cl:
fix: The golem mutation toxin now properly works, instead of being a piece of shit fucking code
/:cl:

### DIFF
--- a/code/modules/golem/golem_types.dm
+++ b/code/modules/golem/golem_types.dm
@@ -33,13 +33,12 @@
 	blacklisted = FALSE
 	dangerous_existence = FALSE
 
-/datum/species/golem/random/New()
+/datum/species/golem/random/on_species_gain(mob/living/carbon/C, datum/species/old_species)
 	. = ..()
-	var/list/golem_types = typesof(/datum/species/golem) - src.type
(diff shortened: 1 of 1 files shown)


code/modules/golem/golem_types.dm
```diff
@@ -33,13 +33,12 @@
 	blacklisted = FALSE
 	dangerous_existence = FALSE
 
-/datum/species/golem/random/New()
+/datum/species/golem/random/on_species_gain(mob/living/carbon/C, datum/species/old_species)
```
Super3222, line 36: what is `datum/species/old_species` used for here?

Time-Green: No idea, I always include the same parameters used in the parent proc. If anyone ever copies it like I did, at least they'll have the whole thing

Cruix: The only thing I know that actually uses it is synths, which use it to know what species to emulate.

PR #2389, opened 2017-07-16
Fixes lava rivers going through abductor ships

### LINKED ISSUE #2388: Lava Rivers in Ruins.. again?
So, this happened on the server again, round counter says round 10519 (if it matters)

screenshot_1

screenshot_117

Not sure why it happened, but.. well, it did. Assuming it isnt actually part of the ruin to have lava like that.

ghost:
fixes #2388

couldnt test this because its an edge-case and i cant test it but I'm about 99.9% sure this fixes it. In the future use the proper area types when you make ruins, theyre there for a reason.

more specifically i switched the lavaland/surface/outdoors area with ruin/unpowered areas. lava rivers only go onto lavaland/surface/outdoors areas which is why the lava river could be created in the ruin.

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_abductorterrorship.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_abductorterrorship.dmm
@@ -24,47 +24,47 @@
 "ag" = (
 /obj/machinery/computer/camera_advanced/abductor,
 /turf/open/floor/mineral/abductor,
-/area/lavaland/surface/outdoors)
+/area/ruin/unpowered)
 "ah" = (
 /obj/structure/table/abductor,
 /turf/open/floor/mineral/abductor,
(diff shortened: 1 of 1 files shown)


ghost:
hey i actually map merged properly. what a surprise.

Super3222:
The intention of placing those outdoor areas on the map was to allow ash storms to get inside of the ruin. This added a level of danger and uncertainty to it instead of making it a safe haven to miners. Lava rivers getting inside of it was unintentional. I'll have to create a new area (unless you want to) that counts as outdoors and is vulnerable to ashstorms, but cannot be touched by the lavaland river generator.

Super3222:
This is being re-done in https://github.com/yogstation13/yogstation/pull/2409, to fix the bug and allow the original feature to exist (ash storm exposed ruin).

PR #2409, opened 2017-07-21
Abductor Terrorship is now mapgen protected

Super3222:
Fixes https://github.com/yogstation13/yogstation/issues/2388

#### Changelog

:cl:
fix: Lava can no longer spawn in the abductor terror ship.
fix: Fake ash storms no longer look like real ash storms.
/:cl:

### DIFF
--- a/_maps/RandomRuins/LavaRuins/lavaland_surface_abductorterrorship.dmm
+++ b/_maps/RandomRuins/LavaRuins/lavaland_surface_abductorterrorship.dmm
@@ -24,15 +24,21 @@
 "ag" = (
 /obj/machinery/computer/camera_advanced/abductor,
 /turf/open/floor/mineral/abductor,
-/area/lavaland/surface/outdoors)
+/area/lavaland/surface/outdoors/mapgen_protected{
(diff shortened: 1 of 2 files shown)


_maps/RandomRuins/LavaRuins/lavaland_surface_abductorterrorship.dmm
```diff
@@ -24,47 +24,65 @@
 "ag" = (
 /obj/machinery/computer/camera_advanced/abductor,
 /turf/open/floor/mineral/abductor,
-/area/lavaland/surface/outdoors)
+/area/lavaland/surface/outdoors/mapgen_protected{
+	icon_state = "unknown"
```
Cruix, line 28: Why are all the area tiles varedited?

Super3222: I changed the icon to tell the difference from the new area that I was putting down vs the old one. I couldn't use tools like notepad++ to replace ever instance of `/area/lavaland/surface/outdoors/` because it was needed outside of the ship.

PR #2414, opened 2017-07-23
removes eating organs

ShadowDeath6:
I saw an issue on hippie about this and I felt prompted to do it here because why not

this is made with the online editor and I'm going to be gone for five days so I can't compile until then anyway

### DIFF
--- a/code/modules/surgery/organs/organ_internal.dm
+++ b/code/modules/surgery/organs/organ_internal.dm
@@ -120,17 +120,7 @@
 	return ..()
 
 /obj/item/organ/attack(mob/living/carbon/M, mob/user)
-	if(M == user && ishuman(user))
-		var/mob/living/carbon/human/H = user
-		if(status == ORGAN_ORGANIC)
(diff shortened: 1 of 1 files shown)


LarissaMayrink:
Huh, but why?

ShadowDeath6:
if you take one bite of an organ accidentally it renders it completely useless

also encourages deepfry / chef recipes for what it's worth

LarissaMayrink:
But that makes sense, if you bite a organ it won't work. Also Deep frying doesn't make a item edible if it wasn't edible before being deep fried, not on yog, so, you can't just eat a deep fried id, the id will drop at your feet and thats all. The same goes for a non edible organ, it would just drop at your feet like everything else.

Also, you mean this issue?
https://github.com/HippieStation/HippieStation/issues/2449

ghost:
then fix the issue somehow else, don't remove edibleness. Lower the bitesize for example

Kmc2000:
Delet this

PR #2419, opened 2017-07-23
[s] adds admin fingerprints to pipes, crayons, and floors

Jc1469:
whine about how I'm not using fingerprints so I can proceed to still not use fingerprints

### DIFF
--- a/code/game/machinery/pipe/construction.dm
+++ b/code/game/machinery/pipe/construction.dm
@@ -231,6 +231,7 @@ var/global/list/pipeID2State = list(
 	if(istype(T))
 		T.flipped = flipped
 	A.construction(pipe_type, color)
+	A.placedby = user.ckey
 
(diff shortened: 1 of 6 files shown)


Kmc2000:
Godchurch

LarissaMayrink:
So, as far as i understood this; it notes the ckey of who made a fake gangtag and some other stuff, so, could you do the same for the letters? it would make possible to check who wrote something on the floor when bigotry happens

Jc1469:
This var covers all crayon cleanables.  Spraycans are classed as crayon cleanables.
It also covers pipe placements and floor tile placements.  Namely in the case of when you wrench a pipe down it spawns a new object entirely and erases the fingerprints from the pipe. 
And in the case of floor tiles where, say, someone fills a hallway with plasma tiles or uranium tiles and suddenly we have claude frollo showing up with no clue whodunnit.

PR #2434, opened 2017-07-29
[READY] KA rework

Time-Green:
Reworks KA's into parts and removes the super and hyper KA.
You can get these parts from vendor, cargo, science and from tendrils.
You can freely swap out parts of the KA to fit your fighting style or just to straight up fucking improve the shit.
:cl:
rscadd: Reworks KA into parts. Super parts can be found in vendors, special parts like 3x3 mining and 2 shots can be printed aswell. Maybe some ancient parts exist in lavaland...
/:cl:

- [x] Finish adding more parts
- [x] Grant people the parts in a practicle way
- [x] Fix the mess left behind from changing them
- [x] Add super parts to mining vendor and fix some misc shit


MoreRobustThanYou:
Freeze, sadly.

code/modules/projectiles/guns/energy/kinetic.dm
```diff
@@ -439,7 +439,8 @@
 	update_icon()
 
 /obj/item/weapon/gun/energy/kinetic_accelerator/proc/attempt_reload()
-	if(overheat)
+	world << "[power_supply.charge]"
```
X-TheDark: Debug message

Time-Green: fek

ShadowDeath6:
what does each upgrade do

Time-Green:
ughhhhhh.
The super parts all build up to a normal super KA, with each part having main attributes. Like a charger holding the most cooldown.


Science ones:
Splash: does the old hyper 3x3 effect
Shotgun: Shittily low range but does alot of damage
Double: Gives the person 2 KA shots, but takes significantly longer to load
wide grip: allows golems and shit to use it

loot ones:
vampiric: heals the user for every hit. like 1 health
triple: gives the person 3 shots, but takes a long time to load

I dont really feel like adding more ATM, probably something for later

ShadowDeath6:
also- any reason why you made crossbows not a subtype of KAs?

Time-Green:
All the stats come from the parts, and an energy crossbow isnt done in parts. 

Also I still need to test that...

ShadowDeath6:
I'd just leave it as is, to avoid copypasta code for the recharging

Time-Green:
I really dont feel like puting checks everywhere for energy crossbows, making a new subtype is alot less effort

ShadowDeath6:
>putting checks everywhere

Wouldn't it just be one check with `var/compatibility` or something when you try and install upgrades on it?

Time-Green:
ALL STATS ARE CALCULATED BY THE PARTS IT HAS. I am not gonna snowflake it when it's 10x more convenient to just give it it's own proc

code/modules/projectiles/guns/energy/kinetic.dm
```diff
@@ -0,0 +1,488 @@
...
+
+/obj/item/kinetic_part/charger/super
+	name = "super charger"
+	icon_state = "charger_super"
+	rank = "super"
+	part = "charger"
+	cooldownBonus = 8
+
+/obj/item/kinetic_part/grip/super
+	name = "standard grip"
```
X-TheDark: super grip

code/modules/projectiles/guns/energy/kinetic.dm
```diff
@@ -0,0 +1,488 @@
...
+	return totalCooldown
+
+/obj/item/kinetic_part
+	name = "kinetic part"
+	icon_state = ""
+	icon = 'icons/obj/guns/kinetic.dmi'
+	var/rank = "standard"
+	var/part              //"barrel","barrel_end","core","charger","grip"
+	var/pixel_x_extra = 0  //If you make a part thats longer then the original, please write down how much pixels longer. Mainly for use in barrel and
+	var/pixel_y_extra = 0           //barell end
```
ShadowDeath6: barell

ShadowDeath6:
you're going to have to snowflake something either way considering this way you'll still need to copy and paste the entirety of the automatic reload into somewhere else.
imo it's more convenient to just make the crossbow a subtype, but you do you

Time-Green:
I had to add a wopping 4 lines to make it work, and it works

Super3222:
Do you wanna maybe elaborate about the parts in the changelog? I get the secrecy part to let the players hunt for the features themselves, but from personal experience of doing the same thing it comes off as an underwhelming change.

Time-Green:
Explained some more in the changelog

ghost:
just port /tg/'s version

code/modules/mining/lavaland/necropolis_chests.dm
```diff
@@ -66,6 +66,9 @@
 			new /obj/item/weapon/spellbook/oneuse/summonitem(src)
 		if(23)
 			new /obj/item/organ/heart/cursed/wizard(src)
+		else
+			new /obj/item/kinetic_part/core/vampiric(src)
+			new /obj/item/kinetic_part/charger/triple_shot(src)
```
ShadowDeath6: Why is it done like this? The chest will hold both of the upgrades (if that's intended), and you're using else over (24), which makes it look snowflakey.

Time-Green: I make the random thingy bigger, I had a brainfart aswell

ShadowDeath6: making it (25) results with this item in specific being more likely than other items to be rolled

Time-Green: Wich is why I did that. I wanted both of them to be more common. I'll have it pick from an item pool inside so the parts dont overwhelm it.

ShadowDeath6:
I still don't like the way you implemented them in the loot list. Just make one (24) and the other (25) if you're going to make another switch inside the else.

Time-Green:
Sooner or later this will overwhelm the loot drops, if people actually add more to it...
Right, no one will

Cruix:
Radiation crossbows no longer update their icon when they recharge.
If you swap to a new charger, your next shot will still have the same number of shots as the old charger for the next shot. It should probably decharge the KA whenever you swap out the charger.

Time-Green:
Fixed

drderp3635:
whenever you put a KA thats charging into your bag it breaks the sprite, and it will always show as uncharged even though you can fire it. 

needs more parts in my opinion

Time-Green:
1. I know that and I'm fixing it
2. I know that too, but I kinda hoped others would

Time-Green:
I fixed the icon breaking

drderp3635:
my ideas that i think you people should code for this:
Bayonet mount- Barrel end that works like a standard barrel but can have a knife attached to it, allow kitchen knives, combat knives, the starting knife, and energy pen
Heavy mount- would take both grip and barrel (if you can make that happen) to mount a melee weapon on the underside of the gun, allow medical saw, katana, and esword.
long barrel- increases the range by several tiles but also increases the cooldown time
momentum charger- charge speed based on your movement. doesn't charge if you are standing still, but charges rapildy if you are constantly running.

meme idea: kenetic color converter barrel- barrel created by injecting blood into a pyrite slime, that converts the accelerators damage into stamina, and paints anything it hits.

drderp3635:
we need more fancy cores as well. 
add a fancy upgraded one that requires earthsblood as a chemical to force interdepartmental interaction and boost the vampirics stats to it

PR #2446, opened 2017-08-01
Adds 1u and 1/3u options to chem dispenser

monster860:
:cl: monster860
tweak: You can now fill up a chem beaker with 1/3u and 1u of a chemical
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -1,7 +1,7 @@
 /obj/item/weapon/reagent_containers/glass
 	name = "glass"
 	amount_per_transfer_from_this = 10
-	possible_transfer_amounts = list(5, 10, 15, 20, 25, 30, 50)
+	possible_transfer_amounts = list(1/3, 1, 5, 10, 15, 20, 25, 30, 50)
 	volume = 50
 	flags = OPENCONTAINER
 	spillable = 1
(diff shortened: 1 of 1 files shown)


Super3222:
Interesting.

Cruix:
Beakers are now more precise than droppers

monster860:
Actually the most precise thing is a chemmaster

Cruix:
The chemmaster uses the same transfer options as the beaker that is in it

monster860:
You also have a "custom" button in there so you can transfer any amount

Cruix:
Then why does it need 1 and 1/3 options?

monster860:
You have those options in the chem master not the dispenser

ShadowDeath6:
1u I'm fine with, one third seems kind of weird. Why not one half? Why not any fraction?

AshCorr:
> 1u I'm fine with, one third seems kind of weird. Why not one half? Why not any fraction?

Yea... Could we just use rational numbers? which saves us having ungodly chem quantities which have a ton of decimal places.

Mrcreepssss:
yeah, keep it to one and not three, if you really want it that badly use three 1u transfers

matskuman5:
^ seconded

ghost:
use 3 1u transfers to get 1/3

yes that makes sense @Mrcreepssss

Mrcreepssss:
ehhhhm @AdamElTablawy  i get the impression you misunderstood me as your comment makes no sense to me (it seems sarcastic), what i said that to get 3 unit transfer you can just use three 1 unit transfers.

PR #2447, opened 2017-08-01
Make the empty implanter actually start out empty

monster860:
:cl: monster860
fix: The empty implanter will actually start out empty now
/:cl:


Mrcreepssss:
yay

PR #2458, opened 2017-08-05
A lot of var logging for airlocks and airlock related things.

Jc1469:
i hate airlocks

### DIFF
--- a/code/datums/wires/airlock.dm
+++ b/code/datums/wires/airlock.dm
@@ -16,7 +16,7 @@
 	..()
 
 /datum/wires/airlock/interactable(mob/user)
-	var/obj/machinery/door/airlock/A = holder
+	var/obj/machinery/door/airlock/A = holder	
 	if(!istype(user, /mob/living/silicon) && A.isElectrified() && A.shock(user, 100))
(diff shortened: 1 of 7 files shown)


ghost:
good pr. might port that to FTL actually

code/datums/wires/airlock.dm
```diff
@@ -16,7 +16,7 @@
 	..()
 
 /datum/wires/airlock/interactable(mob/user)
-	var/obj/machinery/door/airlock/A = holder
+	var/obj/machinery/door/airlock/A = holder
```
AshCorr: :(

AshCorr (changes requested):

code/datums/wires/airlock.dm
Jc1469: .

AshCorr (approved):

Mrcreepssss:
wait uhh, why is AI control disabled done twice?

edit: actually i saw it four times...

Jc1469:
Because there's two seperate functions to disable it per instance.  And 2 instances.

PR #2461, opened 2017-08-06
Dead Ringer

monster860:
:cl: monster860
rscadd: Adds the dead ringer, a traitor item costing 8 TC. Holding this item while being stunned or slipped cancels the effect, and leaves behind a fake body, and the user becomes invisible for 5 seconds. For the first 2 seconds, the user is invulnerable to stuns. After the 5 seconds, the user becomes visible enough and a loud decloak noise can be heard. The device cannot be used for 1 minute, and during this minute waiting period the device blinks blue. Once the waiting period is over the device goes back to being green and can be used again
/:cl:
https://puu.sh/x2KRD/2746692539.mp4


ghost:
can the deadringer's user still attack while cloaked? if so, change it so that attacking cancels the effects.

monster860:
No, the user cannot attack during the 5 seconds of cloak.

ghost:
what if they decide attacking is a good choice tho

code/game/objects/items/devices/deadringer.dm
```diff
@@ -0,0 +1,77 @@
+/obj/item/device/deadringer
+	name = "strange device"
+	icon_state = "motion0"
+	var/ready = TRUE
+	var/mob/living/carbon/human/dummy
+
+/obj/item/device/deadringer/proc/trigger(mob/living/carbon/human/user)
```
Cruix, line 7: But what if user is not a human?

ShadowDeath6:
Suggestions-

If you can't already, have it be usable in hand to deactivate early, in case they accidentally use it or want to uncloak for some reason.

Make it non-examinable while held. I think the ABSTRACT flag does that, but I'm not sure.

Have it display a text log when ready to activate again, just in case they're not paying attention to their inventory waiting for it to reactivate.

code/game/objects/items/devices/deadringer.dm
monster860: Then it doesn't work. There's an istype check

ghost:
make the dummy items and mob dissapear after 10 seconds 👌

code/game/objects/items/devices/deadringer.dm
```diff
@@ -0,0 +1,77 @@
...
+	var/old_invis = user.invisibility
+	user.invisibility = INVISIBILITY_OBSERVER
+	var/revoke_sight = !(user.sight & SEE_SELF)
+	user.sight |= SEE_SELF
+	user.alpha = 50
+	ready = FALSE
+	flags |= NODROP
+	user.next_move = world.time + 50 // Prevent the user from clicking on anything for the next 5 seconds.
+	icon_state = "motion1"
+	spawn(20)
```
Cruix: These spawns should be addtimers, and all the procs should also be called in Destroy() in case the deadringer somehow gets deleted before the invisibility wears off

code/modules/mob/status_procs.dm
```diff
@@ -7,6 +7,10 @@
 /////////////////////////////////// STUNNED ////////////////////////////////////
 
 /mob/proc/Stun(amount, updating = 1, ignore_canstun = 0)
+	var/mob/dummy = check_deadringer()
+	if(dummy)
+		dummy.Stun(amount, updating, ignore_canstun)
+		return
```
Cruix, line 13: If it only works for humans, all these checks should be in /mob/living/carbon/human/Stun(), Weaken(), etc.

code/game/objects/items/devices/deadringer.dm
```diff
@@ -0,0 +1,77 @@
...
+		dummytype = /obj/item/clothing/under
+	var/obj/item/dummy_item = new dummytype
+	dummy_item.appearance = I.appearance
+	dummy_item.item_state = I.item_state
+	dummy_item.lefthand_file = I.lefthand_file
+	dummy_item.righthand_file = I.righthand_file
+	dummy_item.w_class = I.w_class
+	dummy_item.flags_cover = I.flags_cover
+	dummy_item.flags_inv = I.flags_inv
+	dummy_item.alternate_worn_icon = I.alternate_worn_icon
```
Cruix, line 99: The dummy objects should probably be nodrop or dropdel so you cannot farm fake items.

Jc1469:
8 cost
5 second invisibility
2 second stun immunity
60 second cooldown
triggered by slips

The stun immunity is a bit eh.   And if it only triggers on stuns, it seems like a meme item designed to sell yourself out.
I hate memes.

I like the evolution from 10 second invisibility on a 60 second cooldown to this, but this still doesn't seem viable in nearly any situation.  Much like a briefcase of money.

ghost:
it doesn't happen on just stuns u tit.

code/modules/mob/status_procs.dm
monster860: Doesn;t look like there is a /mob/living/carbon/human/Stun()

code/game/objects/items/devices/deadringer.dm
monster860: done

code/modules/mob/status_procs.dm
Cruix: There does not have to be, you can make one and only call ..() if the dummy check fails

monster860:
@Jc1469 I can make it trigger whenever external damage happens, like being attacked.

Jc1469:
I think it should trigger on damage aswell.  Just, on a hit of any sort.  A stun, a slip, anything.  The only issue I see with slipping is you completely sell yourself out if you accidentally have it out or something like that and get slipped.  But as it currently is, only working on stuns, it seems very impractical for the cost.

Cruix:
It would also encourage powergaming security officers to exclusively use disablers.

Notamaniac:
Can I run over ammo boxes to replenish 40% of my cloak?

Cruix:
No, but if you stand still in a corner it will recharge slowly. Also headshots with the revolver now crit.

ShadowDeath6:
when will headshotting actually do double damage?

Kmc2000:
This is fucking cool

Ratvar:
add toy version that can eat foam ammo boxes for charge and only plays sound

code/game/objects/items/devices/deadringer.dm
```diff
@@ -0,0 +1,83 @@
...
+	icon_state = "motion1"
+	addtimer(src, "lose_invuln", 20)
+	addtimer(src, "lose_invis", 50, FALSE, user, old_alpha, old_invis, revoke_sight)
+	return dummy
+
+/obj/item/device/deadringer/proc/lose_invuln()
+	dummy = null
+/obj/item/device/deadringer/proc/lose_invis(mob/user, old_alpha, old_invis, revoke_sight)
+	user.invisibility = old_invis
+	animate(user, alpha = old_alpha, time = 20)
```
AshCorr: If the user was invisible or had a low alpha when they use the dead ringer they'll revert to that alpha after they finish using the ringer. 

This could be abusable
Gets Chameleon skin from genetics> Waits to become invisible > Uses Dead Ringer with some sort of chem that weakens the player > Uses a clean SE Syringe > Stops using Dead Ringer > Free Permanent invisibility?

Its highly unlikely, and you would probably need a partner but doable

Probably better to just set the users alpha and invisibility to their default values with initial(), I'd rather have visible people who are supposed to be invisible then have invisible people who are supposed to have visibility

ShadowDeath6:
https://github.com/yogstation13/yogstation/blob/master/strings/brain_damage_lines.json#L33

Cruix:
I am getting undefined variable runtimes when I use this while wearing glasses, and on the shoes when the dummies move:
undefined variable /obj/item/var/tint
proc name: equip to slot (/mob/living/carbon/human/equip_to_slot)

undefined variable /obj/item/var/bloody_shoes
proc name: Move (/mob/living/carbon/human/Move)

Also the dummies should probably get qdeleted at some point, the cargo bay full of fake quartermasters I created while testing it was very eerie.

code/game/objects/items/devices/deadringer.dm
```diff
@@ -0,0 +1,83 @@
...
+	ready = TRUE
+	icon_state = "motion0"
+
+/obj/item/device/deadringer/proc/create_dummy_item(obj/item/I)
+	if(!I || I == src)
+		return
+	var dummytype = /obj/item
+	if(istype(I, /obj/item/clothing/under))
+		dummytype = /obj/item/clothing/under
+	var/obj/item/dummy_item = new dummytype
```
X-TheDark: The errors Cruix pointed out might be due to this.
Try changing this line to `var/obj/item/dummy_item = new I.type`

This will make the item the same type it was originally on the unit, I think.

Cruix: That would be dangerous and allow for exploits if someone were wearing something that had side-effects they could use. He needs to either do what he did above for clothing, or add a lot of istype() checks to all the code that assumes worn items are of a particular type. This same sort of stuff happened when I made PDAs that could go on your wrist, because some code would try to check if the PDA had certain vars that only gloves had.

monster860:
There I did the changes @Cruix

.atom-build.json
```diff
@@ -0,0 +1,9 @@
+{
+	"cmd": "C:/Program Files (x86)/BYOND/bin/dm.exe",
+	"args": [ "D:/dev/repos/yogstation/yogstation.dme" ],
+	"sh": false,
+	"errorMatch":
+	[
+		"\n(?<file>[\\\\\/0-9a-zA-Z\\._]+):(?<line>\\d+).*"
+	]
+}
```
Cruix: What?

code/game/objects/items/devices/deadringer.dm
Cruix: Still an issue

code/game/objects/items/devices/deadringer.dm
Cruix: Still an issue

Cruix:
Also beepsky still chases you if you turn invisible.

monster860:
Beepsky is supposed to be the lord of time and space though.

ghost:
merge when

Time-Green:
Deadringer did the icon error bug in science jumpsuit for some reason
(E) (C) (M) icon = /icon (icons/obj/clothing/uniforms.dmi) 
(E) (C) (M) icon_state = "toxins" 
same as normal jumpsuit

Also might wanna give it a sprite and make the fake bodies dissapear, since this is fucking weird

Time-Green:
Please reopen when you fix it

PR #2454, opened 2017-08-02
Floors no longer dry too fast

Oakboscage:
Refer to issue #2109 .

### Intent of your Pull Request

Fixes issue #2109 by not having wet floors dry instantly. Should not have any gamebreaking unintended consequences like last time. 

#### Changelog

:cl:
rscadd: Floors take longer to dry
/:cl:

PR #2477, opened 2017-08-11
Increased water turf-wetting time

### LINKED ISSUE #2109: Wet floors dry too fast 2: Electric boogaloo

Cruix:
Fixes #2109 
Closes #2454 

Water turf-wetting time is now increased by a factor of ten, but it is capped at 50 (approximately 90 seconds). This is how long a full mop will wet a floor for.

:cl: Oakboscage  
rscadd: Wet floors take longer to dry
/:cl:

### DIFF
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -108,8 +108,10 @@
 					C.spin(1,1)
 		return 1
 
-/turf/open/proc/MakeSlippery(wet_setting = TURF_WET_WATER, min_wet_time = 0, wet_time_to_add = 0) // 1 = Water, 2 = Lube, 3 = Ice, 4 = Permafrost, 5 = Slide
-	wet_time = max(wet_time+wet_time_to_add, min_wet_time)
+/turf/open/proc/MakeSlippery(wet_setting = TURF_WET_WATER, min_wet_time = 0, wet_time_to_add = 0, max_wet_time = INFINITY) // 1 = Water, 2 = Lube, 3 = Ice, 4 = Permafrost, 5 = Slide
+	if(wet_time >= max_wet_time)
+		return
+	wet_time = min(max_wet_time, max(wet_time+wet_time_to_add, min_wet_time))
 	if(wet >= wet_setting)
 		return
 	wet = wet_setting
--- a/code/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -114,7 +114,7 @@
 	var/CT = cooling_temperature
 
 	if(reac_volume >= 5)
-		T.MakeSlippery(min_wet_time = 5, wet_time_to_add = reac_volume*0.2)
+		T.MakeSlippery(min_wet_time = 5, wet_time_to_add = reac_volume * 2, max_wet_time = 50)
 
 	for(var/mob/living/simple_animal/slime/M in T)
 		M.apply_water()


code/game/turfs/open.dm
```diff
@@ -108,8 +108,10 @@
 					C.spin(1,1)
 		return 1
 
-/turf/open/proc/MakeSlippery(wet_setting = TURF_WET_WATER, min_wet_time = 0, wet_time_to_add = 0) // 1 = Water, 2 = Lube, 3 = Ice, 4 = Permafrost, 5 = Slide
-	wet_time = max(wet_time+wet_time_to_add, min_wet_time)
+/turf/open/proc/MakeSlippery(wet_setting = TURF_WET_WATER, min_wet_time = 0, wet_time_to_add = 0, max_wet_time = INFINITY) // 1 = Water, 2 = Lube, 3 = Ice, 4 = Permafrost, 5 = Slide
+	if(wet_time >= max_wet_time)
```
ghost, line 112: cruix there's already a define for this, Oak's PR was fine. change the define instead

ghost (changes requested):
use the define instead.

code/game/turfs/open.dm
Cruix: The define is for all things that make a turf "wet", including lube and ice. This only affects water specifically.

ghost: is there really a need for that..? seems extremely snowflakey. also if so change the defines and variables to max_slip_time etc and turn max_wet_time into a define so this remains readable

Cruix: "Snowflake" refers to code that does a specific check in a way that is not reusable. Adding a new parameter in this way is generic enough for other pieces of code to use.

The define MAXIMUM_WET_TIME and the max_wet_time parameter serve two different purposes. The define is a blanket check that affects all MakeSlippery() operations, while this parameter can be used for individual operations, such as water's reaction_turf(), to reduce the significance of their effect without inhibiting the maximum signifigance of other operations, such as lube.

ghost: whats the purpose of limiting it in the first place though

ghost: if someone goes through the effort to make a tile that wet just crowbar it

Cruix: To reduce the power of a ubiquitously available method of stunning and murdering people, and griefing.

ghost: that's the point

Cruix: Or you could just walk.

ghost: see!

Oakboscage:
Capping it at 50 is terrible. Dumping a bucket of water out on the floor is now just the same as mopping it, and the wet time for mopping is still rather lame. Just use my PR instead.

Cruix:
Having a bucket add 6 minutes of wetness to a tile is not acceptable.

Oakboscage:
A) The odd puddle from someone running around with a bucket slipping is very acceptable
B) All you need to do to remove it is crowbar the floor tile

Cruix:
I am not going to force everyone to carry around a crowbar because the janitor is being a dick.

ghost:
1/2th the max slippery time then

Oakboscage:
Crowbars are literally everywhere. And there's this new radical concept of WALKING that you can use to avoid wet floors entirely. And if the janitor is being a dick he'll get arrested or robusted. It really isn't that terrible having a wet floor.

ghost:
i don't feel like there should be a cap-if you're enough of a nutjob to dump 1000u of water on one tile it deserves to be wet forever-but cruix has a point, oak. i can easily forsee tiles being wet for-fucking-ever in escape and hysterical as the image of a bunch of people constantly slipping and sliding for ten minutes as they desperately try to get onto the shuttle only to slip on someone else is, botany already does that and it's fucking cancerous.

X-TheDark:
Unfortunately, the "proper" solution would be for it to work like a chem grenade if you dump a bucket of water and have it splash across multiple tiles.

Unfortunately No.2, this isn't exactly a simple solution, plus I fear what would happen if this would be applied to things like High-Capacity Water Tank making the entire station slippery or somesuch.

As it stands, slippery time of 50 seems reasonable enough, in my opinion. Under normal circumstances, that's 25 seconds of slipperiness. While I, personally, wouldn't mind upping it to 100 (for 50 seconds of slip), water is too easy to come by to allow 50 seconds of slipperiness. Slowing people down with a beaker of water (either slip or forcing them to walk) is a big advantage.

Oakboscage:
There already is a cap, it's at 300.

Cruix:
300 is around 8 minutes on a server that is not lagging. 50 is just less than a minute and a half.

Oakboscage:
Can we agree on 150?

PR #2478, opened 2017-08-11
Abductor Terrorship is now mapgen protected

### LINKED ISSUE #2388: Lava Rivers in Ruins.. again?

Cruix:
Fixes #2388
Closes #2409 

I don't think we should be editing area vars in map files, because of the special way DM treats area instances. I replaced the varedited tiles with a proper icon, everything else is the same as #2409.

:cl: Super3222
fix: Lava can no longer spawn in the abductor terror ship.
fix: Fake ash storms no longer look like real ash storms.
/:cl:


ghost:
he mapgen_attac but he also mapgen_protec

PR #2488, opened 2017-08-16
PORT | Datum Component System

ghost:
#Ports the Datum Component System over from TG.
https://github.com/tgstation/tgstation/pull/29324

porting it because I need it for a big PR.

# Datum Component System (DCS)

## Concept

Loosely adapted from /vg/. This is an entity component system for adding behaviours to datums when inheritance doesn't quite cut it. By using signals and events instead of direct inheritance, you can inject behaviours without hacky overloads. It requires a different method of thinking, but is not hard to use correctly. If a behaviour can have application across more than one thing. Make it generic, make it a component. Atom/mob/obj event? Give it a signal, and forward it's arguments with a `SendSignal()` call. Now every component that want's to can also know about this happening.

#### Slippery things

At the time of this writing, every object that is slippery overrides atom/Crossed does some checks, then slips the mob. Instead of all those Crossed overrides they could add a slippery component to all these objects. And have the checks in one proc that is run by the Crossed event

#### Powercells

A lot of objects have powercells. The `get_cell()` proc was added to give generic access to the cell var if it had one. This is just a specific use case of `GetComponent()`

#### Radios

The radio object as it is should not exist, given that more things use the _concept_ of radios rather than the object itself. The actual function of the radio can exist in a component which all the things that use it (Request consoles, actual radios, the SM shard) can add to themselves.

#### Standos

Stands have a lot of procs which mimic mob procs. Rather than inserting hooks for all these procs in overrides, the same can be accomplished with signals

### Vars

1. `/datum/var/list/datum_components` (private)
    * Lazy associated list of type -> component/list of components.
1. `/datum/component/var/enabled` (protected, boolean)
    * If the component is enabled. If not, it will not react to signals
    * TRUE by default
1. `/datum/component/var/dupe_mode` (protected, enum)
    * How multiple components of the exact same type are handled when added to the datum.
        * `COMPONENT_DUPE_HIGHLANDER` (default): Old component will be deleted, new component will first have `/datum/component/proc/InheritComponent(datum/component/old, FALSE)` on it
        * `COMPONENT_DUPE_ALLOWED`: The components will be treated as separate, `GetComponent()` will return the first added
        * `COMPONENT_DUPE_UNIQUE`: New component will be deleted, old component will first have `/datum/component/proc/InheritComponent(datum/component/new, TRUE)` on it
1. `/datum/component/var/list/signal_procs` (private)
    * Associated lazy list of signals -> callbacks that will be run when the parent datum recieves that signal
1. `/datum/component/var/datum/parent` (protected, read-only)
    * The datum this component belongs to

### Procs

1. `/datum/proc/GetComponent(component_type(type)) -> datum/component?` (public, final)
    * Returns a reference to a component of component_type if it exists in the datum, null otherwise
1. `/datum/proc/GetComponents(component_type(type)) -> list` (public, final)
    * Returns a list of references to all components of component_type that exist in the datum
1. `/datum/proc/GetExactComponent(component_type(type)) -> datum/component?` (public, final)
    * Returns a reference to a component whose type MATCHES component_type if that component exists in the datum, null otherwise
1. `GET_COMPONENT(varname, component_type)` OR `GET_COMPONENT_FROM(varname, component_type, src)`
    * Shorthand for `var/component_type/varname = src.GetComponent(component_type)`
1. `/datum/proc/AddComponent(component_type(type), ...) -> datum/component`  (public, final)
    * Creates an instance of `component_type` in the datum and passes `...` to it's `New()` call
    * Sends the `COMSIG_COMPONENT_ADDED` signal to the datum
    * All components a datum owns are deleted with the datum
    * Returns the component that was created. Or the old component in a dupe situation where `COMPONENT_DUPE_UNIQUE` was set
1. `/datum/proc/ComponentActivated(datum/component/C)` (abstract)
    * Called on a component's `parent` after a signal recieved causes it to activate. `src` is the parameter
    * Will only be called if a component's callback returns `TRUE`
1. `/datum/proc/TakeComponent(datum/component/C)` (public, final)
    * Properly transfers ownership of a component from one datum to another
    * Singals `COMSIG_COMPONENT_REMOVING` on the parent
    * Called on the datum you want to own the component with another datum's component
1. `/datum/proc/SendSignal(signal, ...)` (public, final)
    * Call to send a signal to the components of the target datum
    * Extra arguments are to be specified in the signal definition
1. `/datum/component/New(datum/parent, ...)` (protected, virtual)
    * Forwarded the arguments from `AddComponent()`
1. `/datum/component/Destroy()` (virtual)
    * Sends the `COMSIG_COMPONENT_REMOVING` signal to the parent datum if the `parent` isn't being qdeleted
    * Properly removes the component from `parent` and cleans up references
1. `/datum/component/proc/InheritComponent(datum/component/C, i_am_original(boolean))` (abstract)
    * Called on a component when a component of the same type was added to the same parent
    * See `/datum/component/var/dupe_mode`
    * `C`'s type will always be the same of the called component
1. `/datum/component/proc/AfterComponentActivated()` (abstract)
    * Called on a component that was activated after it's `parent`'s `ComponentActivated()` is called
1. `/datum/component/proc/OnTransfer(datum/new_parent)` (abstract)
    * Called before the new `parent` is assigned in `TakeComponent()`, after the remove signal, before the added signal
    * Allows the component to react to ownership transfers
1. `/datum/component/proc/_RemoveNoSignal()` (private, final)
    * Internal, clears the parent var and removes the component from the parents component list
1. `/datum/component/proc/RegisterSignal(signal(string), proc_ref(type), override(boolean))` (protected, final) (Consider removing for performance gainz)
    * Makes a component listen for the specified `signal` on it's `parent` datum.
    * When that signal is recieved `proc_ref` will be called on the component, along with associated arguments
    * Example proc ref: `.proc/OnEvent`
    * If a previous registration is overwritten by the call, a runtime occurs. Setting `override` to TRUE prevents this
    * These callbacks run asyncronously
    * Returning `TRUE` from these callbacks will trigger a `TRUE` return from the `SendSignal()` that initiated it
1. `/datum/component/proc/ReceiveSignal(signal, ...)` (virtual)
    * Called when a component recieves any signal and is enabled
    * Default implementation looks if the signal is registered and runs the appropriate proc

## Examples
    Material Containers: #29268 (Too many GetComponent calls, but not bad)
    Slips: #00000 (PR DIS)
    Powercells: (TODO)

### DIFF
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -1,0 +1,16 @@
+//shorthand
+#define GET_COMPONENT_FROM(varname, path, target) var##path/##varname = ##target.GetComponent(##path)
+#define GET_COMPONENT(varname, path) GET_COMPONENT_FROM(varname, path, src)
+
+// How multiple components of the exact same type are handled in the same datum
+
+#define COMPONENT_DUPE_HIGHLANDER 0 //old component is deleted (default)
+#define COMPONENT_DUPE_ALLOWED 1    //duplicates allowed 
+#define COMPONENT_DUPE_UNIQUE 2     //new component is deleted
+
+// All signals. Format:
+// When the signal is called: (signal arguments)
+
+#define COMSIG_COMPONENT_ADDED "component_added"				//when a component is added to a datum: (datum/component)
+#define COMSIG_COMPONENT_REMOVING "component_removing"			//before a component is removed from a datum because of RemoveComponent: (datum/component)
+#define COMSIG_PARENT_QDELETED "parent_qdeleted"				//before a datum's Destroy() is called: ()
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -5,7 +5,7 @@
 // How multiple components of the exact same type are handled in the same datum
 
 #define COMPONENT_DUPE_HIGHLANDER 0 //old component is deleted (default)
-#define COMPONENT_DUPE_ALLOWED 1    //duplicates allowed 
+#define COMPONENT_DUPE_ALLOWED 1    //duplicates allowed
 #define COMPONENT_DUPE_UNIQUE 2     //new component is deleted
 
 // All signals. Format:
@@ -14,3 +14,5 @@
 #define COMSIG_COMPONENT_ADDED "component_added"				//when a component is added to a datum: (datum/component)
 #define COMSIG_COMPONENT_REMOVING "component_removing"			//before a component is removed from a datum because of RemoveComponent: (datum/component)
 #define COMSIG_PARENT_QDELETED "parent_qdeleted"				//before a datum's Destroy() is called: ()
+#define COMSIG_ATOM_ENTERED "atom_entered"						//from base of atom/Entered(): (atom/movable, atom)
+#define COMSIG_MOVABLE_CROSSED "movable_crossed"				//from base of atom/movable/Crossed(): (atom/movable)
--- a/code/datums/components/README.md
+++ b/code/datums/components/README.md
@@ -1,0 +1,101 @@
+# Datum Component System (DCS)
+
+## Concept
+
+Loosely adapted from /vg/. This is an entity component system for adding behaviours to datums when inheritance doesn't quite cut it. By using signals and events instead of direct inheritance, you can inject behaviours without hacky overloads. It requires a different method of thinking, but is not hard to use correctly. If a behaviour can have application across more than one thing. Make it generic, make it a component. Atom/mob/obj event? Give it a signal, and forward it's arguments with a `SendSignal()` call. Now every component that want's to can also know about this happening.
+
+### In the code
+
+#### Slippery things
+
+At the time of this writing, every object that is slippery overrides atom/Crossed does some checks, then slips the mob. Instead of all those Crossed overrides they could add a slippery component to all these objects. And have the checks in one proc that is run by the Crossed event
+
+#### Powercells
+
+A lot of objects have powercells. The `get_cell()` proc was added to give generic access to the cell var if it had one. This is just a specific use case of `GetComponent()`
+
+#### Radios
+
+The radio object as it is should not exist, given that more things use the _concept_ of radios rather than the object itself. The actual function of the radio can exist in a component which all the things that use it (Request consoles, actual radios, the SM shard) can add to themselves.
+
+#### Standos
+
+Stands have a lot of procs which mimic mob procs. Rather than inserting hooks for all these procs in overrides, the same can be accomplished with signals
+
+## API
+
+### Vars
+
+1. `/datum/var/list/datum_components` (private)
+    * Lazy associated list of type -> component/list of components.
+1. `/datum/component/var/enabled` (protected, boolean)
+    * If the component is enabled. If not, it will not react to signals
+    * TRUE by default
+1. `/datum/component/var/dupe_mode` (protected, enum)
+    * How multiple components of the exact same type are handled when added to the datum.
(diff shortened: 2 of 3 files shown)


Cruix:
The datum component system (DCS) accomplishes two Principles of design:
1. polymorphism by composition rather than inheritance
2. dynamic allocation of behaviors at runtime

Both of these principles are accomplished more efficiently and more simply without this system.

Principle 1 can already be done in Byond by using either vars or procs.

Polymorphism by composition can be accomplished by adding a variable containing an instance of something (in theory this should usually be a lightweight datum, but in practice it often ends up being an atom either because of Byond's special handling of atoms or because people have shoehorned the composition strategy into something that was originally done by inheritance) that handles an object's compositional behavior for it. An example of this is traitor uplinks. All datums of type /obj/item have a var called hidden_uplink that stores a traitor uplink, and when an instance of that datum is handled in a certain way (such as typing a particular password into a PDA's messenger), it will trigger the uplink's access event by calling a proc on it, if the uplink exists. The code that handles the trigger is put on the datum that holds the uplink rather than the uplink itself, because the object holding the uplink understands its own behavior and can most efficiently trigger the "access traitor uplink" event, while the uplink is just along for the ride and does not particularly care how or when it is accessed. In turn, the datum holding the uplink does not have anything to do with the uplink's operation other than sending the access event, because it does not know or care how the uplink works.

The downside of this is that all instances of type /obj/item now have an additional variable. The way Byond's object system works, this var does not take up any additional memory on instances that do not use it, but it does slow down the accessing of other variables that each instance is not using. This is strictly an issue with how Byond handles variables and could potentially be optimized away by Lummox at any time.

An alternative to this is the proc system that the get_cell() example in the DCS documentation cites. This is more efficient because procs are accessed with a binary search rather than a linear one, so in order to slow down the accessing of a proc by an increment of 1, you have to double the number of procs a type has. Even if we had a tremendous number of compositions that might be on any datum, which we do not, it would not significantly affect performance. This system is not used as much because it is syntactically easier to use a variable than a proc, because Byond does not know (or rather, in the style of byond, it does not by default <i>assume</i>) the type of a proc's return value at compile time, and so you must always typecast it.


A perfect example of Principle 2 in existing code is the species datums for /mob/living/carbon/human (actually it is stored in a mob's dna, but it is currently only ever used by humans). It was originally made to handle changing a mob's species at runtime without having to copy over a large number of variables and external references to a new mob whenever a species was changed. The DCS is similar to datum species in many ways, other having a wider scope and having a list of components rather than a single species. However, the DCS using signals while datum species uses procs.

Signals are worse than procs because they add an additional layer of indirection for absolutely no gain. Not only do you still have proccall overhead, you have doubled it, because you have to call SendSignal() which then calls whatever proc it looks up. The only advantage of this is that you can dynamically add and remove procs from a particular signal at runtime, which is unnecessary for a system that should in the first place only be made up of atomic components (that is, components whose behavior cannot reasonably be split up into multiple independent behaviors).

Additional overhead is added because every single component must receive every single signal, even the ones it does not care about. Every time an atom is crossed, it must send COMSIG_MOVABLE_CROSSED to every component it has, whether that is the slippery component or the battery component. In theory this could be optimized away with event subscription lists, but that would have the cost of additional memory usage on every instance using components.

I have not seen an example or theory on /tg/ where the signal system would ever be used meaningfully. Signals are just a completely redundant component system for a component system.


Yogstation has had a system remarkably similar to DCS, though more limited in scope, for almost a year now. #1042 added the module framework to the current repository, and it was on the old pre-rebase repository before that. It has, to my knowledge, never been used since its introduction, either because it has not been necessary or because it is easier to add hooks where you need them than to expand an existing system that uses similar hooks.


I believe that, if the DCS were added to this repository, it would either never be used or it would only be used in cases where an alternative would have been better.

And, in the words of Forrest Gump, "That’s all I have to say about that".

ghost:
rip then i'll just shitcode my PR to work

optimumtact:
good post Cruix

PR #2509, opened 2017-08-24
Distinguishes special antag talk, and also logs clock cult talk

Amelia0010:
Instead of server logs saying "SAY: [message]" when someone talks over shadowling hivemind, it actually says "SHADOWLING: [message]" or "THRALL: [message]" or whatever.

It also adds logging for clockcult chat, as that wasn't logged before, unless you bothered patching together the whispers that came shortly before it
And now whispers aren't logged twice either

#### Changelog

:cl:
tweak: Say logs now distinguish whether it's said normally, or over cult talk, or blob hivemind or etc.
/:cl:

### DIFF
--- a/code/__HELPERS/_logging.dm
+++ b/code/__HELPERS/_logging.dm
@@ -22,7 +22,7 @@
 /proc/log_adminsay(text)
 	if (config.log_adminchat)
 		log_admin("ASAY: [text]")
-		
+
 /proc/log_dsay(text)
 	if (config.log_adminchat)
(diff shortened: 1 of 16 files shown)


ghost:
why should admins be able to approve their own station name

Amelia0010:
why shouldn't they

Super3222:
If you’re playing the game you shouldn’t be adminning

Jc1469:
if only it were that simple.

retmins have that luxury, but an admin on alone, a good one atleast, can admin and play at the same time.

PR #2513, opened 2017-08-26
Made wizards have more objectives

Cruix:
Wizards, by design the most powerful solo antagonists in the game, very often have only one murder objective after which they can sleep in a locker for the rest of the round. Now they will have more.

Wizards now have a 30% chance of getting hijack and 3 steal objectives, a 10% chance of getting 3 murder objectives and die a glorious death, and a 60% chance of getting 6 objectives, 2 to 4 being steal and the rest being murder, and stay alive until the end. They also have a 10% chance of getting a protect objective instead of a murder objective.

:cl:
tweak: wizards now have more objectives.
/:cl:

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -755,7 +755,7 @@ var/global/list/possible_items_special = list()
 	return 0
 
 /datum/objective/summon_magic
-	explanation_text = "Be the last living, human adept on the station."
+	explanation_text = "Be the last living, humanoid adept on the station."
 
 /datum/objective/summon_magic/check_completion()
 	if(!ishuman(owner.current) || (owner.current.stat & DEAD))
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/objective.dm
```diff
@@ -755,7 +755,7 @@ var/global/list/possible_items_special = list()
 	return 0
 
 /datum/objective/summon_magic
-	explanation_text = "Be the last living, human adept on the station."
+	explanation_text = "Be the last living, humanoid adept on the station."
```
ShadowDeath6, line 758: do you even need humanoid? just have them be the last living adept

Cruix: I was worried about people getting staff of changed into a mouse and hiding in a vent all round.

PR #2515, opened 2017-08-28
Makes the only clockcult objective possible to summon

ghost:
Seriously, why doesn't their god want to be saved from Reebee?

:cl:
tweak: Makes the only clockcult objective possible to summon
/:cl:


Kmc2000:
makes sense

ghost:
probably better than 'blow the borgs' or 'leave without doing anything

PR #2523, opened 2017-08-31
Replaced datacore manifest mob reference with a mind reference

Cruix:
This could cause weird edge cases like late-round traitors never targeting someone who has been cloned. Minds are a more reliable indicator of identity than mobs, because minds carry over through operations like cloning and mindswap.


ShadowDeath6:
does this influence anything like borged traitor objectives counting as dead?

Cruix:
No, the only part of objectives that is affected is the selection of a target when the objective is created.

PR #2524, opened 2017-08-31
6 fixes, for the price of 1!

### LINKED ISSUE #2344: Asay error does not promt you with flags you need to use asay
Sorry, what flags i need? Plz tell, . is not a flag, really
image

### LINKED ISSUE #2428: Toilets have disposal bin overlays
untitled

Amelia0010:
Toilets don't have overlays anymore.
Airlock light wire now says lighting and not idscan.
Asay now says only admins can use it, rather than you needing no flags.
show_admin_log now only asks you once, not twice
And a slight grammarfix to examine
Wrenching an inactive AI core now accurately states whether you're fastening or unfastening

fixes #2428 
fixes #2344 

#### Changelog

:cl:
fix: Toilets have been refitted to have a regular toilet seat, instead of a plate of metal!
fix: Asay hotkey now doesn't say you need no flags to use it
fix: Wrenching an inactive AI core now accurately states whether you're fastening or unfastening
/:cl:


code/modules/admin/verbs/adminsay.dm
```diff
@@ -2,7 +2,8 @@
 	set category = "Special Verbs"
 	set name = "Asay" //Gave this shit a shorter name so you only have to time out "asay" rather than "admin say" to use it --NeoFite
 	set hidden = 1
-	if(!check_rights(0))
+	if(!src.holder)
+		src << "Only administrators may use this command."
```
Cruix: 1. This is assay, not dsay
2. We might want to add rights to asay in the future

Amelia0010: 1. I'll fix the bit about deadsay, i'm dumb
2. deadsay just says that only administrators can use it, which is why I went with that for this, rather than an actual check for flags. And the fact that there are no flags to check for

PR #2530, opened 2017-09-02
Wizards now get the badass symbol if they complete all their objectives without any spells

Cruix:
Because why not.

### DIFF
--- a/code/game/gamemodes/wizard/wizard.dm
+++ b/code/game/gamemodes/wizard/wizard.dm
@@ -226,8 +226,10 @@
 			else
 				text += "<br><font color='red'><B>The wizard has failed!</B></font>"
 				feedback_add_details("wizard_success","FAIL")
-			if(wizard.spell_list.len>0)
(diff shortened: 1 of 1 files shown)


Ratvar:
You are a badass if you completely ignore your majuuk theme and instead go lame stealth aka normal name + suit + lost ID giv new + hide in locker route/ traitor hjack with guns from armory and plasma in distro route, converting round into extended/traitor?

Cruix:
Maybe I should phrase it:
"You get the badass symbol if you don't go jaunt/disintigrate/timestop/raise dead, and also don't get any of the spells that no one ever uses."

Now that a wizard needs to complete six objectives, hijack, or kill three people and end the round immediately by suiciding gloriously, stealth wizards will probably not be as much of a problem.

Things like apprentices and artifacts do not could as spells, so you can still run around with a staff of change and wand of death, you just do not have access to all of the get-out-of-jail free abilities that cannot be disarmed and can be used while stunned.

PR #2554, opened 2017-09-09
Fixed sheet subtypes not stacking together

Cruix:


Partheo:
I have an update coming up that affects the same files for stacking purposes. What do these changes do, exactly?

Cruix:
They make it so that subtypes of stacks that exist only for convenience, such as /obj/item/stack/sheet/metal/fifty, will stack with each other.

Partheo:
Got it, thanks.

PR #2569, opened 2017-09-12
Mutes the colors of space a bit

Partheo:
So yeah. This is a bit of a shot in the dark, what do people think of this?

output_bqqs2o

I muted all the large cyan stars so they are grayscaled now. This is a before/after gif, with the cyan stars being what it looks like currently. I think it's an improvement, some people have said they like it so far. If the reception is positive then this is an easy change.

:cl:
imageadd: Space has been unsaturated a little.
/:cl:


Notamaniac:
This PR has a whole bunch of other changes in it, (refactors by the look of it, but I'm code blind), you might want to add that in the PR in case stuff breaks and someone needs to point the finger at who broke it.

Oh and neutral on the whole cyan star thing; you might say it's more realistic to have less blue stars but the artstyle isn't ultra realism

Partheo:
Right, the extra commits are because I accidentally branched this off of one of my other updates, which are also PRd here. If needed I can remake this, it's a quick change.

I did this for a bit of realism and curiosity, and I kind of liked the result.

ShadowDeath6:
color

>AAC2C6

try this instead of completely gray, it might look a bit nicer

Partheo:
Going to re-do this with less of a mess

PR #2579, opened 2017-09-14
Nerfs telebaton

ghost:
telebatons can no longer stunlock

:cl:
tweak: Telebatons no longer stunlock
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/melee/misc.dm
+++ b/code/game/objects/items/weapons/melee/misc.dm
@@ -61,7 +61,7 @@
 					if (H.check_shields(0, "[user]'s [name]", src, MELEE_ATTACK))
 						return
 				playsound(get_turf(src), 'sound/effects/woodhit.ogg', 75, 1, -1)
-				target.Weaken(3)
+				target.Weaken(2)
 				add_logs(user, target, "stunned", src)
 				src.add_fingerprint(user)
 				target.visible_message("<span class ='danger'>[user] has knocked down [target] with \the [src]!</span>", \


Super3222:
Can we get an explanation? Because the way I see it this is not happening.

Notamaniac:
I have a feeling this is going to make the telebaton completely useless.

You do know the telebaton was added to replace the nerfed flash heads got right?

Amelia0010:
Isn't the stuntime already so short that you have to begin cuffing immediately to do it before they get up?

Jc1469:
:policeman: b a l a n c e   p o l i c e :police_car: 


no. this is a bad pr. The ability to stunlock is already unreliable.

There's no gain to this other than "weh i break in to science and rd beat me up every time, nerf because unfun i want greytide easy"

X-TheDark:
Yeah, an explanation for this is definitely needed.
Why do you think permastun from this is bad, considering that's sort of the point of this baton, for heads to be able to defend themselves without needing to call sec.

Also, numbers:
* This is a 3 BYOND tick stun (6 seconds, as it calls the `stun = stun-1` thing once every 2 seconds), with a cooldown of 4 seconds...usually. Weird stuff can happen if you hit the target just as a tick is about to be called...

Chuggachar:
Yeah I don't see a problem with the telebaton, the stunlock is already unreliable as john said.

ToGWtF:
:salt: PR

ShadowDeath6:
the stun is really unreliable since there's a 1 second window where you can actually hit them with the baton rather than the 6 seconds you get with sec's stunbaton

Cruix:
The whole point of telebatons is to stunlock

PR #2596, opened 2017-09-17
Gold slimes can no longer spawn magicarp

matskuman5:
Magicarp and chaos magicarp do nothing but grief the station. Chaos magicarp just spam their staff everywhere and "accidentally" kill 5 people with a fireball bolt

:cl:
rscdel: Gold slimes can no longer spawn magicarp.
/:cl:

### DIFF
--- a/code/modules/events/wizard/magicarp.dm
+++ b/code/modules/events/wizard/magicarp.dm
@@ -37,6 +37,7 @@
 	projectilesound = 'sound/weapons/emitter.ogg'
 	maxHealth = 50
 	health = 50
+	gold_core_spawnable = 0
 
 /mob/living/simple_animal/hostile/carp/ranged/New()
 	projectiletype = pick(typesof(initial(projectiletype)))


code/modules/events/wizard/magicarp.dm
```diff
@@ -37,6 +37,7 @@
 	projectilesound = 'sound/weapons/emitter.ogg'
 	maxHealth = 50
 	health = 50
+	gold_core_spawnable = 0
```
Qustinnus: use FALSE plis

Ratvar:
what about removing blobbernauth who just walk over blobs and shreck them?

Partheo:
Good change

Ratvar:
...will it enfasten process of getting headslug cause mobpull is smaller now?

ShadowDeath6:
there are rules on playing as sentient creatures, just punish the people who grief and stop removing fun for everyone else

Kmc2000:
👎  removes fun, how will xenbios fight the nukeops now?

Time-Green:
I also enjoy them, but they're so horribly overpowered I cant believe it

Partheo:
KMCs tank is also an example of something fun but there are issues with it for a reason. Same for the carps. 90% of the time they are just a nightmare for the crew and only fun for the person playing them. They're absurdly overpowered, and a major cause of grief. 

Sure, we could just ban everyone who griefs with it. When most cases of use with the carps are grief though, there is a problem. I think this is a great change.

X-TheDark:
Maybe a compromise would be to make the sentient magicarp be able to choose a power? Or let the "master" choose it when the magicarp is sentiented.

The only beneficial thing is the "revive blast", so that's for non-antags. Rest can be used by antag Xenos (If you have nothing better to do with your antag, that is).

Until I code that in, though, this could be considered a stopgap for now.

Partheo:
The thing with a revive blast is that it's also absurdly overpowered. Think of a nuke-ops round where the crew have a reviving magicarp.
It's really, really unbalanced.

Time-Green:
@X-TheDark wouldn't that only make it worse?

ShadowDeath6:
half of the fun of getting a magicarp is having it use its magic and hoping you get something fun like heal or change

a better idea might be to remove death bolt carp from spawning since that's the most influential / overpowered type

Jc1469:
I agree to saying fuck magicarp entirely.  summon events or no fun allowed

ghost:
chaos carp are fairly pathetic

regular magicarp certainly need to go

KrabSpider:
Rather not. Having one of either makes my job somewhat easier.

Ratvar:
Stop working as security/medical/staff_of_antagonising_into_ayy/spawndoors, Kreb, you are jelly feeder

PR #2623, opened 2017-09-23
Ninja fix

Time-Green:
Ninja research objective actually works now
:cl:
fix: Ninja research objective works now
/:cl:
PR atomized due to popular demand by literally only adam


ghost:
ninja is supposed to be tough and stealth is supposed to be strong for it's energy cost

make stealth flicker when they attack or something briefly instead, enough to reveal location at the time

PR #2631, opened 2017-09-28
Added new lobby music

Moja10:
Song made by me. Made for an abandoned game and I think it would fit.
Comments/criticism welcome. Didn't have a name so I just named it Plasma Attack but if you think of a better one/it doesn't matter let me know.

Link to song: https://soundcloud.com/jeffimam/title-plasma-attack

:cl:Yogurtshrimp69
soundadd: Added new lobby song
/:cl:


Partheo:
Sounds good 👌
I'm slightly worried about the duration though. The lobby wait time is 180 seconds, and your track is 175 seconds long. However, I'm also almost certain that BYOND seconds are completely different anyway so it probably doesn't matter.

Moja10:
If the song was for some reason longer than the lobby time, wouldn't it just cut off? I'm pretty sure it does that with space oddity

ghost:
yog should probably consider removing some of the title music, or sizing down the filesize of some of the music. You might not care but a lot of people from countries with shittier internet (brazil, east europe, australia) will have a shit time downloading the resource pack.

Moja10:
I can lower the quality and not make it sound like shit because it's 8bit. I actually didn't check the filesize but I can do that no problems. 

EDIT: I just checked and it's 1.8 MB, which compared to the other lobby music, is about the same size. If I compress it the furthest, I can only get it to 900kb, which is decent, but stuff like Mr. Bluesky is 3.8 Mb. Is it really necessary to compress it further?

Partheo:
Compress it as much as you can without sacrificing quality too much. It's just good practice.

PR #2639, opened 2017-10-03
ports /tg/'s fix to dodging radstorms with lockers

ghost:
now you can't hide in lockers to dodge radstorms or ashstorms
#### Changelog

:cl:
rscadd: Lockers and bodybags and such no longer protect against radstorms and ashstorms, as is proper.
/:cl:

### DIFF
--- a/code/datums/weather/weather.dm
+++ b/code/datums/weather/weather.dm
@@ -114,7 +114,7 @@
 
 /datum/weather/proc/can_impact(mob/living/L) //Can this weather impact a mob?
 	var/turf/mob_turf = get_turf(L)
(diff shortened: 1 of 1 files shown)


Partheo:
If you want to include a pop culture reference- do this for all lockers except the chef's fridge.
indie

X-TheDark:
Weird, being in a locker/container shouldn't change a mob's z-level (which is what your PR addresses).

Alek2ander:
Stuff only has actual coordinates when it's on a turf. If something is inside a container its coords are (0,0,0). That's why get_turf exists.

X-TheDark:
Oh, damn, didn't know that.

PR #2643, opened 2017-10-04
Logs sleepers, also... fixes.. emagging them.

Jc1469:
-audible_message("<span class='alert'>[src] begins to overloading and goes REEEEEEEEEEEEEEEEEEE!</span>")
:cl:
tweak: Emagging sleepers now puts people to sleep.
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -201,7 +201,7 @@
 					occupant.attack_log += text("\[[time_stamp()]\] <font color='orange'>[occupant]/[occupant.ckey] has been injected by [usr]/[usr.ckey] with 30 [reagent_id] using an EMAGGED sleeper</font>")
 					log_attack("\[[time_stamp()]\] [usr]/[usr.ckey] injected [occupant]/[occupant.ckey] with 30 [reagent_id] using an EMAGGED sleeper")
 					usr.attack_log += text("\[[time_stamp()]\] <font color='red'>[usr]/[usr.ckey] has injected [occupant]/[occupant.ckey] with 30 [reagent_id] using an EMAGGED sleeper</font>")
-					audible_message("<span class='alert'>[src] buzzes and beeps madly!</span>")
+				audible_message("<span class='alert'>[src] buzzes and beeps madly!</span>")
 				emag_effect = 1
 				open_machine()
 				return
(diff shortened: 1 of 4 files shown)


code/game/machinery/Sleeper.dm
```diff
@@ -198,16 +198,22 @@
...
 				for(var/reagent_id in available_chems)
 					occupant.reagents.add_reagent(reagent_id, 30)
-					audible_message("<span class='alert'>[src] begins to overloading and goes REEEEEEEEEEEEEEEEEEE!</span>")
-					emag_effect = 1
-					open_machine()
-					return
+					occupant.attack_log += text("\[[time_stamp()]\] <font color='orange'>[occupant]/[occupant.ckey] has been injected by [usr]/[usr.ckey] with 30 [reagent_id] using an EMAGGED sleeper</font>")
+					log_attack("\[[time_stamp()]\] [usr]/[usr.ckey] injected [occupant]/[occupant.ckey] with 30 [reagent_id] using an EMAGGED sleeper")
+					usr.attack_log += text("\[[time_stamp()]\] <font color='red'>[usr]/[usr.ckey] has injected [occupant]/[occupant.ckey] with 30 [reagent_id] using an EMAGGED sleeper</font>")
+					audible_message("<span class='alert'>[src] buzzes and beeps madly!</span>")
```
X-TheDark: Might want to move this outside the `for` loop, otherwise it'll show for every reagent in the sleeper.

Jc1469: That's the idea.  The entire loop can only trigger once per sleeper regardless because it breaks the sleeper afterwards, so it's a pretty good wakeup call as it injects the occupant with some nice death.

Super3222:
REEEEEEEEEEEEEEEEEEE!

Jc1469:
I hate it when things begin to overloading

Partheo:
What chem does emagging add again?

KrabSpider:
@Partheo Toxin

code/game/machinery/Sleeper.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: You can fix emagged sleepers with wirecutters so it has potential to show up multiple times

Cruix: Put it outside the loop so upgraded sleepers do not fill people's screens with messages

ShadowDeath6:
emagging sleepers will cause them to inject 40 of every chemical in their storage

PR #2650, opened 2017-10-05
Nerfs the changeling LSD sting

Amelia0010:
WELCOME, LADIES AND GENTLEMEN, TO NICH'S SALT PR SHOW. IN THIS EPISODE, WE'LL BE LOOKING AT THE CHANGELING LSD STING.

Nerfs the changeling hallucination sting by reducing the amount of mindbreaker it injects. It now cuts the amount of mindbreaker in half, meaning I'll only crash every 30 seconds for 5 minutes, instead of 10!

#### Changelog

:cl:
tweak: the changeling LSD-sting has been nerfed
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/powers/tiny_prick.dm
+++ b/code/game/gamemodes/changeling/powers/tiny_prick.dm
@@ -220,7 +220,7 @@
 	feedback_add_details("changeling_powers","BS")
 	return 1
 
-/obj/effect/proc_holder/changeling/sting/LSD
+/*/obj/effect/proc_holder/changeling/sting/LSD
 	name = "Hallucinogenic Sting"
 	desc = "Causes terror in the target."
 	helptext = "We evolve the ability to sting a target with a powerful hallucinogenic chemical. The target does not notice they have been stung, and the effect occurs very quickly."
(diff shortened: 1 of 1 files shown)


Jc1469:
I wouldn't bother.  The worse part is the crash, which I'm pretty sure is caused by dreams when they fake knock you out.

And dreams will be gone soon.

ShadowDeath6:
buttblasted

ghost:
They can just sting twice, then.

Time-Green:
@Jc1469 the crashing isn't caused by dreams, people were crashing long before dreams were added

Amelia0010:
> They can just sting twice, then.

Which would double the amount of chemicals they need for the same effect, which I believe is called nerfing it.

ghost:
Alright mr. smartass, here's my point.

this is pointless, because a ling will still just spam sting and chem regen is fast enough for this to be barely noticeable.

Increase chemical cost too, and then they can't dualsting quite as easily.

X-TheDark:
If hallucinations crash the game, wouldn't it be better to just replace the whole mechanic with something else?

PR #2653, opened 2017-10-06
You can no longer kill entire departments with a single shotgun dart hitting one guy in the corner of a room

ghost:
Shotgun darts were noreact.

As a result, this means you could fill them with the reagents for a plasma smoke + clf3 mix. Which, because it's plasma smoke, would promptly flood most of a department and then set it on fire. From one dart.

For obvious reasons, cryo syringes are supposed to be the only items with this capability, and can only carry 15u, as opposed to the fucking 50 that shotgun darts can carry at roundstart.

Bye, shotgun darts.
#### Changelog

:cl:  
tweak: Shotgun darts now no longer store their chemicals in an alien subdimension and as a result will have chemicals react inside them rather than only when they hit a mob. 
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullets.dm
+++ b/code/modules/projectiles/projectile/bullets.dm
@@ -177,7 +177,7 @@
 /obj/item/projectile/bullet/dart/New()
 	..()
 	create_reagents(50)
-	reagents.set_reacting(FALSE)
+	reagents.set_reacting(TRUE)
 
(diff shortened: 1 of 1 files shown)


Kmc2000:
Ok

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Add "REMOVES FUN" tag please

Jc1469:
Couldn't you just remove the reaction line seeing as it's a reagent container and they by default react?

PR #2656, opened 2017-10-07
Replaces "Absorb X compatible genomes" with "Absorb the genome of X person"

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Back when the game was actually good, it was a game of fear and paranoia of husked corpses turning up in maint as the changeling neared its goal of absorbing 8 people. The advent of extract DNA sting and the murderbone rule has turned the objective from using stealth and cunning into clicking 7 people on the escape shuttle. I hope to bring some of the old gameplay with this PR.

:cl:  
tweak: Changelings now have to absorb the genome of a specific person instead of extracting several compatible genomes.
/:cl:


Iamgoofball:
lmao what

git gud :-1:

1fbff5f83b23d39d38b1dfcb4cac8d9b:
thank you for your constructive feedback

Iamgoofball:
also make dna sting not count for this and you've fixed the problem without making ling easymode

1fbff5f83b23d39d38b1dfcb4cac8d9b:
thank you for actual constructive feedback

Nickvr628:
I like the change.

Ratvar:
Will horrorform absorbs count?

Notamaniac:
Might want to lower the absorb bounds to 3-5. If we have something like 3-4 changelings we have somewhere between 12-24 crew having to die every single ling round (which is about 20-40% of the crew if we're assuming a 60 crew round, and all the lings don't go for monkeyman absorbs).

Edit : Nvm misread it, at best this is just another murder objective so eh

Alexandria404:
If there are multiple lings, is there a check built in so they don´t have to absorb the same person? same question for traitor objectives.
Also to make sure I fully understand this.
The get a certain amount of DNA sample will be replaced with absorb ONE person, right?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Yes Alex, each ling has to absorb one person. There is no check whether multiple changelings have to eat the same person. Traitors also can get an objective to kill the same person, steal the same thing etc.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Those who are thumbing down, can you provide a better alternative to an objective which is at best a stupid speedbump, and at worst banbait for people who missed the rules on absorbing? (inb4 it's their fault)

Ratvar:
Removal?

ShadowDeath6:
Make sure it can't be overwritten like the "escape as x" objective if you aren't already doing that

also I would kind of like this to be 2 absorbs because slightly more challenging

ShadowDeath6:
I just read the code, but I can't edit so you can disregard the overwriting thing

Dagdammit:
Not one of those who thumbed down, but I might be able to offer a better solution?

Track how long the Ling's been in the form of a given person. When they've been that person for 5 minutes (possibly 5 uninterrupted minutes), inform them that they've successfully "mastered" that form. 

Make the objective to master at least X (5?) forms. Also maybe start them with no evolution points and have each mastered form give them 2, potentially up to a max of 10.

The most amazing ling experience I've ever had was a highly unrobust player who kept coming back to medbay & trying to attack me. The reason it was so amazing was that he always came back as someone else. My boss, a sec officer, a janitor... everyone who came into medbay could be a potential threat, and the paranoia this inspired once my dumb ass started to adjust was amazing. Impersonating people should be a central part of what lings do, IMO. Instead a ling will often wear its roundstart face for ages simply because nothing pushes it to bother with disguises.

Ratvar:
Not a fan about "You need 25 minutes or you fail automatically", but it sounds ass great.

Dagdammit:
I'm fine with it because it becomes part of the challenge. Adds a
significant risk to changeling players who go loud/cause chaos too soon,
and they have plenty of tools to help with preventing the escape shuttle's
arrival.

ghost:
Not plausible or reasonable, Dagdammit.

A better option would simply be to remove DNA sting and lower the amount of required absorbs, so that each ling doesn't need to husk 7 people.

Dagdammit:
I'd be interested to hear details behind that judgement, Adam. (be it here
or elsewhere)

Agreed that unless there's some big overall change, DNA extract sting
probably needs to go.

ghost:
@Dagdammit Enforces nothing but a ling transforming into someone and sitting in a corner doing nothing. Makes completing an objective virtually impossible if you're unlucky enough to arrive as a changeling with the shuttle on 4:59 or so forth. Really doesn't add anything whatsoever.

The mastering objective shit is dumb. Also, having a ling gain more evo points the more they absorb is suicidal. Go look at hippie or /vg/ for an example as to why. It just snowballs until the ling is practically unkillable.

Ratvar:
Not if ling starts with with two points and maxes at ten of current ling

Dagdammit:
@Ratvar yeah that's about what I had in mind. Could throw in other balancing factors if they proved necessary, like armblades & similar "overt" flesh-shaping powers only being possible while you're in a "mastered" form.

@AdamElTablaway Solid points. Thanks for providing specific reasons, feel free to ignore these responses if you don't feel inclined to go into it further. 

"Enforces nothing but a ling transforming into someone and sitting in a corner doing nothing."
I don't mind that being an *option* for ling players, though if it became the strategy of choice (and was also real boring for the ling player) I'd certainly consider that a problem in need of fixing. Some disadvantages to this approach, though:
-Not very compatible with acquiring additional forms. (Could be wise to give Extract DNA sting a multi-minute cooldown each time it gets you a new form, or just cut down hard on the number of unmastered forms a ling can simultaneously retain)
-Tricky to explain if you're caught, i.e. "why is the chef hiding in a random locker/isolated corner of maint". 
-Time spent hiding in a corner doing nothing = time not spent obtaining further advantages (obtaining equipment & access, spinning webs of lies, pursuing your other objectives)

"virtually impossible if you're unlucky enough to arrive as changeling with the shuttle on 4:59 and so forth."
I hadn't even thought about midround changelings. You're right, that'd be a problem. I think you'd either want to not give them that objective, or give it to them with a reduced headcount based on how long the round's gone on (and still swap it for a different objective if the station's on red alert or the shuttle's already been called).

code/game/gamemodes/objective.dm
```diff
@@ -704,13 +704,37 @@ var/global/list/possible_items_special = list()
...
+		if(ishuman(target.current))
+			var/mob/living/carbon/human/H = target.current
+			spec = (H.dna && H.dna.species) ? "[H.dna.species.name] " : ""
+		explanation_text = "Absorb the genome of [target.name], the [spec][target.assigned_role]."
+	else
+		explanation_text = "Free Objective"
 
+/datum/objective/absorb/check_completion()
+	var/mob/living/carbon/human/TC = null
+	if(target && target.current)
```
AshCorr, line 729: What if the target's body has been gibbed? Their `current` variable will be set to null. You might just want to toggle the variables completion as soon as the changeling absorbs the individual.

PR #2660, opened 2017-10-07
adds fake secret documents

Amelia0010:

Replaces the documents in the syndicate maintenance ruin with ones which can't be used for the objective.

#### Changelog

:cl:  
fix: Syndicate maintenance documents no longer count for the objective, cause why would they want you to steal something that they left there.
/:cl:


Dagdammit:
Trying to parse the changes.
- Is there a way to tell if the documents you have are fake or not? It looks like lack of a seal means fake, but seals can also be fake since they can be forged?
- Does this mean photos and photocopies of secret documents no longer fulfill the objective?

Amelia0010:
This only means that the documents inside the maintenance room which spawns with syndie stuff is fake.

All other documents still work, including photocopying and the documents in the vault

Partheo:
Makes maintenance less interesting 👎

Amelia0010:
> Makes maintenance less interesting 👎

How? The documents are still there, and identical, now they just don't function for the traitor objective.

Partheo:
It's less interesting because you've removed a unique point of contention that could occur in maintenance. You've stripped all non-RP gameplay that could occur as a result of those documents being present in maintenance.

Jc1469:
walk into maintenance. pick up document. leave maintenance.

Where's the RP?

Partheo:
I wasn't talking about RP, hence me mentioning non-RP.

What I meant was that the room previously held a gameplay element, and now it doesn't. You could have interesting situations arise from sec finding someone in the room, for example.

Also, if they're identical then there's a problem. IRL if two documents are exactly the same, then they function exactly the same. Here you've added a document which looks exactly the same as the real documents, yet for some reason the syndicate won't accept them. Either edit the text or add some description that indicates that they are fake.

Amelia0010:
Now the fake documents have a plastic seal rather than a wax one.

And having an objective called "rush to one place in maint" isn't really that great. Plus, the documents sold for what, 50.000 credits, so this one small maintenance room made both cargo and traitoring immensely easy, because of a themed room.

Partheo:
Didn't know about the cargo points. Taking everything into consideration, I support this PR. Making maintenance interesting is important, but placing a traitor objective inside isn't the way to go about it.

PR #2672, opened 2017-10-11
buffs blob a large amount like, wow 

Jc1469:
no more afk blobs pls


Partheo:
Game breaking

👍👍👍

drderp3635:
Blob powercreep

Cruix:
Make it do a poll instead

Jc1469:
Making it do a poll makes there be a wasted 30 second period of a blob spending the points to split, so I'm not doing that. And a lower timer risks the window not loading and half the dead ghosts outright not having a chance to begin with.

PR #2682, opened 2017-10-12
Everyone can see the "Blobs to Win" in status once level 5 biohazard is announced

ghost:
:cl:  
rscadd: Everyone can see the "Blobs to Win" in status once level 5 biohazard is announced  
/:cl:

### DIFF
--- a/code/game/gamemodes/blob/blob.dm
+++ b/code/game/gamemodes/blob/blob.dm
@@ -89,17 +89,14 @@ var/list/blobs_legit = list() //used for win-score calculations, contains only b
 
 	spawn(0)
 		var/message_delay = rand(messagedelay_low, messagedelay_high) //between 4 and 6 minutes with 2400 low and 3600 high.
-
-		sleep(message_delay)
-
-		send_intercept(1)
-		message_sent = 1
-
-		sleep(24000) //40 minutes, plus burst_delay*3(minimum of 6 minutes, maximum of 8)
-
-		send_intercept(2) //if the blob has been alive this long, it's time to bomb it
+		addtimer(src, "announce_bleb", message_delay)
 
 	return ..()
+
+/datum/game_mode/blob/proc/announce_bleb()
+	send_intercept(1)
+	message_sent = 1
+	addtimer(src, "send_intercept", 24000, 2)
 
 /mob/living/Stat()
 	..()
--- a/code/game/gamemodes/blob/blob.dm
+++ b/code/game/gamemodes/blob/blob.dm
@@ -30,7 +30,9 @@ var/list/blobs_legit = list() //used for win-score calculations, contains only b
 
 	var/messagedelay_low = 2400 //in deciseconds
 	var/messagedelay_high = 3600 //blob report will be sent after a random value between these (minimum 4 minutes, maximum 6 minutes)
-
+	
+	var/message_sent = 0
+	
 	var/list/blob_overminds = list()
 
 /datum/game_mode/blob/pre_setup()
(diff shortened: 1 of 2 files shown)


Partheo:
What? Can you elaborate?

ghost:
@Partheo basically "Blobs to win: 123/400", for example, will appear in the status panel

Slinky5890:
I don't think this is a great idea. There's been a few blobs who have won by virtue of people not knowing just how big of a threat they are- it also would show if a blob has split because the win goal increases, and stealth splitting at the back of the main mass is also a good strategy. It ALSO shows if there is multiple blobs

Partheo:
I really don't like this PR, agreed with Slinky here. It's meta-knowledge that doesn't make sense for a crewmember to have. You're making it too much of a game.

ghost:
@Partheo @Slinky5890 you COULD say it's Nanotrasen analyzing the blob's mass and updating the crew on it...

at least let the blob have it.

Partheo:
The blob does have it currently.

ghost:
@Partheo it does not, did you not see where i added that in the code?

Alek2ander:
Just send an extra centcom alert when the blob is at 75% goal or something, just in case the crew is being dumb

ghost:
for fucks sake why are you guys anti-qol

Jc1469:
Additionally, if people see a blob at 400/450 they probably just wont bother and will just space themselves.  I don't really see any point in this PR either way.

Ratvar:
Cause it's not quality of life but outright meta that fucks blobs

ghost:
It's not outright meta that fucks blobs. Everyone knows when a blob is there anyway. There is no such thing as a stealth blob. 

We could change it so that centcomm periodically sends updates as to blob size if you wanted. That would be about the same effect.

Dagdammit:
At least name it "Biohazard Saturation" or something.

ghost:
Complaining about immersion in a gamemode where everyone knows every last detail down to the minutae of how much damage the chem does per tick and how many tiles the blob has to cover to win anyway and being a roleplayer is actively discouraged.

Does it matter, Dagdammit?

Dagdammit:
Just seems weird to label it in a needlessly meta fashion.

Partheo:
With the counter, you can see that the blob could be expanding even though you can't see it or are in a different department. By knowing the strength of the blob at any given time regardless of where you are in station, or even if people are doing a good job fighting it (if you see dwindling numbers) then... Don't know what that is but meta-knowledge.

ghost:
@Partheo Why does it matteeeeeer.

Boo hoo, people know the size of the worst fucking antagonist in the game. Maybe it'll motivate them to get off their ass and fight. Stealth concerns are not a concern.

Partheo:
You're entirely right Adam. It doesn't matter. There's no reason for this PR to get merged, because the counter doesn't matter.

What does matter is that it's objectively meta knowledge, regardless of how you look at it. We both agree on this. If what we get out of this PR is so meaningless that it does not matter, then it's not worth merging it.

Cruix:
Best remove the shuttle ETA, station time, and gang domination timers from the status panel as well.

ghost:
thank god for cruix

code/game/gamemodes/blob/blob.dm
```diff
@@ -87,13 +89,20 @@ var/list/blobs_legit = list() //used for win-score calculations, contains only b
...
-
-		send_intercept(2) //if the blob has been alive this long, it's time to bomb it
+		addtimer(src, "announce_bleb", message_delay)
 
 	return ..()
+
+/datum/game_mode/blob/proc/announce_bleb()
+	send_intercept(1)
+	message_sent = 1
+	addtimer(src, "send_intercept", 24000, 2)
```
Cruix, line 99: The fourth parameter should be blank or FALSE, as it is now the send_intercept() call will not be sent the 2.

AshCorr:
Denied by council.

PR #2684, opened 2017-10-14
EMPs can damage energized fiber blobs

ghost:
what exactly was the logic behind commenting this out anyway, that the tesla would kill it? big whoop, just means they release singuloth instead

#### Changelog

:cl:  
 
tweak: EMPs can damage energized fiber blobs.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/blob_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/blob_reagents.dm
@@ -152,9 +152,9 @@
 /datum/reagent/blob/energized_fibers/tesla_reaction(obj/effect/blob/B, power)
 	return 0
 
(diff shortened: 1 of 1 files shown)


tsit9:
It's for when you split and end up with an EMP blob partner.

ShadowDeath6:
>>>>>>>>>>>making the ion rifle able to damage blobs
>>>>>>>>>>>>>>>>>>>making the retarded hos actually useful

yeah that's a no from me

also energized blob is super ass why would you make it even worse

ghost:
@ShadowDeath6 This was originally intended function of the blob as its unique weakness.

ShadowDeath6:
if you're nerfing it then give it something else unique to compensate, like having it shock people who are near it when it gets emped or something, because currently energized fibers are on the lower tier of offensive capabilities

Cruix:
They are weak even without the ion rifle eating them alive.

PR #2692, opened 2017-10-15
Improves logging and log viewing

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Because Alex humblebragged about his log diving script and I can't fathom the thought of being worse than someone else. Changes:
- Admin, attack, and other logs are now all in one file.
- Logs, instead of being a clusterfuck of no timestamps, no prefixing, and only one log type having the round number, now all follow the same format:
|\<Round number>|\<Timestamp>|\<Log type>|\<Content>
- New log viewer! Screenshot:
image

On the left you can check what log types you want to see. 
On the bottom you can type in comma-separated search terms, which must or must not be included in the final log. You can select whether all or just one of the search terms need to appear in each log entry.
Check "Use old log format" to search through logs from before this PR is inevitably merged or if you are so amazed you want to use it RIGHT NOW. Don't check it if browsing new logs or it will crash. Exception handling is for the weak.
Check "Relative Search" to only search through what you see on the screen instead of combing through the entire log again.
Finally press "Copy" in the upper right to copy what you see to the clipboard so you can quickly paste funny logs into Discord for everyone to laugh at.

Sample usage: https://i.imgur.com/1zcdPXD.mp4


Amelia0010:
I'm actually in love with you

Jc1469:
For the sake of years of logs not being in this format, and not wanting to have to bother to use a program to sort through logs and miss context out because that's all this log diving script seems to do, I'm against this.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
If we didn't accept new things because "that's how it always was" we'd still be in the stone age. What do you not like about the proposed format?

If you don't want to use the program, of course you can manually browse logs. I just think it would be easier and faster with it.

Alexandria404:
The other log-viewing tools won´t work anymore which bothers me(mine).
While that is less of an issue as long as your tool works, we can´t use other tools anymore in the future as well, which I´m not comfortable with.
You could maybe make your tool be able to combine the logfiles and export it in one file, so everybody can have their cake, while nothing is lost?

Cruix:
I am not going to add an exe file to the tools folder if the source code is not included. Ideally I would not add an exe at all, unless I had compiled it myself, and even then I would be hesitant.

PR #2693, opened 2017-10-15
Removes electrocution from TK

### LINKED ISSUE #2683: Can get shocked when use TK on cables from distance. Suppose the same when hacking things, APC, such.
Title says it. TK=/=insulated ranged gloves.

Amelia0010:
fixes #2683

You now won't be electrocuted if you're not next to the machine shocking you. They'll still spark, but no longer will you be electrocuted

#### Changelog

:cl:  
tweak: Brainwaves no longer conduct electricity
/:cl:

### DIFF
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -124,7 +124,7 @@
 	cameranet.updateVisibility(src, 0)
 
 /obj/machinery/camera/proc/shock(mob/living/user)
-	if(src.Adjacent(user))
+	if(!src.Adjacent(user))
 		return 0
 	if(!istype(user))
 		return
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -124,6 +124,8 @@
 	cameranet.updateVisibility(src, 0)
 
 /obj/machinery/camera/proc/shock(mob/living/user)
+	if(src.Adjacent(user))
+		return 0
 	if(!istype(user))
 		return
 	user.electrocute_act(10, src)
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -225,7 +225,7 @@ var/list/airlock_overlays = list()
 	var/datum/effect_system/spark_spread/s = new /datum/effect_system/spark_spread
 	s.set_up(5, 1, src)
 	s.start() //sparks always.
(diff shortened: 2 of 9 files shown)


code/game/machinery/camera/camera.dm
```diff
@@ -124,6 +124,8 @@
 	cameranet.updateVisibility(src, 0)
 
 /obj/machinery/camera/proc/shock(mob/living/user)
+	if(src.Adjacent(user))
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: Is this not supposed to be if(!src.Adjacent(user))

code/game/machinery/doors/airlock.dm
```diff
@@ -225,6 +225,8 @@ var/list/airlock_overlays = list()
 	var/datum/effect_system/spark_spread/s = new /datum/effect_system/spark_spread
 	s.set_up(5, 1, src)
 	s.start() //sparks always.
+	if(src.Adjacent(user))
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: this too

code/modules/atmospherics/machinery/airalarm.dm
```diff
@@ -358,6 +358,8 @@
 	var/datum/effect_system/spark_spread/s = new /datum/effect_system/spark_spread
 	s.set_up(5, 1, src)
 	s.start() //sparks always.
+	if(src.Adjacent(user))
```
1fbff5f83b23d39d38b1dfcb4cac8d9b: and this

code/game/machinery/camera/camera.dm
Amelia0010: i'm dumb

ghost:
ot a fix, intentional downside

Amelia0010:
🤔 Then consider this a tweak/rebalance/whatever

AshCorr (approved):

code/game/machinery/suit_storage_unit.dm
```diff
@@ -270,11 +270,14 @@
 
 /obj/machinery/suit_storage_unit/proc/shock(mob/user, prb)
 	if(!prob(prb))
```
Cruix, line 272: The only things that call this proc do not send the second parameter, so this will always be if(!prob(null)), which will always be true, so the shock will never happen.

PR #2702, opened 2017-10-17
lowers cost of blob relocation to 50

ShadowDeath6:
lets you do more strategical shit

it still costs a lot, just less so

:cl: ShadowDeath6
tweak: Blob Relocation now costs 50 BP, down from 80.  
/:cl:


Nickvr628:
Oh my god

Ratvar:
Yes, boi. Thought, you can move your core and plop node near, then move back and have two nodes near this way with ease, stacking them

PR #2716, opened 2017-10-22
Nich's incredibly non-spitefull admin ballet-dance

Amelia0010:

Moderators can now use Pick_random_player.


Slinky5890:
But does pick random player actually function properly now? Previously it only worked when picking from all players regardless of any criteria it offers you

Amelia0010:
doesn't work if you don't check from everyone, i'll fix it in another PR

PR #2718, opened 2017-10-23
buffs the sleepypen

ghost:
given that the thing is no longer refillable, this should make it actually usable.

let's cover exactly what the issue is.

it's certainly fine at knocking people out. however, morphine removes slowdown from people-as a result, they always run at full speed. When someone feels the tiny prick, they're likely to ask what the hell, and when they can't speak because of the mute toxin, they run. Now, the tirizine was INTENDED to slow them down so they couldn't run as fast before the morphine knocked them out...but of course, morphine removes slowdown.

As a result, without changing the chems in it, it was effectively useless as an item.

Previously, you could change the sleepypen's chem content. Unfortunately, as it turns out, a 45u pocket hypospray you could fill with a deathmix of your choice that wasn't very visible inhand and was completely unsuspicious turned out to be horrifically overpowered, so after dumping the initial chem mix people would just fill it with black powder and phlog, jab someone with it and watch them fucking gib, or something equally silly. For obvious reasons this got changed, but left the sleepypen an effectively completely useless item.

With this change, that should fix it. I've changed the morphine out for a cocktail of other particularly effective knockout drugs, but nothing that works immediately. The new drugs are also distributed in low quantities, so if someone should happen to chug charcoal or something, they would certainly not be knocked out in time as compared to the previous sleepypen, which was a virtually guaranteed but effectively useless knockout.

To replace the morphine the new chems are chloral, haloperidol and sodium thiopental, because sodium thiopental's a syndicate knockout chem and why wouldn't they stock their pen with it?

Considering this is a 4tc investment to kill just one person and can't be refilled ever, this is a fair trade.

#### Changelog

:cl:  
tweak:The sleepypen should actually not suck and be effective at knocking people out now to compensate for the inability to refill it.
/:cl:

### DIFF
--- a/code/modules/paperwork/pen.dm
+++ b/code/modules/paperwork/pen.dm
@@ -98,7 +98,9 @@
 
 /obj/item/weapon/pen/sleepy/New()
 	create_reagents(45)
-	reagents.add_reagent("morphine", 20)
+	reagents.add_reagent("chloralhydrate", 8)
+	reagents.add_reagent("haloperidol", 4)
+	reagents.add_reagent("sodium_thiopental", 8)
 	reagents.add_reagent("mutetoxin", 15)
 	reagents.add_reagent("tirizene", 10)
 	..()


AshCorr (approved):
Tested, seems decent for 4tc

PR #2726, opened 2017-10-24
Adds cargo shutters

Partheo:
image
image

Now you can stop fucks from climbing onto the table. Not access-restricted so people that flush themselves down disposals can still get out.

:cl: Partheo
rscadd: The Cargo Mail office now has privacy shutters.
/:cl:


Alek2ander:
:-1: Why would you want to stop assistants digging through the trash? Not a high sec area

Partheo:
Neither is the chef's kitchen.

It's an area they don't have access to. Plain and simple. If they really want to get in they can go flush themselves down disposals.

ghost:
cargo powercreep

ToGWtF:
-1 

now AI can stop EVERYONE who tries to disposal themselves to escape.

Partheo:
@ToGWtF You realize they can just press the button and leave, right?

PR #2727, opened 2017-10-24
Adds a way to deal with looping immovable rods

1fbff5f83b23d39d38b1dfcb4cac8d9b:
"Oh no, the immovable rod has started looping! Call the shuttle!"
How many times has this happened to you? Pissed off because random chance screwed up your traitor round? Fear no more! Just ask your nearest RD to grab a plant box then grab the rod and suplex it into the poor bush.
*Slipping and penetration insurance sold separately.
:cl:  
rscadd: Research Directors can now attempt to slam looping immovable rods into plant pots. Grab the nearest pot and try to catch the rod!
/:cl:

### DIFF
--- a/code/modules/events/immovable_rod.dm
+++ b/code/modules/events/immovable_rod.dm
@@ -51,6 +51,29 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 /obj/effect/immovablerod/ex_act(test)
 	return 0
 
+/obj/effect/immovablerod/proc/check_suplex(mob/living/carbon/human/H)
+	if(!istype(H) || !H.mind || H.mind.assigned_role != "Research Director" || !H.in_throw_mode)
+		return FALSE
+	var/obj/structure/flora/kirbyplants/K = locate() in range(1, H)
+	if(!K)
+		return FALSE
+
+	var/turf/open/floor/T = get_turf(K)
+	H.forceMove(T)
+
+	H.visible_message("<span class='userdanger'>[H] suplexes \the [src] into [K]!</span>","<span class='userdanger'>You suplex \the [src] into [K]!</span>")
+	var/obj/structure/festivus/P = new(T)
+	P.desc = "During this year's Feats of Strength the Research Director was able to suplex this passing immovable rod into a planter."
+	if(istype(T))
+		T.break_tile()
+	playsound(T, "explosion", 100)
+	for(var/mob/bystander in urange(10, src))
+		shake_camera(bystander, 7, 2)
+
+	qdel(src)
+	qdel(K)
+	return TRUE
+
 /obj/effect/immovablerod/Bump(atom/clong)
 	if(prob(10))
 		playsound(src, 'sound/effects/bang.ogg', 50, 1)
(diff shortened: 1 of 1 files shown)


Partheo:
What do you actually have to do to get rid of the rod?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
1. Be RD
2. Grab a plant box (there is one in front of engineering lobby, below the vending machines)
3. Click the rod
4. Become the station's hero

Partheo:
I don't see why only the RD should be able to do this. Also, clicking the rod sounds nearly impossible- i think a better solution would be to stand in front of the rod and turn throw mode on- that way you "catch" the rod and dunk it.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
It's a reference to the festivus pole's description - "During last year's Feats of Strength the Research Director was able to suplex this passing immobile rod into a planter."

As to catching the rod, eeeeeeeeehhhhhhhh, might be a bit too easy

Partheo:
Oh, in that case I'm fine with only the RD doing it. Good meme.

Given how rare I think this occurence is going to be(only those who see this PR will know) I think catch mode might be better. However, I also concede that it will be hilarious to see a swarm of ghosts watching an RD about to suplex the rod and instead getting gibbed.

You should make it play the flip emote on a succesful suplex too.

ShadowDeath6:
good pr

don't make it flip, have it do custom emote where he flexes his muscles or something

Partheo:
The thing with flip is that it fits a suplex animation. There already is a custom message.

ShadowDeath6:
flipping is overdone and bad though

code/modules/events/immovable_rod.dm
```diff
@@ -72,3 +72,30 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
...
 			clong.ex_act(2)
 	return
+
+/obj/effect/immovablerod/attack_hand(mob/M)
+	if(!ishuman(M))
+		return
+	if(!M.mind || M.mind.assigned_role != "Research Director")
+		return
+
+	var/obj/structure/flora/kirbyplants/K = locate() in orange(2, M)
```
X-TheDark:
```
Format:
    orange(Dist,Center=usr)

Returns:
    A list of objects within Dist tiles of Center, >>>excluding<<< Center.
```
Are you sure you want it to **exclude** the tile RD is standing on (meaning it also includes the pot being on your persona).

1fbff5f83b23d39d38b1dfcb4cac8d9b: good catch, thank

Dagdammit:
Okay but should't we just patch the stupid looping immovable rod bug in the first place

ShadowDeath6:
it's not a bug

Kmc2000:
good PR

CrazyDrakkon:
Add a flash-like blinding light effect (just visual) immediately after success and flash sound after explosion sound.

Ratvar:
allow people to throw potted plant to relise rod back into that direction

Amelia0010 (approved):
Tested this and it works 👍

code/modules/events/immovable_rod.dm
```diff
@@ -67,8 +67,34 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 	else if (istype(clong, /mob))
 		if(istype(clong, /mob/living/carbon/human))
 			var/mob/living/carbon/human/H = clong
+			do
```
AshCorr: _For_.
 ### Gods.
#  **Sake**.

Please. Jus this do... while (false) code into a separate proc so that you can do this returning shit.

1fbff5f83b23d39d38b1dfcb4cac8d9b: no fun allowed

code/modules/events/immovable_rod.dm
```diff
@@ -51,6 +51,29 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
...
+	playsound(T, "explosion", 100)
+	for(var/mob/bystander in urange(10, src))
+		shake_camera(bystander, 7, 2)
+
+	qdel(src)
+	qdel(K)
+	return TRUE
+
 /obj/effect/immovablerod/Bump(atom/clong)
 	if(prob(10))
```
AshCorr: You will probably have to use the `deleted(src)` proc to check if the rod has been queued for deletion as the Garbage collection system doesnt usually queue objects for deletion straight away.
```
if(qdeleted(src))
    return
```

PR #2729, opened 2017-10-25
New updated sprites for vents and scrubbers

Partheo:
New sprites for vents and scrubbers. Eh?

image

Turning a regular vent on and off. Notice wind-up and wind-down animation.
smoothboye

succ vents have received the same treatment.

Turning a scrubber on, off, and on siphon mode. Note how the sprite changes when the "expanded" range is selected.
smoothgal

Overall, it's a big change. The fans are very distracting at first. It took me about an hour to get used to. Just recall what the old sprite is and you'll see it really isn't a problem.

Had to do a whole bunch of dumb logic in the vent's update_icon() to make the fan accelerate at the right times.
I moved the vent and scrubber sprites from unary_devices.dmi to their own new respective files in the same folder.
Also, Flick() is a thing. Who knew?

:cl: Partheo 
imageadd: Vents and scrubbers have received new sprites
/:cl:


Partheo:
Expect salt in OOC if this gets merged. "Fans are so distracting!"

Give it a few hours and you'll drown it out just like the current vent animation.

Alek2ander:
- Vent animation should be slower (x0.5?). It *is* pretty distracting right now, due to speed. Possibly use the fast animation for refill/flood modes.
- Scrubbers look kinda like holopads. Could confuse someone eventually.
- Scrubbers normal mode (scrubbing) animation shouldn't look like old siphon. Maybe use static blue lighting, or close/open the slats.
- Siphon has been green for ages, does it need to be red? (maybe reserve red for widenet siphon during panic siphon mode?)

Partheo:
1. I maintain that it stops being distracting after a few hours of playing around the vents. Shouldn't be too much of an issue IMO, just takes some getting used to. I might make the blades darker so it looks like there's less movement in the vent.

2. Fair point. I'm going to leave them as is for now and update holopads in the near future.

3. I'll slow down the regular animation. However, given how easy to see panic siphon is, I think it's just going to be a matter of getting used to it. Again, it's different and all take some getting used to.

4. Green didn't make much sense to me. It's also a similar color to blue, which is important for subtle effects like the one used in the scrubber animation. I also wanted to make siphoning more scary, so you really shit your pants when the AI bolts down your doors and turns it on.

ghost:
you're missing the point here, animations for background objects should not be this distracting, even if it "drowns out" after a few hours, it's still a shitty thing to have. Unless an object has some really high importance, it shouldn't be the most active thing on your screen.

Ratvar:
Scrubbers/vents filling in plasma or sucking air are high importance, no

ghost:
in that case we might as well put huge flashing lights on assistants because they're dangerous. Unless they're actually doing something dangerous / important at the time they have no need to be this distracting

Partheo:
Fair point. I'm going to leave scrubbers as is since they're fairly muted and make the fans more subtle.

ghost:
that's fine, the scrubbers aren't distracting.

Partheo:
Note to all those taking this PR for other codebases and not linking it for me or others- this is still WIP.

Amelia0010:
image

Zostroll:
The scrubbers look pretty terrible I like the vents though.

Ratvar:
old scrubbers looked pretty terrible, new looks much bettah even if they look like sinkholes

PR #2744, opened 2017-10-28
Fixes a few bugs

### LINKED ISSUE #2700: Fail to use construct ability still cooldowns it
Attempts to use soulshard ability in wall (aka cult wall, for example, where arts can use freely) uses ability but won't place shard itself.

### LINKED ISSUE #2732: OOC: Sauwit: I hid in a closet to avoid the ash storm
Fixed radiation lockers? Fix ashstorm locker

### LINKED ISSUE #2742: pAI DNA Request Bug
Often during rounds where I play as an observer I take up pAI when requested. When activated, I always request a DNA sample from the person holding me, however when given the DNA sample it tells me the DNA does not match my master's DNA, even though when listed in my directives tab that I am not synced with any master. This issue only appears to happen when I request a DNA sample and not when given a DNA sample when downloaded. I've been on Yogstation since around early or mid-August 2017 and this has been an issue that can happened on numerous occasions.

### LINKED ISSUE #2743: Chem Dispenser Issue
So, while trying to mix up to 100 units of chemicals, I got a strange issue with the chem dispenser.

I've have only had this issue today and have been able to replicate this issue in a private session. 

For example:
Salicyclic acid needs 1 part Sodium,1 part Phenol, 1 part Carbon, 1 part Oxygen, 1 part Sulfuric Acid. When added in the last 20 units of Sulfuric acid the chem dispenser doesn't add up the last 5 units even though though it should.

I believe this issue is caused by the chemical dispenser believing there is more chemicals in the dispenser than there really is and is possibly also related to the amount of chemicals trying to be mixed at once but also it might be an issue with previous chemical mixtures (Such as phenol having to be added before hand) I haven't has issues with mixtures such as 100 units of Kelotane or Styptic Powder (which use only the base chemicals in the dispenser). However I am not sure if this is the case also because I have no problem with 100 units of Silver Sulfadiazine(requires ammonia to be mixed) either.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Fixes #2743 
Fixes #2742 
Fixes #2700 
Fixes #2732 

:cl: McDonald072
bugfix: You can no longer hide in lockers to evade ash/radstorms.
bugfix: Fixed reagents sometimes reaching precision levels that shouldn't be possible.
bugfix: Constructs no longer fail to create shells and shards while inside walls.
:/cl:

### DIFF
--- a/code/__HELPERS/maths.dm
+++ b/code/__HELPERS/maths.dm
@@ -165,6 +165,3 @@ var/gaussian_next
 		gaussian_next = R2 * working
 	return (mean + stddev * R1)
 #undef ACCURACY
-
-/proc/floor(num)
-	return num - num%1
--- a/code/__HELPERS/maths.dm
+++ b/code/__HELPERS/maths.dm
@@ -168,6 +168,3 @@ var/gaussian_next
 
 /proc/floor(num)
 	return num - num%1
-
-/proc/round_precision(number, prec)
-	return floor(number * 10**prec + 0.5) / 10**prec
--- a/code/__HELPERS/maths.dm
+++ b/code/__HELPERS/maths.dm
@@ -165,3 +165,9 @@ var/gaussian_next
 		gaussian_next = R2 * working
 	return (mean + stddev * R1)
 #undef ACCURACY
+
+/proc/floor(num)
+	return num - num%1
+
+/proc/round_precision(number, prec)
+	return floor(number * 10**prec + 0.5) / 10**prec
--- a/code/controllers/subsystem/weather.dm
+++ b/code/controllers/subsystem/weather.dm
@@ -19,7 +19,8 @@ var/datum/subsystem/weather/SSweather
 		if(W.aesthetic)
 			continue
 		for(var/mob/living/L in mob_list)
-			W.impact_if_possible(L)
+			if(W.can_impact(L))
+				W.impact(L)
 	for(var/Z in eligible_zlevels)
 		var/list/possible_weather_for_this_z = list()
 		for(var/V in existing_weather)
--- a/code/controllers/subsystem/weather.dm
+++ b/code/controllers/subsystem/weather.dm
@@ -19,9 +19,7 @@ var/datum/subsystem/weather/SSweather
 		if(W.aesthetic)
 			continue
 		for(var/mob/living/L in mob_list)
-			var/area/A = get_area(L)
-			if((L.z == W.target_z) && !(W.immunity_type in L.weather_immunities) && (A in W.impacted_areas))
-				W.impact(L)
+			W.impact_if_possible(L)
 	for(var/Z in eligible_zlevels)
 		var/list/possible_weather_for_this_z = list()
 		for(var/V in existing_weather)
--- a/code/modules/reagents/chemistry/holder.dm
+++ b/code/modules/reagents/chemistry/holder.dm
@@ -144,7 +144,7 @@ var/const/INJECT = 5 //injection
 	var/trans_data = null
 	for(var/reagent in reagent_list)
 		var/datum/reagent/T = reagent
-		var/transfer_amount = round_precision(T.volume * part, 3)
+		var/transfer_amount = round(T.volume * part, 0.001)
 		if(preserve_data)
 			trans_data = copy_data(T)
 		R.add_reagent(T.id, transfer_amount * multiplier, trans_data, chem_temp, no_react = 1) //we only handle reaction after every reagent has been transfered.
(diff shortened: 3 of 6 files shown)


code/modules/reagents/chemistry/holder.dm
```diff
@@ -144,7 +144,7 @@ var/const/INJECT = 5 //injection
 	var/trans_data = null
 	for(var/reagent in reagent_list)
 		var/datum/reagent/T = reagent
-		var/transfer_amount = T.volume * part
+		var/transfer_amount = round_precision(T.volume * part, 3)
```
X-TheDark: Try to replace this with `round(T.volume * part, 0.001)` (BYOND native proc) and see if that fixes the issue.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -190,7 +190,7 @@ var/const/INJECT = 5 //injection
 	var/datum/reagents/R = target.reagents
 	if(src.get_reagent_amount(reagent)<amount)
 		amount = src.get_reagent_amount(reagent)
-	amount = min(amount, R.maximum_volume-R.total_volume)
+	amount = round_precision(min(amount, R.maximum_volume-R.total_volume), 3)
```
X-TheDark: Try to replace this with `round(min(amount, R.maximum_volume-R.total_volume), 0.001)` (BYOND native proc) and see if that fixes the issue.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -527,6 +527,7 @@ var/const/INJECT = 5 //injection
 		var/datum/reagent/R = A
 		if (R.id == reagent)
 			R.volume += amount
+			R.volume = round_precision(R.volume, 3)
```
X-TheDark: `round(R.volume, 0.001)`

code/modules/reagents/chemistry/holder.dm
```diff
@@ -575,6 +576,7 @@ var/const/INJECT = 5 //injection
 		var/datum/reagent/R = A
 		if (R.id == reagent)
 			R.volume -= amount
+			R.volume = round_precision(R.volume, 3)
```
X-TheDark: `round(R.volume, 0.001)`

code/controllers/subsystem/weather.dm
```diff
@@ -19,9 +19,7 @@ var/datum/subsystem/weather/SSweather
 		if(W.aesthetic)
 			continue
 		for(var/mob/living/L in mob_list)
-			var/area/A = get_area(L)
-			if((L.z == W.target_z) && !(W.immunity_type in L.weather_immunities) && (A in W.impacted_areas))
-				W.impact(L)
+			W.impact_if_possible(L)
```
X-TheDark: Instead of defining new procs, would you mind trying, instead, to replace the line 22 with this (and keeping the rest the same):
`var/area/A = get_area(get_turf(L))`

X-TheDark (changes requested):
Could you test the changes I proposed and remove the new procs you made if the tests confirm you don't need those new procs?

code/controllers/subsystem/weather.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: Gonna move impact_if_possibe() here instead, the weather subsystem instead of using can_impact() does its own check

X-TheDark: No, you don't understand. I don't think you need ANY new proc for this. Just add the bit I suggested and keep the lines 23-24 (the conditions you changed) as they were (revert).

code/modules/reagents/chemistry/holder.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: Totally failed to understand how round works, go me... Gonna leave floor() in though, might be useful in the future

code/controllers/subsystem/weather.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: I'm not adding a new proc, can_impact() already exists in `/datum/weather` and already contains the fix for the bug. The subsystem just neglected to use it.

code/__HELPERS/maths.dm
```diff
@@ -165,3 +165,6 @@ var/gaussian_next
 		gaussian_next = R2 * working
 	return (mean + stddev * R1)
 #undef ACCURACY
+
+/proc/floor(num)
+	return num - num%1
```
X-TheDark: Remove this, if the tests show `round` can deal with the issue of reagent amount accuracy.

X-TheDark (approved):

PR #2757, opened 2017-11-06
Allows for the TEG to be wrench/unwrenched and adds it to cargo

Amelia0010:

First off: Adds the Thermo-Electric Generator to cargo. It costs 10.000 credits, just like a supermatter shard.

Second: Allows the Thermo-Electric Generator to be moved around, by wrenching/unwrenching it.

#### Changelog

:cl:  
rscadd: Nanotrasen has begun selling Thermo-Electrig Generators through their supply vendors! Buy yours today!
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/binary_devices/circulator.dm
+++ b/code/modules/atmospherics/machinery/components/binary_devices/circulator.dm
@@ -8,15 +8,18 @@
 	icon_state = "circ1-off"
 
 	var/side = CIRC_LEFT
+	var/circ = null
 	var/status = 0
 
 	var/last_pressure_delta = 0
 
-	anchored = 1
+	anchored = 0
 	density = 1
 
 	var/global/const/CIRC_LEFT = 1
 	var/global/const/CIRC_RIGHT = 2
+
+	var/obj/machinery/power/generator/generator = null
 
 
 /obj/machinery/atmospherics/components/binary/circulator/proc/return_transfer_air()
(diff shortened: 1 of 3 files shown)


ShadowDeath6:
does this actually work

Partheo:
Let the autism commence

AshCorr (approved):
We can rebalance it later.

It shouldnt be able to generate power just by using a heater and a cooler.

Amelia0010:
image

I got 2MW out of a heater and cooler.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
that is 2 million watts

Amelia0010:
I can't do numbers

Alexandria404:
Sounds like way too much power.

Amelia0010:
Well, the singularity and the tesla both make more than enough power to power the whole station several times over, with no cost, whereas this would power the station several times over, with some cost.

Not to mention that having a lot of power doesn't really matter. Either you have enough, or not. Excess really doesn't matter. You don't gain anything from having extra power, other than the fact that you have extra power.

Partheo:
Thinking that this would break balance is ridiculous. SS13 has next to no balance and lives off of its features and complexity. Excess power generation really doesn't matter.

grubstank:
It doesn't matter in the current meta, but it is a good idea to maintain balance for future features - for instance, if we were to one day port over the power export laser from goon

If you actually want to add complexity to the game, at least the core features need some semblance of balance - for instance there really is no meaningful strategic difference between the tesla and singulo

PR #2758, opened 2017-11-07
improves horrorform

ShadowDeath6:
the right way to code this was in the back of my mind for like 17 weeks and I just remembered how to do it

:cl: ShadowDeath6
rscadd: Abominations can now rip apart human corpses!
tweak: Reversion from Horror Form now deals less genetic damage.
tweak: Lowered the amount of time abominations knock humans down for on hit.
tweak: Abominations are slightly weaker against lasers and bullets now.
tweak: Horror Form requires 30 chemicals to enter, up from 20.
bugfix: Fixed abomination devour selecting multiple targets
bugfix: Fixed abominations being able to devour monkeys and husks
bugfix: Abominations now correctly annihilate monkeys, rather than hitting them like normal humans. They can also rip them apart once they're dead!
/:cl:


ShadowDeath6:
good start

Ratvar:
Please, let them devour monkey/husk changelings. Reverting just to find glassshard in time/get enough points for armblade is, uh, meh.

ShadowDeath6:
I'll see what I can do, if a ling turns into a monkey to avoid you devouring them you can just kill them anyway, it takes longer to regen than it does to get an armblade

ShadowDeath6:
I made it so abominations can rip apart monkeys, lings are dead if they get husked anyway

Ratvar:
Nope, due to unfixed bug lings can revive when husked, Shaadooow. I experienced it, i even wrote issue about it,

ShadowDeath6:
fixing bugs isn't in the scope of this pr, horrorforms have no reason to devour unsuitable stuff

Ratvar:
Why? You can clone (updated cloner) and get brain out of husks. Eating it in horrorrofm removes that probability on a go.
Horrorform is a way to dispose of bodies, and husking leaves bodies. Here is the reason.
Allow horrorforms to tear husks apart the same way as monkeys?

ShadowDeath6:
alright

ShadowDeath6:
I think this is r4m

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -5,10 +5,18 @@
 		..(user, 1)
 		playsound(loc, user.dna.species.attack_sound, 25, 1, -1)
 		if(user.dna.species.id == "abomination")//snowflake abomination hulk damage
```
AshCorr: Can be replaced with `isabomination(user)`

AshCorr (approved):
Tested, all works fine.

The 1 second knockdown looks quite silly though.

ShadowDeath6:
it basically serves as a disarm, people with armor won't be stunned at all and people without can be stunlocked because getting into a horrorform's melee range isn't a good idea

PR #2775, opened 2017-11-14
Adds server logs for roundstart job and job changes.

ghost:
Intent of your Pull Request

Make it easier to find a person's job from a pure log standpoint.

Changelog

:cl:
rscadd: Adds server logs for roundstart job and job changes.
/:cl:

### DIFF
--- a/code/game/machinery/computer/card.dm
+++ b/code/game/machinery/computer/card.dm
@@ -416,6 +416,7 @@ var/time_last_changed_position = 0
 
 					modify.access = ( istype(src,/obj/machinery/computer/card/centcom) ? get_centcom_access(t1) : jobdatum.get_access() )
 				if (modify)
+					log_game("[modify.registered_name]'s ID had its job changed to [t1] from [modify.assignment]")
 					modify.assignment = t1
 		if ("demote")
 			if(modify.assignment in head_subordinates || modify.assignment == "Assistant")
--- a/code/modules/jobs/job_types/job.dm
+++ b/code/modules/jobs/job_types/job.dm
@@ -233,6 +233,7 @@
 		PDA.owner = H.real_name
 		PDA.ownjob = H.job
 		PDA.update_label()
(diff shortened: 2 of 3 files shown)


ghost:
I actually have some issues with this one that I would like some help with.

1. For some reason joining as mime on roundstart doesn't give the correct name but for clowns it works.
2. I can't find a good place for adding this for roundstart AI/Borgs but I haven't looked too hard yet
3. Also any feedback on the job change log would be cool as I feel it could be expanded upon.  Thinking about adding an investigate for ID changes in game as well for admins.  Not sure where that is though so I'll have to look into that.

Either way any feedback would be great.

code/modules/jobs/job_types/job.dm
```diff
@@ -233,7 +233,8 @@
 		PDA.owner = H.real_name
 		PDA.ownjob = H.job
 		PDA.update_label()
-	log_game("[H.real_name]/[H.ckey] joined the round as [H.job].")
+	if(H.job != "Mime" && H.job != "Clown")
```
AsV9: Hahahahahaha what a hack

ghost: Interestingly enough the original version would give mimes random mime names and clowns would have the regular name instead of the clown name.  Dunno why but interesting either way.

code/game/machinery/computer/card.dm
```diff
@@ -416,6 +416,7 @@ var/time_last_changed_position = 0
 
 					modify.access = ( istype(src,/obj/machinery/computer/card/centcom) ? get_centcom_access(t1) : jobdatum.get_access() )
 				if (modify)
+					log_game("[modify.registered_name]'s ID had its job changed to [t1] from [modify.assignment]")
```
AshCorr, line 419: Perhaps this should also say who changed the ID?

ghost: For the purpose it's not really relevant information but I'll see if I can't add it as more info is never bad.

PR #2777, opened 2017-11-15
Ports Object-decals from TG

Amelia0010:
Adds decals as an object, which allows mappers to place it on any open turf. Currently a few tiles come prepackaged with the decals.


AshCorr:
If you're going to port this then you could atleast do the whole job and port the map changes too so that this is actually doing something.

You can perhaps use this script to update our maps with the new decals.
https://github.com/tgstation/tgstation/pull/23385/files#diff-ad79adf84d8d00b267739ee928e9c2c0

Zostroll:
Nich are you going to port the map changes?

Time-Green:
https://github.com/tgstation/tgstation/pull/34492/files
Port this ples

PR #2785, opened 2017-11-16
Fixes a couple bugs pointed out by Kojuh

1fbff5f83b23d39d38b1dfcb4cac8d9b:


code/modules/surgery/prosthetic_replacement.dm
```diff
@@ -35,7 +35,8 @@
 	if(target_zone == tool_body_zone) //so we can't replace a leg with an arm.
 		user.visible_message("[user] begins to replace [target]'s [parse_zone(target_zone)].", "<span class ='notice'>You begin to replace [target]'s [parse_zone(target_zone)]...</span>")
 	else
-		to_chat(user, "<span class='warning'>[tool] isn't the right type for [parse_zone(target_zone)].</span>")
+		to_chat(user, "<span class='warning'>[tool] isn't the right thing to replace [target]'s [parse_zone(target_zone)] with.</span>")
```
ShadowDeath6: can you just say "part" instead of "thing"

AshCorr: `to_chat(user, "<span class='warning'>[tool] cant be used to replace [target]'s [parse_zone(target_zone)].</span>")`

AshCorr (changes requested):

code/modules/surgery/prosthetic_replacement.dm
1fbff5f83b23d39d38b1dfcb4cac8d9b: `to_chat(user, "<span class='warning'>Well hello there Mr player. Unfortunately, it has come to my attention that you are attempting to use [tool] to replace [parse_zone(target_zone)] of the poor [target]. Don't you think it's quite a strange idea? You'd be able to use any body part or any prosthetic to replace people's limbs. This would allow players to attach 10 arms to each other, so they can wield eswords forever and become absolutely unkillable. Doesn't sound nearly as fun, does it? Now, go ahead and grab \a [parse_zone(target_zone)] to replace [target]'s missing limb properly, ya silly goose.</span>")`

AshCorr (approved):

PR #2798, opened 2017-11-20
Lets moderators offer to ghosts

Amelia0010:
It's an incredibly usefull button, and it's hard to abuse without being discovered as it not only is logged, but it tells the player they've been offered to ghosts. 

Oh, and oak approved of this earlier when I asked him whether I could make it.

image


ActualRory:
What a beautiful window that screenshot is into admin-chat.
Nonetheless, a very good change if it works as it should.

PR #2825, opened 2017-12-06
Fixes donors not having access to extra slots.

AshCorr:
Previously the preferences datum would check to see if a user had donor before the client could assign themselves a donor status.

:cl:
bugfix: Donors now have access to 6 character slots!
/:cl:


AsV9:
Looks good to me.

PR #2829, opened 2017-12-09
Makes forcesay check for lowercase

Amelia0010:
Previously it'd only trigger if you manually typed in "Say "". Which meant that anyone using the "toggle chat" preference weren't affected by it, since it autofills with a lowercase "say "".


AsV9:
wouldn't it be easier to just make all the characters lowercase _then_ check if you have a match so I can't do SAY "..." to get around the forcesay

Amelia0010:
I mean... I suppose it would be...

PR #2831, opened 2017-12-09
Take 2 of #2825 

AshCorr:
Take 2 of #2825


Amelia0010:
update_character_slots isn't called anywhere?

Amelia0010:
ah

PR #2846, opened 2017-12-14
Fixes bartender ID colours

Amelia0010:
Time accidentally put a comma in there, which means it's grey and black, rather than green and black


drderp3635:
Wouldn't grey and black be better for a bartender though?

Time-Green:
But that'd fuck with the department system

matskuman5:
also, it should be "civilian", not "civillian"

PR #2850, opened 2017-12-16
Adds a flw message to AOs in asay

Amelia0010:
I don't know why they don't have one as is.


AsV9:
Probably because people forgot, Nich.

PR #2856, opened 2017-12-17
Makes flashes burn out after use, not before

### LINKED ISSUE #2855: Head revs that are created midround spawn with a broken converter
Had this happen to me before, had this happen to someone else just now. When made into a head revolutionary in the middle of the round, the suspicious device generated for you doesn't work. Can be somewhat worked around by making a new one with with the electronic flashbulb hijacker.

Should be easy to recreate, I think. Just acquire head rev mid-round and try to flash someone with the suspicious device spawned in for you.

Amelia0010:
Closes #2855.

:cl:  
fix: flashes will now burn out after use, instead of before
/:cl:


Zostroll:
I kinda think not know if a flash is going to work is one of the disadvantages that influence how they are used and this would be a major buff to flashes.

PR #2857, opened 2017-12-17
Only non-braindead crew get antag weight increase at round end

AlpacaGoesCrazy:
This PR is to force players to value their life and to actually play in a round and not go braindead.
Antag weight (probability to become antag in next round) increases now at round end. Only alive, not braindead people who didnt roll antag are eligible for antag weight increase.

**EDIT: as people dont appreciate punishing crew who died now this only punish crew who were braindead when round ended**

:cl:    
tweak: Increased probability to become an antagonist is now awarded for non-antag players who were active during the round.
/:cl:

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -474,7 +474,7 @@
 		for(var/v in not_chosen)
 			var/datum/mind/candidate = v
 
-			if(candidate.current && candidate.current.client)
+			if(candidate.current && candidate.current.client && candidate.current.stat != DEAD) //dead and braindead people wont get their weight increased
 				var/client/C = candidate.current.client
 
 				if(C && C.last_cached_weight) //if it's not null, we've updated it during the picking stage
@@ -484,6 +484,7 @@
 					var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[SQL_ckey]'")
 					query.Execute()
 					C.last_cached_weight = weight
+					to_chat(candidate.current, "<font color='green'><b>Your chance to become an antagonist in next round increased!</b></FONT>")
 
 /datum/game_mode/proc/get_players_for_role(role)
 	var/list/players = list()
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -484,7 +484,7 @@
 					var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[SQL_ckey]'")
 					query.Execute()
 					C.last_cached_weight = weight
-					to_chat(candidate.current, "<font color='green'><b>For playing thru the whole round your chance to become an antagonist increased!</b></FONT>")
+					to_chat(candidate.current, "<font color='green'><b>For playing through the whole round your chance to become an antagonist increased!</b></FONT>")
 
 /datum/game_mode/proc/get_players_for_role(role)
 	var/list/players = list()
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -474,7 +474,7 @@
 		for(var/v in not_chosen)
 			var/datum/mind/candidate = v
 
-			if(candidate.current && candidate.current.client && candidate.current.stat != DEAD) //dead and braindead people wont get their weight increased
+			if(candidate.current && candidate.current.client) //braindead people wont get their weight increased
 				var/client/C = candidate.current.client
 
 				if(C && C.last_cached_weight) //if it's not null, we've updated it during the picking stage
@@ -484,7 +484,7 @@
 					var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[SQL_ckey]'")
 					query.Execute()
 					C.last_cached_weight = weight
-					to_chat(candidate.current, "<font color='green'><b>Your chance to become an antagonist in next round increased!</b></FONT>")
+					to_chat(candidate.current, "<font color='green'><b>For playing thru the whole round your chance to become an antagonist increased!</b></FONT>")
 
 /datum/game_mode/proc/get_players_for_role(role)
 	var/list/players = list()
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -440,6 +440,9 @@ var/datum/subsystem/ticker/ticker
 			else
 				to_chat(Player, "<font color='red'><b>You did not survive the events on [station_name()]...</b></FONT>")
 
+	if(mode.name != "extended (secret)" && mode.name != "extended (announced)")
+		mode.update_not_chosen_candidates() //reward non-antags for making it thru round
+
 	//Round statistics report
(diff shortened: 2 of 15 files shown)


Alexandria404:
What´s the amount of the increase. What if I just leave it open and afk in a locker?

matskuman5:
Say hello to more rampant powergame and murderbone

How does this work? Does it stack? Does it last forever, or decay over time or what?

AlpacaGoesCrazy:
The amount is not changed. It used to be that all people who were eligible for antag but didnt have it got that increase. Now only alive and not braindead people will get it, others will not.

As for how much is that increase:
`weight = min(400, weight * 1.5)`
I didnt touch this part. it increases but has a maximum limit at 400.

Its NOT A BUFF for getting antag, its a NERF for going braindead and not value your life

Jc1469:
I don't think punishing antag targets for being antag targets is a particularly nice thing to do.

AlpacaGoesCrazy:
>  What if I just leave it open and afk in a locker?

@Alexandria404 yes you can do that with this PR, but currently its even more simple: press ready and you get antag weight increase if you didnt get antag.

@Jc1469 on other hand not punishing players who dont value their lifes and go validhunting, jumping out of airlocks or simply going braindead is even worse thing to do

Alexandria404:
I don´t care which way, but we need to somehow not reward afkers

Alexandria404:
possbile solutions:
alive+escaped
afk var

drderp3635:
Alive + escape would just fuck things up if someone succeeds on hijack

AlpacaGoesCrazy:
> Alive + escape would just fuck things up if someone succeeds on hijack

@drderp3635 Antags dont get weight increase anyway, so if noone gets weight increase nothing will change and no things will be fucked up.

ShadowDeath6:
:-1:

a very big part of this game is learning from your mistakes and dying. Someone being unlucky or new to the game being gimped out of antag chances is extremely poor conduct. 

if someone is going braindead at roundstart as an important job then it's up to admins to deal with them, not coders.

Super3222:
jexp

Jc1469:
400 rounds

ghost:
it's a fucking video game and sometimes people don't want to play

this is going to encourage powergame for people who want to be antag more than anything else, and god help EVERYONE if it's like, a fucking nuke round/blob round/clock round/cult round because everyone is going to die in those cases

so in THOSE cases it's a catch-22 because if you don't actually try to stop the cult or whatever and you die doing it you get nothing, and if you don't stop the cult you get nothing anyway

you're going to end up with the same few people powergaming it if they want antag because rather than being braindead the WHOLE round they'll be braindead the majority of it then board the shuttle when it arrives and not be counted as braindead

and they won't actively participate in the round at all, because they won't want to die to the antagonist at risk of losing their antag weight, so it'll end up with a bunch of people quietly jerking off in their departments with no security because nobody wants to be security at risk of being killed by the traitor and losing their antag weight gain for the round, completely unwilling to interact or actually try to stop the antagonist

you've gained no form of activity here, because it'll go one of two ways:

people powergame to the extreme to not die and you accomplish the opposite of what you intended and people circumvent the braindead thing by only not being braindead when whatever amount gets tallied

or 2

people hide away and do nothing when the antagonist shows up which leads to boring rounds and easy wins, and nobody plays security at risk of losing their antag weight because they run the risk of getting killed pretty much every fucking round. i forsee a majority of the crew fleeing to mining pretty much every wiz round rather than actively participate in it at risk of losing their sweet, sweet, antag rolls, because that's all they care about



on top of THAT, you've pretty much guaranteed that nobody but the most robust people will get antagonist, which goes against why antag weight was implemented in the first place: so that everyone had a fair shot at actually getting antagonist at some point, so that you wouldn't have month long dry streaks and the other guy got it constantly. in this case, only people who survive get it, which either means the ones who hide and don't participate but are technically still there, ie mason miami, or the ones who are too robust to die.

shit PR.

Ratvar:
If you are trying to punish antagrollers-how you will punish people who go and get borged roundstart if they didn't roll antag in attempt to roll malf AI/clockcult/emagged status?

AsV9:
I'm a bit skeptical. I believe leaving the system unchanged is probably wise for now. I'd rather have the admins enforce powergaming rules instead of trying to apply a technical solution.

Closing this pull request unless someone has objections in the next 24 hrs.
You may, of course, open a new pull request any time.

AlpacaGoesCrazy:
> I'd rather have the admins enforce powergaming rules instead of trying to apply a technical solution.

@AsV9 This is just VERY wrong thinking here.

I suggest i change it so antag weight will be added to everyone who is non-braindead at round end. So you dont have to be alive but you cant just close game and wait for next round ping on discord to try rolling antag again. (Sorry i didnt reply earlier, was really busy)

AsV9:
I don't think making admins properly enforce powergaming rules is quote 'very wrong' here, but I'd be alright with a solution that just encourages people to stay around for another round.

Then again, your argument for preventing powergame or encouraging people to value their lives doesn't work in this case.

AlpacaGoesCrazy:
Enforcing rules thru game mechanics is easier for everyone: less works for admins and lesser chance that someone manages to break that mechanic. Thinking that a few additional rules will do the trick is just wrong. Rules are not always enforced, they are hard to memorize for newer players (at this point there are too many), they are not always clear and finally there always some rulebreakers who dont get caught.
On other hand its hard to trick a game mechanic and if someone manages to do so you can just adjust it  so it wont be possible to take advantage on said mechanic in the future.

Anyway, @AsV9  what do you think about my last solution:

> i change it so antag weight will be added to everyone who is non-braindead at round end.

Will this do the trick? It surely doesnt punish those who died but encourages people to actually play the game (or not close the client and at least pay /some/ attention to game process)

AsV9:
Sure, I'd be alright with trying that.

code/game/gamemodes/game_mode.dm
```diff
@@ -484,6 +484,7 @@
 					var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[SQL_ckey]'")
 					query.Execute()
 					C.last_cached_weight = weight
+					to_chat(candidate.current, "<font color='green'><b>For playing thru the whole round your chance to become an antagonist increased!</b></FONT>")
```
AsV9: Grammar police.

AsV9 (changes requested):

Alexandria404:
Still issues with implementation, we are discussing a possible fix.

code/game/gamemodes/game_mode.dm
AlpacaGoesCrazy: I am bad at grammar, please suggest how it should be :(

Partheo: Honestly I feel like this should be hidden. I don't feel like it should display a message.

AsV9: `thru` is too informal to use in this context, use `through` instead or do what Partheo says.
both satisfy the police

ghost:
Enforcing rules through game mechanics is a cute idea that doesn’t work in practice. It’s shoddy, mostly nonfunctional, and leaves a lot of area for particular oddities or exceptions that an administrator would be able to discern. And that’s with a good system, not this shit.

You’ve also failed to address half my points.

AsV9 (approved):

ThatLing:
I'm gonna close this since it's been pretty dead and you haven't implemented the requested changes yet. Feel free to reopen it or make a new PR if you do.

Issue #2851, opened 2017-12-16
New atmos is completely shielded against radstorms.


Moja10:
Something about the new atmos means that radstorms doesn't affect any of it. During a round, a radstorm hit, and rather than being coated in that delicious green radiation, it was completely shielded, like it's a maint tunnel. 

Should be easy to reproduce, just spawn a rad storm and see how it's not affected.

Partheo:
Thanks for posting this- I'll make a PR that fixes this ASAP.

Partheo:
![image](https://user-images.githubusercontent.com/10659549/34091021-fc96261e-e388-11e7-8906-b8864a120b3b.png)
Just tried to reproduce this now. Radstorm seems to work fine. Are you sure about this?

nichlas0010:
![image](https://user-images.githubusercontent.com/20558591/34735611-55b0e6a0-f570-11e7-8f1a-b9aecc384e0c.png)

PR #2860, opened 2017-12-18
New atmos department- tweaks

Partheo:
:cl: Partheo
rcsadd: A box of replacement lights has been added to atmospherics.
bugfix: Some of the consoles and pumps in atmospherics have been updated with correct names.
tweak: Small map tweaks to atmospherics and the engineering foyer.
/:cl:

That's it. Just a little batch of fixes and tweaks. I wanted to solve #2851 as well, but during my testing I found no issue.
I think John mentioned that there were too many cameras in Discord. There definitely were too many, so I've removed about half of them. AI vision is still fairly ideal.
Thanks to Empirical for the help spotting most of the issues, and for suggestions.


Amelia0010:
and this please 
image

Partheo:
Thanks for the tip, Nich. Fixed in my next commit.

Also, while I'm at it- thoughts on this modification? I've added a window to the engineering foyer.
image

apodgo2:
Any chance you could add in a pair of magboots to each of the Atmos Tech lockers? It's ridiculous that the Tech's, despite almost being guaranteed to interact with pressure differentials, don't have access to magboots unless they have the suit. 

I can understand that this might be a balance change so if not, no worries.

Partheo:
The atmos techs get access to one pair of magboots to share between the three- given that there's only one hardsuit, I'm fine with that staying how it is. 

I understand how ridiculous it is that they don't have full access to them, but I feel like it would just cheapen the item itself if they had too many.

Anyone else have thoughts on this?

Altoids1:
The tools that are available to a job on-spawn _defines_ what that job is. I think adding magboots would give an advantage to atmos techs over engineers (at least on higher pops) in dealing with areas with large pressure/temperature differences (breaches and fires). While it does cheapen the item somewhat, it also empowers the Atmos Tech, making their role more useful and unique over what the Station Engineers provide, and I think that would probably be worth it.

Jc1469:
atmos techs really don't need magboots.  they look cool with casually worn atmos jumpsuit, thats about it.

The vast majority of the time they'd need magboots, they'll be in a hardsuit so they can just turn on the jetpack stabiliser.  Or, I dunno, ask to be let into EVA.

grubstank:
I think it would be good to add another random pair into the room somewhere. Vent space wind has dragged me away from air alarms on countless occasions - breach or no.

PR #2861, opened 2017-12-18
Adds its own suicide for rubber ducky

AlpacaGoesCrazy:
Fixes this:

If you have more fun ideas on how rubber duck suicide should look please do tell me.

:cl:  
rscadd: You just discovered a new fun way to play with a rubber duck.
/:cl:

### DIFF
--- a/code/game/objects/structures/watercloset.dm
+++ b/code/game/objects/structures/watercloset.dm
@@ -295,6 +295,13 @@
 	icon_state = "rubberducky"
 	item_state = "rubberducky"
 
+/obj/item/device/assembly/bikehorn/rubberducky/suicide_act(mob/user)
+	user.visible_message("<span class='suicide'>[user] tells [src] that they don't want to live anymore! It looks like they're trying to commit suicide..</span>")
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/watercloset.dm
```diff
@@ -295,6 +295,13 @@
 	icon_state = "rubberducky"
 	item_state = "rubberducky"
 
+/obj/item/device/assembly/bikehorn/rubberducky/suicide_act(mob/user)
+	user.visible_message("<span class='suicide'>[user] tells [src] that they don't want to live anymore! It looks like they're trying to commit suicide..</span>")
```
ShadowDeath6: inconsistent punctuation with other suicide messages I'm pretty sure

AlpacaGoesCrazy: Can you suggest what should be changed?

ShadowDeath6: punctuation, namely the .. At the end which isn't only inconsistent but not gramatically correct either

PR #2874, opened 2017-12-23
Malf-doomsday now uses world.time instead of process()

Amelia0010:
Before, any amount of lag would mean that the doomsday timer would say it was "450 seconds" when in reality it was usually 900, if not more. This makes the doomsday timer actually count down accurately

image


AsV9:
RIP byond seconds

PR #2884, opened 2017-12-26
Ports shuttle ripples

Amelia0010:
Ports: https://github.com/tgstation/tgstation/pull/18723

Adds ripples to outline where shuttles are landing, to prevent crushing from custom shuttles.

https://gyazo.com/dfd506657e9d094c77f02a45df28b6fb

#### Changelog

:cl: Coiax
rscadd: Due to budget cuts, the shuttle's hyperspace engines now create a visual distortion at their destination, a few seconds before arrival. Crewmembers are encouraged to use these "ripples" as an indication of where not to stand.
/:cl:

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -40,6 +40,7 @@
 #define SPACEVINE_LAYER 4.8
 #define SPACEVINE_MOB_LAYER 4.9
 //#define FLY_LAYER 5 //For easy recordkeeping; this is a byond define
+#define RIPPLE_LAYER 5.1
 
 #define GHOST_LAYER 6
 #define AREA_LAYER 10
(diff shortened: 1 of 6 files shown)


Jc1469:
I for one prefer shuttles smacking people and killing them.  We already added a travel time to the mining one, and the emergency one has a timer anyone can see..

wwtall1:
its for the CUSTOM ONES that are fucking shaped different, so if its a roller coaster you know that there's a FUCKING SPACE CAR crashing ten tiles north.

ghost:
@wwtall1 Actually be correct before you post.

It is not for the custom ones. It is for all shuttles.

The reason these were added is because /tg/ has a custom shuttle called 'asteroid with engines strapped onto it', which is...exactly what it implies. And does exactly what you think it does.

AsV9:
I'm going to merge this, and we'll see what happens. It *is* funny when people get run over by the shuttle though.

PR #2892, opened 2017-12-29
Fix hotkeys

### LINKED ISSUE #2891: Hotkey (tab) mode delay
**_Ever since the update on the 27th of December I have noticed that when pressing tab to switch from hotkey mode to normal, there is a slight delay before the text box opens or closes._**

This is problematic because like me, and other players, we play in both hotkey mode (movement with wasd, r to throw, x to switch, z to activate...) and normal mode (say, me, and cancel-camera-view...). Now with the slight delay it makes the game kind of unplayable, and one of the reasons why I regularly play on yog is because it is a server that doesn't have this issue, until recently.

I hope this is relatively easy to fix, it's kind of an annoying issue. x

ThatLing:
This adds missing azerty keys to the default hotkeys. *Should* also fix #2891 since it doesn't use verbs anymore.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1192,9 +1192,11 @@ var/list/preferences_datums = list()
 
 				if("hotkeys")
 					hotkeys = !hotkeys
+					user.client.sethotkeys(1)
 
 				if("hotkeysmode")
 					hotkeysmode = !hotkeysmode
+					user.client.sethotkeys(1)
 
 				if("tgui_fancy")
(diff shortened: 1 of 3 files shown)


Time-Green:
<i>should</i>?
Good job on the quick fix, but did you test to make sure the delay is gone?

ThatLing:
The delay is caused by the verb being executed on the server. Since my ping is 0 on my local server, I can't really test it unless you know of a sane way of increasing the ping. It doesn't use verbs anymore and instead uses the system it used before, so technically it should be fixed.

PR #2896, opened 2018-01-01
Buffs the bulletproof helmet

Time-Green:
I painfully discovered that the bulletproof helmet actually only gives you 10 extra armor compared to the normal helmet. While bulletproof ARMOR has 80 bulletresistence. Since literally everyone aims for the head, it only makes sense to have the gear actually matter in a fight, instead of it still being a three shot kill.
For reference:
Normal helmet: 30 bulletresistence
Bulletproof helmet : 40 resistence (now 80 as of this PR)
Bulletproof armor: 80 bullet resistence

:cl:
tweak: Bulletproof helmets are actually resistent to bullets now. 
/:cl:


simonset55:
Git gud instead

Amelia0010:
Apparently TeeGee also has this compile error. gulp is fun

PR #2910, opened 2018-01-06
changed a bunch of projectiles to not make bullet holes, including most magic bolts and the KA

### LINKED ISSUE #2822: Kinetic Accelerator leave bullet holes
So nearly every single rock you mine leaves a bullet hole on the floor....

Amelia0010:
fixes #2822


Time-Green:
Could we have people denting their faces into walls if they hit it hard enough?

PR #2920, opened 2018-01-07
removes a bunch of src. from the communications console, and makes sh…

Amelia0010:
…uttles not buyable during escape/docking


Partheo:
Sad!

PR #2932, opened 2018-01-12
Datum Component System

Time-Green:
Ports: https://github.com/tgstation/tgstation/pull/29324 by @cyberboss who ported it from /vg/

The link to the original PR has more info
I'll summarize what this does:
"instead of copypasting slip code everywhere for stuff you slip on, you can do AddComponent(slip) something like that. That component holds the slip code and is triggered once you walk over something. This allows us to eat pens and slip on clowns more easily, and is also the only reason I am porting this

This doesn't actually do anything yet, it's just the framework
:cl: 
rscadd: Ported components from /tg/
/:cl:
I'll just go and put myself in the changelog because too much time was spent fucking dissasembling their callback system and making everything work correctly. Wich it probably still doesnt, but I'm probably the only one that's gonna use this.


1fbff5f83b23d39d38b1dfcb4cac8d9b:
The datum component system (DCS) accomplishes two Principles of design:

1. polymorphism by composition rather than inheritance
2. dynamic allocation of behaviors at runtime
Both of these principles are accomplished more efficiently and more simply without this system.

Principle 1 can already be done in Byond by using either vars or procs.

Polymorphism by composition can be accomplished by adding a variable containing an instance of something (in theory this should usually be a lightweight datum, but in practice it often ends up being an atom either because of Byond's special handling of atoms or because people have shoehorned the composition strategy into something that was originally done by inheritance) that handles an object's compositional behavior for it. An example of this is traitor uplinks. All datums of type /obj/item have a var called hidden_uplink that stores a traitor uplink, and when an instance of that datum is handled in a certain way (such as typing a particular password into a PDA's messenger), it will trigger the uplink's access event by calling a proc on it, if the uplink exists. The code that handles the trigger is put on the datum that holds the uplink rather than the uplink itself, because the object holding the uplink understands its own behavior and can most efficiently trigger the "access traitor uplink" event, while the uplink is just along for the ride and does not particularly care how or when it is accessed. In turn, the datum holding the uplink does not have anything to do with the uplink's operation other than sending the access event, because it does not know or care how the uplink works.

The downside of this is that all instances of type /obj/item now have an additional variable. The way Byond's object system works, this var does not take up any additional memory on instances that do not use it, but it does slow down the accessing of other variables that each instance is not using. This is strictly an issue with how Byond handles variables and could potentially be optimized away by Lummox at any time.

An alternative to this is the proc system that the get_cell() example in the DCS documentation cites. This is more efficient because procs are accessed with a binary search rather than a linear one, so in order to slow down the accessing of a proc by an increment of 1, you have to double the number of procs a type has. Even if we had a tremendous number of compositions that might be on any datum, which we do not, it would not significantly affect performance. This system is not used as much because it is syntactically easier to use a variable than a proc, because Byond does not know (or rather, in the style of byond, it does not by default assume) the type of a proc's return value at compile time, and so you must always typecast it.

A perfect example of Principle 2 in existing code is the species datums for /mob/living/carbon/human (actually it is stored in a mob's dna, but it is currently only ever used by humans). It was originally made to handle changing a mob's species at runtime without having to copy over a large number of variables and external references to a new mob whenever a species was changed. The DCS is similar to datum species in many ways, other having a wider scope and having a list of components rather than a single species. However, the DCS using signals while datum species uses procs.

Signals are worse than procs because they add an additional layer of indirection for absolutely no gain. Not only do you still have proccall overhead, you have doubled it, because you have to call SendSignal() which then calls whatever proc it looks up. The only advantage of this is that you can dynamically add and remove procs from a particular signal at runtime, which is unnecessary for a system that should in the first place only be made up of atomic components (that is, components whose behavior cannot reasonably be split up into multiple independent behaviors).

Additional overhead is added because every single component must receive every single signal, even the ones it does not care about. Every time an atom is crossed, it must send COMSIG_MOVABLE_CROSSED to every component it has, whether that is the slippery component or the battery component. In theory this could be optimized away with event subscription lists, but that would have the cost of additional memory usage on every instance using components.

I have not seen an example or theory on /tg/ where the signal system would ever be used meaningfully. Signals are just a completely redundant component system for a component system.

Yogstation has had a system remarkably similar to DCS, though more limited in scope, for almost a year now. #1042 added the module framework to the current repository, and it was on the old pre-rebase repository before that. It has, to my knowledge, never been used since its introduction, either because it has not been necessary or because it is easier to add hooks where you need them than to expand an existing system that uses similar hooks.

I believe that, if the DCS were added to this repository, it would either never be used or it would only be used in cases where an alternative would have been better.

And, in the words of Forrest Gump, "That’s all I have to say about that".

Time-Green:
Wait we already had a thing similair to this?
fuck

Time-Green:
Couldnt you have fucking told me that while I was ranting in #coder about how horrible it was to port?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
sorry

Time-Green:
im sorry for swearing at you
*brohug*

Time-Green:
Right so, X's framework can only be used in extremely situational and specific ways and I've tried for days to make it work. Sadly, it's not made to work the way I wanted.

The main argument against components was that no one would use it. Well, I've been trying for weeks

Time-Green:
Doesn't actually work but I fixed that in another PR. Unless someone else actually wants to use it, I'm gonna leave it like this

Time-Green:
i fixed it reee

ghost:
@Time-Green why did you get the oldest possible version of DCS

Time-Green:
Because I needed it in a pinch and I see no point in spending 5 hours porting and debugging everything for something only I will probably ever use

PR #2938, opened 2018-01-14
Anthrax

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Adds a new uplink item, the anthrax packet. You can use it to 'poison' any item. Any person who touches it without gloves will become infected with anthrax and die within a few minutes. The cure is charcoal, spaceacilin, and perfluorodecalin, administered all at once.

The cost of one packet is 6 TC.

I feel like this is a bit underpowered. A lot of people on the station wear gloves roundstart, and it does not transmit from person to person, only via contact with the poisoned item. Any suggestions to make it scarier but not "kill the entire station" horrible?

:cl:  
rscadd: The Syndicate has developed a new bioweapon, the anthrax spores. Handle with care.
/:cl:


Alexandria404:
I like it a lot.
Very situational, but strong if your target has no gloves and it´s rather stealthy too.
Is there a way to tell the items in contaminated?

Time-Green:
I’m begging you, please use the modular framework x made. Creating a var for something so specific just seems dirty

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Alex: Nope, there's no way to check right now.
Time: I'll look into it later

Alexandria404:
Nice. With no way of checking I feel it´s immensely strong as a stealthy killer.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
The infection may be extremely stealthy, but the death itself isn't... Any idiot with half a brain will start yelling for help once they start vomiting blood and stumbling in circles.

Ratvar:
Make meat of poisoned contain anthrax?
Make bumping into antharax covered structure with uncovered bodyparts infect people?
Make washing/cleaning items remove anthrax?
Make advanced chemical scaner from R&D check for anthrax?
Anthrax poisoned sharp items applying it to people wia attacks/throws?
Walking over anthrax without shoes? ;Clown bnnas are poisond halp

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Those are some good ideas, thanks.
I also have an idea, clothes can transmit the disease too, so one can create a clown mask cursed with pestilence. But again, this can spiral out of control for a non-hijack traitor

Ratvar:
Nonhjack traitor/any antag/griefer can let plasma out in atmos/make bombs, so, no precautions should be taken here - they already can fuck rounds up on large scale without anthrax. What's the cost on syndie bomb?

ghost:
This is fine as is. If you make it so that it spreads on contact with things, then it's going to be a lethal version of fungal tuberculosis.

Precautions must be taken to avoid this being a 6tc round ender. Leave it as is.

Could potentially make this a virologist-only job traitor item.

AsV9:
image
:thinking:

1fbff5f83b23d39d38b1dfcb4cac8d9b:
shush, the probabilities were way too high, already changed em

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Changes:
- Putting on equipment which is contaminated will infect you. Will need some changes though
- Illness starts out with almost no symptoms and progresses slowly but ramps up and hits you hard near the end
- Sterilizine will sterilize contaminated objects
- Changed oxygen damage so you dont get hit with a flat amount

MenacingManatee:
Quick question, does having all 3 in your system give a chance to heal per tick, or does it auto remove the anthrax from your system?

1fbff5f83b23d39d38b1dfcb4cac8d9b:
There is a 5 percent chance per tick to cure, same as with all/most other diseases

MenacingManatee:
Maybe not charcoal for the cure then, since it removes other chems from the body per tick, meaning its a lot harder to get cured and takes more of the other 2 chems. Instead you could use maybe trico? Unless of course you're wanting that to make it more difficult to cure.

PR #2814, opened 2017-11-29
Optimizes, bugfixes for NTSL; adds two new channel vars

Altoids1:
### Making NTSL Not Be Shite: Part 1

This is mostly just a packet of minor additions/bugfixes done during a quick scan of the code I did yesterday. All of them have been personally tested, pinky promise.

#### Changelog

:cl:  
rscadd: Added two new on-boot NTSL variables, $aiprivate and $centcom
bugfix: NTSL will now properly handle having $content be set to a number value.
bugfix: NTSL should now parse comments correctly.
bugfix: Optimized how NTSL deals with numeric operators and comments. Should run/compile 1% faster or something. *shrug
spellcheck: Fixed minor typoes within some of the telecomms computers
/:cl:

PR #2948, opened 2018-01-17
Revert "Merge pull request #2814 from Altoids1/master"

### LINKED ISSUE #2942: I broke NTSL AND I DON'T KNOW WHY
https://github.com/yogstation13/yogstation/pull/2814/files

Apparently my rewrite of a particular part of Scanner.dm has caused NTSL to stop being able to compile correctly. It seems that the pointer it uses to parse through the original code text gets into weird places, resulting in random, nonsensical errors about unknown tokens or unterminated strings.

Looking at the PR, you might say "Aaahh yes, I see. In my wise coder wisdom I notice that you have twochars be created using only one character, not the two required for this to work, causing it to ignore comments."

So I fixed that and recompiled, and it still didn't work.

So you might say "Yes, young Padawan, you must not remove var/commenting from this list, for ye know not what it does."

So I added it back again, recompiled, and it still didn't work.

So you might say "Heed these words newcoder, for I am all-knowing: Reduce ye function to its purest form, and rid thyself of anything suspicious or bug-inducing."

Fine, here is the new version of proc/ReadComment()
```
		ReadComment()
			// Remember that we still have that $codepos "pointer" variable to use.
			var/longeur = lentext(code) // So I don't call for code's length every while loop
			if(copytext(code, codepos, codepos+2) == "//") // If line comment
				while(++codepos>longeur)
					if(copytext(code, codepos, codepos+1) == "\n") // then stop on the newline
						return
			else
				while(++codepos>longeur) // If long comment
					if(copytext(code, codepos, codepos+2) == "*/") // then stop on any */ 's'
						return
				//Else if the longcomment didn't end, do an error
				errors+=new/scriptError/UnterminatedComment()
```
And it's now called like this:
```
		for(, src.codepos<=lentext(code), src.codepos++)

			var/char=copytext(code, codepos, codepos+1)
			var/twochar=copytext(code, codepos, codepos+2) // For finding comment syntax
			if(char=="\n")
				line++
				linepos=codepos

			if(ignore.Find(char))
				continue
			else if(twochar == "//" || twochar == "/*")
				ReadComment()
			else if(end_stmt.Find(char))
				tokens+=new /token/end(char, line, COL)
			else if(string_delim.Find(char))
				codepos++ //skip string delimiter
				tokens+=ReadString(char)
			else if(options.CanStartID(char))
				tokens+=ReadWord()
			else if(options.IsDigit(char))
				tokens+=ReadNumber()
			else if(options.symbols.Find(char))
				tokens+=ReadSymbol()
```

And, guess what, still doesn't work.

I've been trying to fix this for the past 2 hours and have to sleep ,so I'm leaving this here for whenever I/somebody figure out why this is happening or Asv9 reverts the PR. :honk:

AsV9:
This reverts commit 932990dd7ac5fa2ca02f28c63616d241c30062fc

ree

closes #2942

### DIFF
--- a/code/game/machinery/telecomms/computers/telemonitor.dm
+++ b/code/game/machinery/telecomms/computers/telemonitor.dm
@@ -113,7 +113,7 @@
 		var/newnet = stripped_input(usr, "Which network do you want to view?", "Comm Monitor", network)
 		if(newnet && ((usr in range(1, src)) || issilicon(usr)))
 			if(length(newnet) > 15)
-				temp = "<font color = #D70B00>- FAILED: NETWORK TAG STRING TOO LONG -</font color>"
+				temp = "<font color = #D70B00>- FAILED: NETWORK TAG STRING TOO LENGHTLY -</font color>"
 
 			else
 				network = newnet
--- a/code/game/machinery/telecomms/computers/traffic_control.dm
+++ b/code/game/machinery/telecomms/computers/traffic_control.dm
@@ -281,7 +281,7 @@
 
 		if(newnet && canAccess(usr))
 			if(length(newnet) > 15)
-				temp = "<font color = #D70B00>- FAILED: NETWORK TAG STRING TOO LONG -</font color>"
+				temp = "<font color = #D70B00>- FAILED: NETWORK TAG STRING TOO LENGHTLY -</font color>"
 
 			else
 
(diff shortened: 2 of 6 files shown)


Altoids1:
I really wish I knew why that happened.

PR #2953, opened 2018-01-18
Fixes NTSL, but, like, more so.

Altoids1:
### There were more bugs and I fixed them

So the new comment hotfix doesn't completely work, since it actually overskips the comment syntax by 1 place, meaning long-comments that are inlined into code weren't working, so I fixed that.

I have additionally fixed the Type Mismatch bug, as well as improved some of the error messaging around that error while I was there.

It's fixed, for sure this time. I promise. Really.

#### Changelog

:cl:  

bugfix: Nanotransen will now, for sure this time, stop distributing faulty telecomms equipment.

/:cl:


ghost:
Make your changelogs make sense and describe what they actually do instead of in-jokes newer players won't understand.

Amelia0010:
TCOMMS STOP NO WORK, K?

Altoids1:
Nich gets it

AsV9:
image
pls work off branches

Altoids1:
But it has master in the name. That makes it the better branch, clearly.

PR #2955, opened 2018-01-18
Order admin notes from new to old

ThatLing:
This makes new notes  appear above older ones.

### DIFF
--- a/code/modules/admin/sql_notes.dm
+++ b/code/modules/admin/sql_notes.dm
@@ -120,7 +120,7 @@
 		output = navbar
 	if(target_ckey)
 		var/target_sql_ckey = sanitizeSQL(target_ckey)
-		var/DBQuery/query_get_notes = dbcon.NewQuery("SELECT id, timestamp, notetext, adminckey, last_editor, server FROM [format_table_name("notes")] WHERE ckey = '[target_sql_ckey]' ORDER BY timestamp")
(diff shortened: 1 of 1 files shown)


matskuman5:
[a quiet "ree" in the distance...]

retrosquid64:
I'm not an admin, but isn't having the older ones at the top better? They could get buried in particularly busy rounds, right?

Amelia0010:
Nope, notes are on a per-person-basis, so if they get buried in a single round then the person doing it probably deserves a ban at that point. 

Just to show what it looks like right now, here's my notes:
image

And making them show from new to old is better, since the newer notes are generally more relevant.

retrosquid64:
Whoops, confused it with tickets.

Amelia0010:
Tickets are currently shown from new to old, since some admins tend to forget to close their tickets, meaning that having 17 tickets open that are solved, would just bloat the ticket interface

AsV9:
I mean this is going to get some admins confused and angrr but let's just have it and see what happens.

Xantam-13:
Same thing I posted on the admin vote;
For people with small amounts of notes, the date order really doesn't matter at all. 

For people with lengthy notes, it's more efficient to see their most recent offences first, than their oldest, you tend to use recent notes against a player when issuing a punishment rather than age-old notes right? So why want to see oldest first?

PR #2765, opened 2017-11-10
changes "dna extracted" objective to "lifeforms absorbed" objective

ShadowDeath6:
:cl: ShadowDeath6
tweak: Changelings now need to absorb lifeforms with the Absorb ability (or devour, if you're into that) in order to complete their "extract DNA" objective.
/:cl:

PR #2957, opened 2018-01-19
Removes extraction sting, changes absorb objective

Amelia0010:
Absorb objective changed to somewhere between 2 and 10, depending on server pop. for every 5 pop, the amount increments by 1, up to a max of 10, at 50

Closes #2656 Closes #2765 

#### Changelog

:cl:  
rscdel: Extraction Sting has been removed from changelings, absorbing is now necessary
tweak: absorb objective scaling changed, for more information check github
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/changeling.dm
+++ b/code/game/gamemodes/changeling/changeling.dm
@@ -128,7 +128,7 @@ var/list/slot2type = list("head" = /obj/item/clothing/head/changeling, "wear_mas
 
 	var/datum/objective/absorb/absorb_objective = new
 	absorb_objective.owner = changeling
-	absorb_objective.gen_amount_goal(6, 8)
+	absorb_objective.gen_amount_goal(10)
 	changeling.objectives += absorb_objective
 
 	if(prob(60))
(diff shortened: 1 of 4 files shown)


drderp3635:
Make it one for every ten, not five

Alexandria404:
There are discussions going on in council about how to address the issue.
Nich was not aware of it at the time.
Please don´t merge it for now as we are still considering possible directions.

Alexandria404:
After repeated, repeated talks, the conclusion from council seems to be majority is fine with it if the numbers are lowered to 1 for every ten.

retrosquid64:
I think not having the extraction sting is good, makes changeling rounds more paranoia filled.

Ikoden:
It's a good change yet there is one problem that I see. I think that people on low pop (of course on other pops too, yet it would be the biggest problem on low pop) will see it as a permission for murderbone absorbing. "Oh, I've already got 3 absorbs. I've finished everything else, so let me be an even better ling by absorbing some more people.", "I've absorbed 2 people, why not absorb 6 people.", "If absorbing people is an objective, then I can totally absorb as many as I want and it certainly does not constitute murderbone." - I can see those as thought processes of many people. Just my 2 cents.

Amelia0010:
it's now 1 for every 10, and if people see "kill 2 people" as "let me kill 14 people to be good" then they should just die

PR #2961, opened 2018-01-20
I hate our admin team

Amelia0010:
Reverts yogstation13/yogstation#2955

Revert for the time being, as all of our admins decided to throw a hissy fit about a change they think is bad, but never said is bad, being merged. 

A staff vote will be created about it (because of course it will) after which we can implement it


Partheo:
Classic Yogs

Jc1469:
:popcorn: 
probably could've guessed this would've happened.   for the same reasons as the log change one. People are used to one thing so it should either have an option or not exist, especially for admin-related things.

PR #2967, opened 2018-01-21
Stunbatons faintly glow when turned on

retrosquid64:
This PR makes stunbatons glow softly when turned on. Barely enough to see anything in the dark, but the wielder will be visible. Luminosity is 0.9 .
#### Changelog

:cl:  retrosquid64
tweak: Stunbatons now glow faintly when turned on
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -94,6 +94,16 @@
 			to_chat(user, "<span class='warning'>[src] does not have a power source!</span>")
 		else
 			to_chat(user, "<span class='warning'>[src] is out of charge.</span>")
+	if(status == 1)
+		if(loc == user)
+			user.SetLuminosity(0.9)
+		else if(isturf(loc))
+			SetLuminosity(0.9)
+	else
+		if(loc == user)
+			user.SetLuminosity(0)
+		else if(isturf(loc))
+			SetLuminosity(0)
 	update_icon()
 	add_fingerprint(user)
 


code/game/objects/items/weapons/stunbaton.dm
```diff
@@ -94,6 +94,16 @@
 			to_chat(user, "<span class='warning'>[src] does not have a power source!</span>")
 		else
 			to_chat(user, "<span class='warning'>[src] is out of charge.</span>")
+	if(status == 1)
```
Time-Green, line 97: You can do if(status) without the == 1
if(status) just checks if it's true. That being literally anything but null, 0 and FALSE
If status actually goes further then 0 and 1, ignore me

retrosquid64: I see, thanks.

ThatLing: It's can only be 0 or 1. See this.

retrosquid64: I could change it, but does it really matter for now?

Time-Green: It's technicly not wrong, but it's against standards and honestly it's really bothering me to the point I will make a seperate PR to fix it if you dont

Amelia0010:
If you activate the stunbaton so it glows, then drop it on the ground, the light stays on the user, and the stunbaton lies in darkness

https://gyazo.com/0f5891083ccb577017d3bf10876ef89f

code/game/objects/items/weapons/stunbaton.dm
retrosquid64: If it's against the standards I will happily edit this PR to accommodate them.

retrosquid64:
On it.

Time-Green:
It may be because you made the user light up, and not the actual baton
Same with the turf, you're making the turf glow, not the user

code/game/objects/items/weapons/stunbaton.dm
```diff
@@ -94,6 +94,16 @@
 			to_chat(user, "<span class='warning'>[src] does not have a power source!</span>")
 		else
 			to_chat(user, "<span class='warning'>[src] is out of charge.</span>")
+	if(status == 1)
+		if(loc == user)
+			user.SetLuminosity(0.9)
```
Amelia0010, line 99: Please use [AddLuminosity](https://github.com/yogstation13/yogstation/blob/6c3d36b0fc06c2a7a874d0d45fb84008695ac9c7/code/modules/lighting/lighting_system.dm#L218) instead.

If you use SetLuminosity, you run into this issue: https://gyazo.com/a9c7f8677478c572547a137941ef2f6f

Amelia0010 (changes requested):

ghost:
Does veil turn off the stunbaton if a shadowling uses veil.

code/game/objects/items/weapons/stunbaton.dm
ghost: or instead maybe point out that he's changing the luminosity of the fucking character lmao

retrosquid64: I think the flashlight does this too, that's what I based it on.

PR #2971, opened 2018-01-22
Fixes OOC

Amelia0010:
No more ick ock, no more salt, only pure sweet admemes

:cl:
fix: players can no longer use OOC
/:cl:


Kmc2000:
Good pr

PR #2991, opened 2018-01-24
Removes randseed from NTSL

ThatLing:
There is literally no reason NTSL should be able to alter the random seed. This can potentially be used to trigger random events at will.


Altoids1:
It doesn't actually need randseed to get consistently random results; the server naturally does the randseed itself on-gamestart, and there are constant calls to rand() across the game to make it fly through the generated list regardless.

I approve.

PR #2992, opened 2018-01-24
Fix examine exploit

ThatLing:
Currently, you can examine tiles you cannot see. This will show you what type of floor the tile has but will give you no information about any possible items. This information can still be used to identify cults.


namjay1170:
I like it

PR #2996, opened 2018-01-25
Fixes Suit Storage Units losing auxiliary slots

### LINKED ISSUE #2995: Suit Storage Units can lose their auxiliary slots
The captain's Suit Storage Unit (SSU) and atmospheric's SSU both have an auxiliary slot. If you remove all the items from the SSU and put a person into it and then take them out, the auxiliary slot(s) in that unit will be gone. The auxiliary slots will also be removed if the items inside the SSU are incinerated.

Yolopanther12:
This fixes Suit Storage Units losing their auxiliary slots when a human is
placed inside of and then removed from a Suit Storage Unit or the items
inside are incinerated. Instead of setting extra_items to null when
drop_contents is called, it now just sets it to an empty list.

Refer to issue #2995. Closes #2995.

:cl:  
bugfix: fixed suit storage units losing their auxiliary slots  
/:cl:

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -178,7 +178,7 @@
 	mask = null
 	shoes = null
 	occupant = null
-	extra_items = null
+	extra_items = list()
 
 /obj/machinery/suit_storage_unit/ex_act(severity, target)
 	switch(severity)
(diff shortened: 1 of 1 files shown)


Yolopanther12:
I'd like to just note here that I still am very new to BYOND programming so this fix could be atrocious. If it is, I hope it's at least able to convey the issue at hand so some else can properly fix it.

PR #2990, opened 2018-01-24
Fix borer event

ThatLing:
Borers now work again. This also slightly nerfs the event by reducing the amount of borers spawned. This took way too long for a meme.

:cl: ThatLing and MacHac
bugfix: The borer event can now happen again.
/:cl:

PR #3004, opened 2018-01-26
Makes blob exclusively a midround antag

ThatLing:
Ports https://github.com/tgstation/tgstation/pull/31780 with some changes to support the 'Split Conciousness' power and a few minor bug fixes. Refer to the /tg/ PR for more information.

This essentially makes the blob a midround antag. They can spawn during any gamemode. If left alone, they will quickly convert tiles. Once they have converted enough, they reach critical mass. This ends the round and gives the blob infinite points. This is an incentive for the crew to actively destroy the blob. It also asks people now if they want to become a blob overmind.

## Note to HC
This edits the config which isn't up-to-date on the github.

### Side-note
This will have merge conflicts when #2990 is merged.

:cl:  
rscadd: The blob is now a random event.
rscdel: The blob gamemode is removed.
rscadd: The blob can now win in any mode by gaining enough tiles.

/:cl:


Moja10:
I like blob as a midround antag, and as a roundstart antag. I think midround blob currently is toooooo weak but roundstart is decent. Just my two cents.

ghost:
Midround blob gets buffed to be normal blob tier @Moja10

Nickvr628:
Errrrrr no.

code/game/gamemodes/blob/blobs/core.dm
```diff
@@ -63,50 +66,20 @@
 		overmind.update_health_hud()
 
 /obj/effect/blob/core/Life()
-	if(!overmind || !overmind.client)
-		create_overmind()
+	if(!overmind)
+		qdel(src)
```
Amelia0010, line 70: You sure you wanna just delete the blob, not find a replacement if they ghost or DC?

Alexandria404:
While I can see mid round blob being an interesting concept, fully removing the standalone blobmode doesn´t sit well with me and a lot of others.
I can see balance changes being needed and issues with it, but I´d rather not see it go.
Furthermore I feel that a blob midway in your average highpop round where it would spawn I assume would be next to unstoppable/very hard due to immense chaos next to other antags. Among other things.
Rounds are immensely chaotic at that stage anyway and I´m unsure whether this will work out.
Where I´d definitely like to see how it will play out, removing the blobmode for this is wrong in my opinion.

I´ll be asking council (who might not check the git that often) and will report back with more thoughts as it´s a major change.

code/game/gamemodes/blob/overmind.dm
```diff
@@ -78,6 +123,8 @@
 		BM.overmind = null
 		BM.update_icons()
 	overminds -= src
+	SSshuttle.emergencyNoEscape = 0
```
Amelia0010, line 126: this might fuck with rev rounds, no?

code/modules/events/blob.dm
```diff
@@ -1,23 +1,16 @@
 /datum/round_event_control/blob
 	name = "Blob"
 	typepath = /datum/round_event/blob
-	weight = 5
+	weight = 10
 	max_occurrences = 1
 
 	min_players = 20
```
Amelia0010: current minimum for blob is 30, might wanna change this to reflect that

code/modules/events/blob.dm
```diff
@@ -1,23 +1,16 @@
...
 	name = "Blob"
 	typepath = /datum/round_event/blob
-	weight = 5
+	weight = 10
 	max_occurrences = 1
 
 	min_players = 20
-	earliest_start = 36000 //60 minutes
 
 	gamemode_blacklist = list("blob") //Just in case a blob survives that long
```
Amelia0010: blob gamemode doesn't exist anymore

code/game/gamemodes/blob/overmind.dm
```diff
@@ -65,7 +71,46 @@
...
+
+		var/area/Ablob = get_area(T)
+		if(!Ablob.blob_allowed)
+			continue
+
+		playsound(L, 'sound/effects/splat.ogg', 50, 1)
+		L.death()
+		new/mob/living/simple_animal/hostile/blob/blobspore(T)
+		for(var/V in sortedAreas)
+			var/area/A = V
```
Amelia0010, line 100: couldn't you just do
>for(var/area/A in sortedAreas)

instead?

ThatLing: This is faster. Doing ``var/area/A in sortedAreas`` will do an ``istype`` check on every element in the list before casting it. But we know sortedAreas only contains ``area`` objects so we can just cast it without checking the type.

code/game/gamemodes/blob/overmind.dm
ThatLing: I will blacklist the event from nukies and rev rounds.

ShadowDeath6:
I don't think this will work with our playerbase. Tg's blob mode was unbalanced and removed due to their players being a lot more proactive and experienced (xray lasers being made every round immediately when blob is called) and I don't think our rounds will be able to handle having a blob dropped in among the other antags, especially group antags like cult or rev, which will probably turn into a steamroll by the blob on the already weakened crew or an unfun steamroll on the blob by 2 lings trying to run at it and armblade its core to death or something along those lines since they HAVE to kill it or they lose automatically.

code/modules/events/blob.dm
ThatLing: I'm just gonna leave it in there in case the gamemode is brought back.

code/game/gamemodes/blob/blobs/core.dm
ThatLing: Ghosting or DCing won't remove it since overmind won't be null.

Amelia0010 (approved):

Alexandria404:
After talking to the rest of council, from those who answered, only negative opinions were voiced about removing the blob as a game mode of its own.
Maybe this PR can be arranged to something else or used as a place for possible ideas, else I´ll be making a brainstorming thread on the forums for possible changes.

As always any ideas for improving the game-mode are welcome, but removing it is not an option for now.

Amelia0010:
AsV override when

ghost:
As an FYI:

Midround blob is now able to win

The shuttle will not leave if midround blob exists

Midround blob now has a strength equal to roundstart blob

Pros of PR:

No more roundstart five minute blob rounds

Interesting interaction where blobs are in combat with other antags sometimes

Crew can have a better shot at winning against the blob since later on in the round R&D tends to be mostly done and science has minerals, etc

Midround blob no longer sucks ass

Cons of PR:

Gamemode that demands full cooperation among the crew is now relegated to a side antagonist


that's basically it. i'm in favor of it because it's probably going to mean that blob steamrolls LESS often, since antags and a better armed crew will be fighting it most of the time, and if the crew is fucked when the blob spawns then they were fucked anyway and the blob is irrelevant

ShadowDeath6:
cons of midround blob
-crew is either worn down from another antag or focused on something else by the time it spawns in
-blob can spawn at a point where it blocks shuttle from being called thus making rounds that should have ended a long time ago (someone bombed the station or something) even longer and doesn't offer any difficulty to the blob or any real entertainment for any party involved
-gamemode that demands full cooperation among the crew can now spawn and win the round if it goes unchecked because it's a side antagonist who can kill the entire station once it grows large enough

ghost:
@ShadowDeath6 

-emergent gameplay

-yeah that's kinda bad

-revenant but with a win condition

ShadowDeath6:
I wouldn't really call it emergent gameplay when it'll happen every round that the blob spawns in on regardless of who's playing

ghost:
eh, depends really

remember that blob can ALREADY spawn midround on any antag at the moment, and this change doesn't increase the weight

if the crew beats it then, then they can still beat it as is @ShadowDeath6 

essentially, if it didn't happen before the change, it shouldn't happen now

ShadowDeath6:
crew doesn't usually beat it, they usually leave or the round ends due to x antag
if they already steamrolled the antags they'll steamroll the blob as well depending on how long it's been left to itself

PR #3009, opened 2018-01-27
holorays!

Amelia0010:
stolen from https://github.com/tgstation/tgstation/pull/35020/files
https://gyazo.com/19cae4bfdac28bfccfdb542a94aa9e03

:cl: MRTY
rscadd: HOLORAYS!
/:cl:


matskuman5:
I don't like the look of that

Alexandria404:
muh immersion is happy.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Might look less awful if you make it move with the hologram and not jump from tile to tile

Kmc2000:
Would it not look less trash if you had the pad fastprocess instead of on movehologram? and just move the beam to the holoman?

simonset55:
Looks like a graphical error, not sure if i like that, but it makes sense since it's the AI

PR #3014, opened 2018-01-27
let's the det's suit contain police batons

Amelia0010:
:cl:
tweak: the detective's suits can now hold his baton!
/:cl:

the armour still can't hold the baton, which I suppose is the tradeoff for being armoured :/

### DIFF
--- a/code/modules/clothing/suits/jobs.dm
+++ b/code/modules/clothing/suits/jobs.dm
@@ -81,7 +81,7 @@
 	item_state = "det_suit"
 	blood_overlay_type = "coat"
 	body_parts_covered = CHEST|GROIN|LEGS|ARMS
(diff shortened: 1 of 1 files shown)


Jc1469:
the most fun part of being detective is losing your baton and shooting people for stealing it. this is a huge removal of fun both from the detective and others.

Amelia0010:
You can still lose it though...

Jc1469:
much less if you have somewhere to put it that isn't your backpack full of your masterful disguise pieces

Amelia0010:
well, if you disguise yourself then you won't be wearing the suit, will you

Jc1469:
yes but if I wear my disguise in the brig I'll get perma'd for having murdered the detective(this has happened) so I have to keep my disguise in my backpack and then people wont be able to disarm me in the brig and cause a shitstorm because of questionable bad balance decisions.

ghost:
lets

not let's

PR #3015, opened 2018-01-28
Fixed borer event

### LINKED ISSUE #2582: Borers are broken
Mins were talking about it on discord, apparently it never triggers and when spawned by admins nothing happens

MacHac:
Fixes #2582.

The event that spawns borers was failing silently; I fixed two issues that were preventing it from functioning normally.

- pollCandidates() sleeps for 30 seconds, so the event would try to start 15 extra times on separate threads (one for each tick of sleep.)  I moved the actual borer selection out of the start function to the less-picky tick() (and added a check to make sure that block of code is only executed once.)
- The arguments given to pollCandidates() were wrong, so the returned list would always be empty.

#### Changelog

:cl:   
bugfix: Cortical Borers, once thought extinct, have begun reappearing on space stations throughout Nanotrasen territory.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer_event.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer_event.dm
@@ -8,9 +8,13 @@
 
 /datum/round_event/borer
 	announceWhen = 3000 //Borers get 5 minutes till the crew tries to murder them.
+	endWhen = 2
 	var/spawned = 0
 
-	var/spawncount
+	var/spawncount = 0
+
+	var/borersSpawned = 0
+	var/list/vents = list()
 
 /datum/round_event/borer/announce()
 	if(spawned)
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
closes https://github.com/yogstation13/yogstation/pull/2990 I assume

MacHac:
Oops.

ThatLing:
I liked your fix more though.

MacHac:
@ThatLing 
Then you should take my fix, but your PR already has an established balance discussion in progress.  Not much point in migrating to a second PR over 7-8 changed lines of code.

Plus, I don't really feel like spending two weeks arguing for/about this. :) I thought this would be a quick and uncontroversial fix, which is part of why I started working on it in the first place.

Issue #2982, opened 2018-01-23
make #2888 not apply to ash drakes


drderp3635:
https://github.com/yogstation13/yogstation/pull/2888
above pr makes it so that mobs get thrown when there are nearby explosions but makes fighting ashdrakes a random shitfest due to thier fireball attacks possesing explosive properties

ThatLing:
@Time-Green

Time-Green:
Feature
This makes ash drakes challenging again

drderp3635:
If you want this to be a feature than you need to rebalance the number of fireballs around it, since currently you can't stand on or adjacent to any tile where a fireball is landing. People thought the field wide fireball attack was an install before but now its not even an exaggeration since it will kill you every time without fail

PR #3039, opened 2018-01-31
Fixes ashdrake pummeling you back to to mars

Time-Green:
actually a rebalance
closes https://github.com/yogstation13/yogstation/issues/2982
:cl:
bug: ashdrakes were getting tired of fireballs flinging shit around
/:cl:


ShadowDeath6:
really no reason to do this, ash drakes were braindead easy to fight and this actually made it hard to beat them. Surprisingly enough megafauna should actually be difficult and even when this bug happens you should be capable enough as a player to beat them regardless and if you can't you shouldn't be trying to kill megafauna anyway.

PR #3043, opened 2018-02-01
Adds department dependent wires!

Amelia0010:
The wires of a door are now seperated into different sets, depending on the door-type:
Command/secure (read: AI/Vault/Bridge)
Security
Engineering
Medical
Science
regular (read: Maint, cargo, service, regular doors)

Why? 
Well, it stops the whole "hack into this one door in primary tool storage, and you can now hack into the armoury and AI core, cause we decided to have the same set of wires for all fo them" without randomizing every wire in the game. There's a clear pattern, it's still usable, it just requires that little bit more effort. Also, this was agreed upon with AsV a while back, but I completely forgot about it and just picked it up and fixed it cause it's 2am and I can't sleep.

:cl:
tweak: Different types of doors now have different wires, depending on the door type. The types are: Command/secure, security, engineering, medical, science, and the rest.
tweak: Any doors from the same type will have the same set of wires, but a door from security won't have the same set of wires as a door in primary tool storage
/:cl:


ShadowDeath6:
is hacking really such an issue that you want to force every player to reveal themselves when trying to hack into any different area

MenacingManatee:
Someone add the removes fun tag.
Also do head office doors count as command or their department for this?

Amelia0010:
they are a blue command door, so they are a command door

ArgyleSweatshop:
-1 every server that had something like this just resulted in a ton of shock deaths and it serves little to no purpose. Good players will just write stuff down for each door

Its pointless and makes the game less fun

Time-Green:
I can understand command and security, but this is too much

Alexandria404:
Love the idea.

AsV9:
I mean the code looks fine but I'm skeptical on the whole idea, feel like the entire thing is mostly unneccesary and is probably just going to turn into more pain.

Amelia0010:
Okay, how about I just add one more set of wires, for secure doors, IE command, security and AI/vault. That way, you can't just hack the one maint door in locker rooms, then hack into vault, but you can still easily remember which ones are what

Nickvr628:
If I use the door painter to change a random door into a sec door, does it change the wiring too? Or have you accounted for that with the code?

Amelia0010:
nope, the wires do not change with that, it only changes when you make a new door

MenacingManatee:
What happens when doors are made with an rcd? Does it default to the regular wiring, or is it dependent on paint job/access selected when making it?

Kmc2000:
ehhh I don't really like this one tbh. I can see where you're coming from but would it not be easier to make hacking all airlocks different?

Jc1469:
replaces learning wires and hacking doors with getting caught learning wires.

we need that shit pr tag back

Amelia0010:
>What happens when doors are made with an rcd? Does it default to the regular wiring, or is it dependent on paint job/access selected when making it?

Dependant on the paint job

>ehhh I don't really like this one tbh. I can see where you're coming from but would it not be easier to make hacking all airlocks different?

Everyone complains that having 6 circuits is bad, how much do you think people would complain if every wire is different?

PR #3046, opened 2018-02-01
Optimizes the asset cache

ThatLing:
Credits to MrStonedOne.

Ports the following PRs:
- https://github.com/tgstation/tgstation/pull/35003

Refer to these for full information.

Removes ``Winexists`` calls because they are blocking and it now sends multiple assets at the same time.


Moja10:
Mad man

PR #3047, opened 2018-02-01
Refactor atom/Initialize

ThatLing:
Credits to Cyberboss.

Ports the following PRs:
- https://github.com/tgstation/tgstation/pull/22778

Refer to these for full information.

Changes ``atom/initialize`` to ``atom/Initialize`` which is more consistent with ``atom/Destroy``. It is now called from ``atom/New`` if the world has already loaded. This also has the added bonus that's is less blocking than before. Tested by loading the server and using some of the items affected. Couldn't find any runtimes but it still needs more testing.


Amelia0010:
Is this a step towards moving things from new -> initialize?

Amelia0010:
https://github.com/yogstation13/yogstation/pull/3082/files

Please rename this initialize -> Initialize and it should be good to merge

PR #3048, opened 2018-02-02
Gives the HoP a mini Energy gun

Amelia0010:
Changes the Energy gun in the closet into a smaller variant. Comes equipped with a flashlight, although it has fewer shots.

The HoP doesn't really need to have a full energy-gun, and having the smaller gun emphasises that it's for self-defense, and not for validhunting.

https://i.imgur.com/Wpf1cHc.png

:cl:
tweak: The HoP has had their energy gun replaced with a standard issue energy pistol, which is much smaller, has a smaller cell capacity, but comes pre-equipped with a flashlight!
/:cl:


Nickvr628:
Are the mini pistols anywhere else? Like the armory or orderable from cargo? This seems like a great thing to give the CMO, RD, etc as a light ranged weapon for defending their department.

Amelia0010:
It's not anywhere else currently, no. But if there's support for it I suppose I can add it

ThatLing (approved):

Ratvar:
now give heads mini-telebatons, so they can't validhunt... wait

PR #3049, opened 2018-02-02
Throwing a baton at people now stuns them

Amelia0010:
Closes #3049 

:cl:
rscadd: Stunbatons will now stun if you throw them at people. Be careful as your target may just catch the baton instead!
/:cl:

EDIT: third commit uses something from this: https://github.com/tgstation/tgstation/pull/19083/files

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -140,29 +140,38 @@
 	else
 		if(!deductcharge(hitcost))
 			return 0
-
-	user.lastattacked = L
-	L.lastattacker = user
+	if(user)
+		user.lastattacked = L
+		L.lastattacker = user
+		L.visible_message("<span class='danger'>[user] has stunned [L] with [src]!</span>", \
+								"<span class='userdanger'>[user] has stunned you with [src]!</span>")
+		add_logs(user, L, "stunned")
 
 	L.Stun(stunforce)
 	L.Weaken(stunforce)
 	L.apply_effect(STUTTER, stunforce)
 
-	L.visible_message("<span class='danger'>[user] has stunned [L] with [src]!</span>", \
-							"<span class='userdanger'>[user] has stunned you with [src]!</span>")
 	playsound(loc, 'sound/weapons/Egloves.ogg', 50, 1, -1)
 
 	if(ishuman(L))
 		var/mob/living/carbon/human/H = L
 		H.forcesay(hit_appends)
-
-	add_logs(user, L, "stunned")
 	return 1
 
 /obj/item/weapon/melee/baton/emp_act(severity)
 	if(bcell)
 		deductcharge(round(bcell.charge / severity))
 	..()
+
+/obj/item/weapon/melee/baton/throw_impact(atom/target)
+	..()
+
+	if(!isliving(target) || !status || loc == target)
+		return
+	var/mob/living/L = target
+	baton_stun(L)
+
+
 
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
why

Nickvr628:
No

KrabSpider:
merge

matskuman5:
Why

Amelia0010:
Why not?

Alexandria404:
I ALWAYS DREAMED OF THIS.

ThatLing:
Maybe add a random chance to not stun them but deal brute damage instead?

Amelia0010:
It already does brute as part of being hit by it, I think. but I suppose I can just put the stun in a prob

ShadowDeath6:
so this is basically an rng chance of instantly winning a fight at close range because you can't miss a throw

Amelia0010:
you can if you can't aim

ShadowDeath6:
point blank in front of someone is literally impossible to miss unless you theow it 90 degrees to the side

Amelia0010:
if you're point blank in front of someone, then you might as well just fucking click them.

Ikoden:
Good change. High risk, high reward.

Geikkamir:
Don't really see how this would be a balance issue. If you miss you now no longer have your baton and your target can easily pick it up and turn it on you. Or, as previously mentioned, if they see it coming they can just catch it and do the same thing.

Especially if it's not even a reliable stun on hit, it seems more like a desperation move than anything.

ShadowDeath6:
you can't miss a throw, you can miss a click on pixels

Geikkamir:
You can also spam a click on pixels. You can't spam a throw.

You would honestly have to be stupid to throw your baton at someone at point blank range. You only get a single shot and it's only a 50% chance to get a stun even if you hit and if they juke or if they press R or if the proc just doesn't go off, they get your stun baton vs. a guaranteed stun if you hit with no downside if you miss and you can just keep spam-clicking until you connect.

ShadowDeath6:
in the majority of situations you miss a click with your stun baton you're going to die unless whoever you're  fighting has no form of stun or lethal weaponry (or they're just bad, in which case you'll probably win no matter what happens)

considering that 50% is still a one in two chance that you instantly bypass dual esword and other protective items (like ablative armor that can negate ranged stuns) without being in melee range it's extremely viable and you might as well take the risk because unless they're godlike they won't react fast enough to A. turn on throw intent (which makes you unable to click things while it's active, clicking anything will disable it regardless if you're holding something) before you throw it or B. pick it up if it doesn't stun

it probably won't matter because security can just spam flashbangs but I don't think that a humorous feature that will rarely ever be used should have the potential to be a better ranged weapon than a taser in a good amount of situations

Super3222:
Make sure shields block this.

ghost:
rip dual energy swords

Geikkamir:
Oh shit I actually hadn't even thought about the fact that this bypasses e-swords/shields

Disregard everything I have said

Ikoden:
That's a good point, Adam. How about making people with desword totally immune to it? Also are shields blocking thrown things?

EDIT
I did not see Geikkamir's post "Oh shit I actually hadn't even thought about the fact that this bypasses e-swords/shields". So, sorry, my apologies for the question about shields. I'm sure the problem can, nonetheless, be solved with some code allowing shields and deswords to totally block it.

Jc1469:
I approve of instant easy to target maxcaps.

Amelia0010:
i'm a moron, ignore me

PR #3070, opened 2018-02-05
Fix examine exploit v2

ThatLing:
Ok, this is more thoroughly tested this time. You can no longer examine items that more than 1 tile away from you and are pitch dark. You can still examine stuff you can see, including stuff in your hands/equipped stuff. This effect is bypassed by equipping mesons, which will allow you to examine any tile again.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -276,6 +276,16 @@ var/next_mob_id = 0
 	set name = "Examine"
 	set category = "IC"
 
+	var/see_turfs = FALSE
+	if(ishuman(src))
+		var/mob/living/carbon/human/H = src
+		if(H.glasses && H.glasses.vision_flags & SEE_TURFS)
+			see_turfs = TRUE
+
+	if(!isobserver(src) && isturf(A) && !(A in view(src)))
+		if(!see_turfs)
+			return
+
 	if(is_blind(src))
 		to_chat(src, "<span class='notice'>Something is there but you can't see it.</span>")
 		return


Jc1469:
Why did this become an issue that needed fixing btw

ThatLing:
Because you can just examine the floor behind a wall and figure out if it's a cultfloor or not.

Jc1469:
Instead of all the spaghetti about lighting and everything, why not just make pylons not convert turfs until its toggled?  That way you can also sneak in a completely un-needed cult buff by making the cult turfs do something like give speedboosts to cultists, while having the pylon just passively do the regen to people near it.

ShadowDeath6:
I'm on mobile so I cba to read the code- does this work with nightvision organs like lings and glasses like thermals? Also if this applies to items and people it's a huge buff to people

ShadowDeath6:
a huge buff to people hiding in the dark since you can't see them

Notamaniac:
How dare people hide in the dark so they can't be seen.

ThatLing:
@Jc1469 
This isn't about specifically about fixing cult floors though. Being able to know what floortile any turf has, whether visible or no, seems like a bug to me.

@ShadowDeath6 
Works with ling and thermals. This changes the examine behaviour so you can't examine things that are completely in the dark. Doesn't affect mobs or items.

matskuman5:
>a huge buff to people hiding in the dark since you can't see them

Why is this a problem

ShadowDeath6:
it's not really a problem unless you want shadowlings to win more, because you can potentially mouse over one in the dark and get away if you notice them

Amelia0010:
>it's not really a problem unless you want shadowlings to win ~~more,~~

FTFY

ThatLing:
Like I said, this doesn't even affect mobs since you can't see them in the dark anyway.

AsV9 (approved):

PR #3075, opened 2018-02-06
[WIP] Makes all the maps playable, and non-runtimey

Amelia0010:
1st commit: Added a clerk room, sig-tech and outdated plasma tank to asteroid station, and added the asteroid and ministation map to the config.


Time-Green:
What about we delete those maps instead?

Amelia0010:
go die

Amelia0010:
Unless you have a suitable replacement for maprotation

Time-Green:
I understand metastation for highpop, but all those other maps are only here to cause fucking conflicts

Amelia0010:
hmm, maybe just deleting the maps and fixing metastation is actually a better solution than manually fixing so many fucking shitty issues

PR #3081, opened 2018-02-06
The emergency shuttle now gets recalled on rev

Amelia0010:
:cl:
tweak: The shuttle will now be randomly recalled during revolutions!
/:cl:

L0rd_english talked about it on discord, and it seems like a harmless thing that makes more sense than "shuttle doesn't leave lul".


Ratvar:
yay

ghost:
easy way to revtest, though, if revs aren't spotted. or you're just not sure. call and if it zips away it's rev.

1fbff5f83b23d39d38b1dfcb4cac8d9b:
Make it get recalled randomly in any gamemode then

PR #1859, opened 2017-02-18
Spawning an academy teacher creates a central command announcement

Super3222:
#### Intent of your Pull Request

Yogstation officially has """""away missions""""".

When the wizard academy spawner creates a wizard the system takes about 50 seconds until Central Command announces that there's a wizard controlled derelict in their start system.

Eventfully telling the crew to basically "pillage and conquer" the ship.

In the future we could do more things like this, and quite possibly expand on it. Who knows?
_Maybe we could disable the teleport until the wizard is spawned_

##### Changelog

:cl:
rscadd: Central Command will now tell you when there's an active wizard academy.
/:cl:

PR #3091, opened 2018-02-07
reenables wizard academy

Amelia0010:
but replaces the teleporter with a teleport beacon. The whole "hey, let's teleport 20 npc wizards to the station" thing is kinda shit, and letting people teleport to the wiz base just increases the chance that the wizard on board actually gets something to do. Do note that the NPCs still exist though, so it's gonna be a pain in the arse to get to the d20

🆑 
rscadd: The wizard academy exists again
/🆑


Time-Green:
Could you fix the teleport runes? There's tele ruins around for accessing parts of te academy otherwise unreachable, and the runes dont work

Amelia0010:
The runes work fine, there's nothing to fix

Super3222:
Can there be an announcement every time the academy is triggered? Sorta like in https://github.com/yogstation13/yogstation/pull/1859

PR #3104, opened 2018-02-10
I LOVE YOU ~~KITCHEN~~ BRIEFCASE GUN

Amelia0010:
Adds a 10tc briefcase gun. Works like a normal briefcase, except using it inhand removes the magazine, using a magazine on it adds the magazine if the gun doesn't have one. I'm more than willing to change the cost, as I just picked 10 randomly

#### Changelog

:cl:  
rscadd: Traitor lawyers are now able to buy briefcase guns, a gun within a briefcase, that works as both a gun, AAAAAANNNNDDD a briefcase
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_item.dm
+++ b/code/modules/uplink/uplink_item.dm
@@ -538,12 +538,6 @@ var/list/uplink_items = list() // Global list so we only initialize this once.
 	item = /obj/item/weapon/pen/edagger
 	cost = 3
 
-/datum/uplink_item/stealthy_weapons/briefcasegun
-	name = "Briefcase Gun"
-	desc = "A gun concealed within a briefcase, used by her majesty's secret service"
-	item = /obj/item/weapon/storage/briefcase/gun
-	cost = 15
-
 /datum/uplink_item/stealthy_weapons/foampistol
 	name = "Toy Gun with Riot Darts"
 	desc = "An innocent-looking toy pistol designed to fire foam darts. Comes loaded with riot-grade \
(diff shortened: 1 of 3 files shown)


code/modules/uplink/uplink_item.dm
```diff
@@ -1088,6 +1088,13 @@ var/list/uplink_items = list() // Global list so we only initialize this once.
 	restricted_roles = list("Janitor")
 	cost = 19 //they need 1 remaining to buy syndicate soap
 
+/datum/uplink_item/role_restricted/briefcasegun
+	name = "Briefcase Gun"
+	desc = "A gun concealed within a briefcase, used by her majesty's secret service"
+	item = /obj/item/weapon/storage/briefcase/gun
+	restricted_roles = list("Lawyer")
+	cost = 10
```
ThatLing: 15tc 🤔

AsV9: :thinking:

Time-Green: 294c94cf-794c-4d9d-a844-748ee1f92272

Ratvar:
I hope you can put a silencer on it seconds before you start shootin

Kmc2000:
neat

code/modules/uplink/uplink_item.dm
AshCorr: it was 15 originally, but I made it buyable only for lawyers ~~nich

Amelia0010: yeah, what I said above^^

ShadowDeath6:
might be a good idea to tell the person buying it what ammo it needs if it's reloadable

code/modules/uplink/uplink_item.dm
Moja10: isn't Majesty capitalized? tut tut

Moja10:
Neat idea, 10 TC for an uzi seems unbalanced but I honestly say keep it and merge anyway because it will be hilarious to witness and use

Ikoden:
If it's an UZI then make it cost more. The idea is great but the price is too low. If you change it to a normal handgun then 10 TC is fine.

Moja10:
Definitely keep it as an uzi rather than a handgun, but yeah, maybe 12 tc so they can get briefcase gun + emag? I still think 10 tc would be acceptable, a large amount of good stuff is blocked by only having 10 tc left to spend. If it was too much TC, like 15+, then there wouldn't be much interesting combos, it would pretty much mean the traitor has a gun + assorted small things, like soap, cham kit, etc. As long as the combo of briefcase + stimpack doesn't exist, i'm fine with it being 10 - 12 tc. 

Also, consider adding the uzi magazines to the menu as well, I think a fair price for those would be 3tc each, possibly.

Ikoden:
Moja gave good suggestions. Make the briefcase gun cost 12 TC and add the ammo for 3 TC.

Amelia0010:
there

PR #3116, opened 2018-02-13
fixes KA shakes

Time-Green:
:cl:
fix: Fixes the kinetic accelerator shaking with a wideguard
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/kinetic.dm
+++ b/code/modules/projectiles/guns/energy/kinetic.dm
@@ -270,11 +270,11 @@
 	rank = "special"
 
 /obj/item/kinetic_part/grip/wide/attach(obj/item/weapon/gun/energy/kinetic_accelerator/KA,mob/user)
-	KA.weapon_weight = WEAPON_MEDIUM
+	KA.w_class = 4
 	KA.trigger_guard = TRIGGER_GUARD_ALLOW_ALL
 
 /obj/item/kinetic_part/grip/wide/detach(obj/item/weapon/gun/energy/kinetic_accelerator/KA,mob/user)
-	KA.weapon_weight = WEAPON_LIGHT
+	KA.w_class = 3
 	KA.trigger_guard = TRIGGER_GUARD_NORMAL
 
 /obj/item/kinetic_part/charger/double_shot


code/modules/projectiles/guns/energy/kinetic.dm
```diff
@@ -270,11 +270,11 @@
 	rank = "special"
 
 /obj/item/kinetic_part/grip/wide/attach(obj/item/weapon/gun/energy/kinetic_accelerator/KA,mob/user)
-	KA.weapon_weight = WEAPON_MEDIUM
+	KA.w_class = 4
```
Kmc2000, line 273: Why did you change it to 4 and not another define? looks weird

code/modules/projectiles/guns/energy/kinetic.dm
```diff
@@ -270,11 +270,11 @@
...
 	KA.trigger_guard = TRIGGER_GUARD_ALLOW_ALL
 
 /obj/item/kinetic_part/grip/wide/detach(obj/item/weapon/gun/energy/kinetic_accelerator/KA,mob/user)
-	KA.weapon_weight = WEAPON_LIGHT
+	KA.w_class = 3
```
Kmc2000, line 277: Again here...

code/modules/projectiles/guns/energy/kinetic.dm
```diff
@@ -270,11 +270,11 @@
 	rank = "special"
 
 /obj/item/kinetic_part/grip/wide/attach(obj/item/weapon/gun/energy/kinetic_accelerator/KA,mob/user)
```
Kmc2000, line 272: This doesn't return ..() , does this code work? I.E was it tested?. Looks like the grip just applies magic to it and is never deleted as an item.

code/modules/projectiles/guns/energy/kinetic.dm
Time-Green: Couldn't find any defines. They were all integers

code/modules/projectiles/guns/energy/kinetic.dm
Kmc2000: Feel free to correct me on that front

Time-Green: switchPart() does the actual attaching and de-attaching. Line 56
I prefer to split my functions, that way I don't need to constantly use ..() and it makes the code easier to work with

Kmc2000: Ahh alright

code/modules/projectiles/guns/energy/kinetic.dm
ShadowDeath6: look at backpacks or something, like w_class = w_class_normal

ShadowDeath6: I was wrong actually, weird

PR #3134, opened 2018-02-19
[admin] VV bugfixes

ThatLing:
Admins keep screaming at me to fix this. This fixes the broken css when viewing a list. Fixed in a retarded way but who needs security anyway. Also changes the font of values to something more readable.


Wardog390:
Yees

PR #3124, opened 2018-02-16
mapdiffbot test

Amelia0010:
### Intent of your Pull Request

Type here what the purpose of your PR is.
Mention if you tested your changes. If you changed a map, make sure you used the mapmerge tool.

Mention if your PR is related to an issue.
You can type "Fixes issue #6" and issue 6 will be automatically closed when your pull request is merged. This is much easier for everyone.

Edit the changelog at the bottom for changes that are noticeable by the players. Remove it if this isn't the case.
If you add a name after the ':cl', that name will be used in the changelog. Leave it empty to use your GitHub name.
Include [admin] in the title if it's something admin related.

#### Changelog

:cl:  
rscadd: Added new things  
rscdel: Removed old things  
bugfix: fixed a few things  
wip: This is work in progress  
tweak: tweaked a few things  
soundadd: added a new sound thingy  
sounddel: removed an old sound thingy  
imageadd: added some icons and images  
imagedel: deleted some icons and images  
spellcheck: fixed a few typos  
experimental: This is experimental  
/:cl:

PR #3137, opened 2018-02-22
Fix various syntactic problems

SpaceManiac:
Code cleanup necessary for MapDiffBot to work, see #3124.

Deletes or fixes nonsense code constructs that *just happen* to compile, because of course they do, but that the map renderer's backend doesn't yet understand.

Some other wacky code constructs were fixed on the parser side, so thanks for the stress-test.

### DIFF
--- a/code/game/machinery/doors/firedoor.dm
+++ b/code/game/machinery/doors/firedoor.dm
@@ -4,8 +4,8 @@
 #define CONSTRUCTION_GUTTED 3 //Wires are removed, circuit ready to remove
 #define CONSTRUCTION_NOCIRCUIT 4 //Circuit board removed, can safely weld apart
 
-/var/const/OPEN = 1
-/var/const/CLOSED = 2
+/var/const/FD_OPEN = 1
+/var/const/FD_CLOSED = 2
 
 /obj/machinery/door/firedoor
 	name = "firelock"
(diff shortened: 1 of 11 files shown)


code/game/machinery/doors/firedoor.dm
```diff
@@ -4,8 +4,8 @@
 #define CONSTRUCTION_GUTTED 3 //Wires are removed, circuit ready to remove
 #define CONSTRUCTION_NOCIRCUIT 4 //Circuit board removed, can safely weld apart
 
-/var/const/OPEN = 1
```
SpaceManiac, line 7: To explain this, another file has `#define OPEN 1`, and BYOND substitutes this but fails to reject the nonsense line `/var/const/1 = 1`.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -368,9 +368,9 @@
 	return 1
 
 
-/obj/item/weapon/storage/empty_object_contents(burn, src.loc)
+/obj/item/weapon/storage/empty_object_contents(burn, new_loc = src.loc)
```
SpaceManiac, line 371: This gets interpreted as a typepath, like `src/loc`, and passes compile because `loc` is never accessed, so the type is never checked.

code/modules/mob/living/simple_animal/friendly/drone/extra_drone_types.dm
```diff
@@ -80,7 +80,7 @@
 	picked = TRUE
 	languages_spoken = RATVAR
 	languages_understood = HUMAN|RATVAR
-	pass_flags = PASSTABLE || PASSMOB
+	pass_flags = PASSTABLE | PASSMOB
```
SpaceManiac, line 83: This is just a programming error that I happened to catch.

